PR #4, opened 2018-07-27
w&b error correction interpolation algorithm implemented

changxu2:
Update with current miniviff 
Not ready for merge

### DIFF
--- a/honeybadgermpc/test_batch_reconstruction.py
+++ b/honeybadgermpc/test_batch_reconstruction.py
@@ -1,0 +1,199 @@
+import asyncio
+import random
+# from honeybadgermpc.wb_interpolate import decoding_message_with_none_elements
+from honeybadgermpc.field import GF
+from honeybadgermpc.polynomial import polynomialsOver
+from honeybadgermpc.batch_reconstruction import batch_reconstruction
+
+order = q = 17
+
+
+def random_element():
+    return random.randint(0, order)
+
+
+def simple_router(N):
+    """
+    Builds a set of connected channels
+    @return (receives, sends)
+    """
+    # Create a mailbox for each party
+    mbox = [asyncio.Queue() for _ in range(N)]
+
+    def makeSend(i):
+        def _send(j, o):
+            # print('SEND %8s [%2d -> %2d]' % (o[0], i, j))
+            # random delay
+            asyncio.get_event_loop().call_later(
+                random.random()*1, mbox[j].put_nowait, (i, o))
+        return _send
+
+    def makeRecv(j):
+        async def _recv():
+            (i, o) = await mbox[j].get()
+            # print('RECV %8s [%2d -> %2d]' % (o[0], i, j))
+            return (i, o)
+        return _recv
+
+    sends = {}
+    receives = {}
+    for i in range(N):
+        sends[i] = makeSend(i)
+        receives[i] = makeRecv(i)
+    return (sends, receives)
+
+
+class Runtime():
+    def __init__(self, id, N, t, send, recv):
+        assert type(n) in (int, long)   # noqa TODO n is undefined
+        assert 3 <= k <= n  # noqa TODO fix: k is undefined
+        self.N = N
+        self.t = t
+        self.id = id
+
+        asyncio.get_event_loop().create_task(self._run)
+
+    async def _run(self):
+        while True:
+            await  asyncio.sleep(1) # noqa TODO fix: await ?; e.g.: await asyncio.sleep(1)
+
+    def createshare(self, val):
+        s = Share(self)
+        s._val = val
+        return s
+
+    def _send():
+        pass
+
+
+class Share():
+    def __init__(self, runtime):
+        pass
+
+    async def open(self, _):
+        # reveal share
+
+        # wait for shares
+        pass
+
+
+async def progtest(shared, N, myid, send, recv):
+    print('myid:', myid)
+    print(shared)
+    Fp = GF(p)
+    Poly = polynomialsOver(Fp)
+    tmp_poly = Poly(shared)
+
+    for i in range(N):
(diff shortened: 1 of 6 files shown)


polynomial.py
```diff
@@ -17,7 +20,7 @@ def __init__(self, coeffs):
             self.coeffs = strip_trailing_zeros(coeffs)
             self.field = field
 
-        def isZero(self): return self.coeffs == []
+        def isZero(self): return self.coeffs == [] or (len(self.coeffs) == 1 and self.coeffs[0] == 0)
```
amiller: This shouldn't be necessary if `strip_trailing_zeros` was called earlier

wb_interpolate.py
```diff
@@ -0,0 +1,83 @@
...
+# decoder uses the Berlekamp-Welch algorithm
+
+# for solving a linear system
+from linearsolver import someSolution
+
+from field import GF
+from polynomial import polynomialsOver
+
+# n is the total number of messages send out == total number of nodes
+# k is the total number of shared secret, k = t + 1 with degree t polinomial
```
amiller: Slight terminology correct - this is Reed Solomon not secret share encoding. So `k` is not the number of shared secrets, but it could be called the number of code symbols.
Polinomial -> Polynomial

changxu2: Fixed!

tests/test_wb_interpolate.py
```diff
@@ -0,0 +1,45 @@
...
+def corrupt(message, numErrors, minVal=0, maxVal=131):
+    indices = random.sample(list(range(len(message))), numErrors)
+
+    for i in indices:
+        message[i][1] = random.randint(minVal, maxVal)
+
+    return message
+
+
+test_decoding()
```
sbellem: We do not need to explicitly invoke the test function (`test_decoding`) as `pytest` takes care of finding and executing the test functions and classes. 

So, when you run

```bash
pytest test_wb_interpolate.py
```

all functions starting with `test_` and classes starting with `Test` will be executed.

More information about this behavior can be found under https://docs.pytest.org/en/latest/goodpractices.html#conventions-for-python-test-discovery.

honeybadgermpc/test_batch_reconstruction.py
```diff
@@ -0,0 +1,199 @@
...
+from honeybadgermpc.batch_reconstruction import batch_reconstruction
+
+order = q = 17
+
+
+def random_element():
+    return random.randint(0, order)
+
+
+def simple_router(N):
```
amiller: We should import this simple_router from somewhere rather than copy-pasting it.

sbellem: We could probably put it in a `pytest` fixture.

sbellem: I would suggest the following:

1. Move the test module `test_batch_reconstruction.py` under the `tests` dir.
2. Add a `simple_router` fixture in `tests/conftest.py`, e.g.:

```python
# tests/conftest.py
import asyncio
import random

from pytest import fixture


@fixture
def simple_router():

    def _simple_router(N):
        """
        Builds a set of connected channels
        @return (receives, sends)
        """
        # Create a mailbox for each party
        mbox = [asyncio.Queue() for _ in range(N)]

        def makeSend(i):
            def _send(j, o):
                # print('SEND %8s [%2d -> %2d]' % (o[0], i, j))
                # random delay
                asyncio.get_event_loop().call_later(
                    random.random()*1, mbox[j].put_nowait, (i, o))
            return _send

        def makeRecv(j):
            async def _recv():
                (i, o) = await mbox[j].get()
                # print('RECV %8s [%2d -> %2d]' % (o[0], i, j))
                return (i, o)
            return _recv

        sends = {}
        receives = {}
        for i in range(N):
            sends[i] = makeSend(i)
            receives[i] = makeRecv(i)
        return (sends, receives)

    return _simple_router
```
3. Remove the `simple_router` function definition in `tests/test_batch_reconstruction.py` and  use the `simple_router` fixture (defined in `conftest.py`) in the test like so:

```python
def test(simple_router):
    N = 4
    p = 73
    t = 1

    async def _test():
        # Test with simple case: n = 4, t =1
        # After AVSS, poly1 = x + 2, poly2 = 3x + 4, secret1 = 2, secret2 = 4
        # Hard code the shared secret value as input into batch_reconstruction function
        # The final constructed polynomial should be p = 4x + 2
        sends, recvs = simple_router(N)
        towait = []
        for i in range(N):
            # ...
```

sbellem: The above is demonstrated in 19737f7fc2a4728ea1bde9c1b185154de52fb1b5 (and 0768e061398cbaa58920d99290c55f01251739a3).

amiller:
@changxu2 what is left to do before this can be merged? I think you need to add some integration with an MPC program, maybe `online_active.py` that uses the batch opening operations and that we can run with custom programs

amiller:
@changxu2 One thing missing from batch_reconstruction is that if a fault is detected, we should return (along with the solution) a list of the player ids corresponding to any faults. Then we can ignore subsequent messages from them and effectively remove them from future comparisons

changxu2:
1. For the simple_router issue: Thank you very much @sbellem ! I indeed encoutered the issue with module name containing "dashes", thank you for pointing out the direction for refactoring!
2. Things to do before merging (that I can think of now): a) I will implement the evil nodes number in return b) some clean up work c) fft version of the algorithm d) active_online.py. For a) , b) and d) , I am planing on finishing before merging. For d) @amiller , do you think it would be better to merge first or implement first and then merge altogether?

amiller:
I think you mean you'll leave c) the fft computation for after merging? If so yes that's fine

amiller:
What is the status of this @changxu2 it looks like a test is failing due to code style but that is it?

PR #38, opened 2018-08-31
BugFix: _reconstruct hangs when N != t+1

### LINKED ISSUE #37: PassiveMpc _reconstruct hangs when future result is set multiple times
When a party gets `t+1` shares we do `self._openings[shareid].set_result(s)`.
The present code:
```
if len(shares) < self.t+1:
            raise NotEnoughShares
        # print('[%d] reconstruct %s' % (self.myid, shareid,))
        s = Poly.interpolate_at(shares)
        # Set the result on the future representing this share
        self._openings[shareid].set_result(s)
```
Consider n=6, t=1, once we get two shares we set the result, but now when we receive the third share it still attempts to set the result which was previously set and it hangs (`set_result` as per the documentation throws an exception when the result has been set previously which for some reason is not getting displayed).

It is working in the present cases because we have n=3, t=2 (i.e. n == t+1)

smkuls:
Only set the result of future if it has not been previously marked done.
This works because seting the result of a future also marks it done.

This fixes #37.

### DIFF
--- a/honeybadgermpc/passive.py
+++ b/honeybadgermpc/passive.py
@@ -51,10 +51,10 @@ class PassiveMpc(object):
 
         # print('[%d] reconstruct %s' % (self.myid, shareid,))
 
-        s = Poly.interpolate_at(shares)
-
-        # Set the result on the future representing this share
-        self._openings[shareid].set_result(s)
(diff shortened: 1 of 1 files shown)


smkuls:
@amiller Done!

Issue #15, opened 2018-08-23
Triple Refinement


amiller:
The triple refinement process relies on batch opening #4, as well as batch beaver #14 

It was already partially implemented as a Viff application. To finish this issue, port it into our framework here
https://github.com/amiller/viff/blob/triple-refinement/apps/triple_refinement.py#L28

amiller:
Triple refinement is done in #41

PR #41, opened 2018-08-31
Implement triple refinement

smkuls:
Issue: #15

* This is simply a port from the Viff implementation.
* There are still some methods which need to ironed out / optimized,
there are TODOs left in appropriate places.

### DIFF
--- a/honeybadgermpc/triple_refinement.py
+++ b/honeybadgermpc/triple_refinement.py
@@ -1,0 +1,63 @@
+from .passive import Poly, Field
+from .polynomial import get_omega
+import asyncio
+import itertools
+
+
+def renameAndUnpackInputs(_a, _b, _c, d, m):
+    n = m // 2  # number of triples
+
+    def pad(arr, k): return arr + [Field(0)]*(d-len(arr))
+    a, b, c = pad(_a[:n], d), pad(_b[:n], d), pad(_c[:n], d)
+    x, y, z = pad(_a[n:], d), pad(_b[n:], d), pad(_c[n:], d)
+
+    return a, b, c, x, y, z
+
+
+def getExtrapolatedValues(a, b, d, omega):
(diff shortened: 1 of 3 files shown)


tests/test_triple_refinement.py
```diff
@@ -0,0 +1,21 @@
+from pytest import mark
+
+
+@mark.asyncio
+async def test_triple_refinement(triples_files_prefix):
+    from honeybadgermpc.passive import runProgramInNetwork, generate_test_triples
+    from honeybadgermpc.triple_refinement import refineTriples
+    N, t = 6, 1
+    # FIXME: This needs to be replaced with a fixture
+    generate_test_triples(triples_files_prefix, N, N, t)
```
smkuls: @sbellem How can I use a fixture here? Note: I want to pass different values of `N` and `t` to `triples_shares_files`

smkuls:
This PR should be merged only after #38

honeybadgermpc/triple_refinement.py
```diff
@@ -0,0 +1,62 @@
...
+    c = []
+    assert len(a) == len(b) == len(x) == len(y) == len(z)
+    for _a, _b, _x, _y, _z in zip(a, b, x, y, z):
+        d, e = context.Share(_a - _x).open(), context.Share(_b - _y).open()
+        c.append(d*e + d*_y + e*_x + _z)
+    return await asyncio.gather(*c)
+
+
+async def refineTriples(context, a_dirty, b_dirty, c_dirty):
+    """This method takes dirty triples and refines them.
```
amiller, line 34: What are the constraints on the number of triples that can be passed in total? Must it be an even number? Is the number that is returned exactly m-f?
It looks like the TODO for chopping the output in case of non-powers-of-2 triples is not yet written

PR #43, opened 2018-09-01
BugFix: Raise exception from background task

### LINKED ISSUE #39: Exceptions from methods running in the background task are not showing up on console
Ideally, #37 should have raised an `InavlidStateError` which should have shown up on the console which didn't happen and the program just hung making it really hard to debug the root cause.

Investigation is required so as to what it will take enable showing up of the error on the console for ease of debugging.

smkuls:
Cause: If _recvloop background task errors out, the prog still continues
to wait for messages which are never going to arrive. This causes the
main event loop to hang. Thus, when background task errors out, stop the
loop explicitly in order to display the exception.

This fixes #39.

### DIFF
--- a/honeybadgermpc/passive.py
+++ b/honeybadgermpc/passive.py
@@ -78,6 +78,14 @@ class PassiveMpc(object):
         # Run receive loop as background task, until self.prog finishes
         loop = asyncio.get_event_loop()
         bgtask = loop.create_task(self._recvloop())
+
+        def handle_result(future):
+            if future.exception():
+                # Stop the loop otherwise the loop continues to await for the prog to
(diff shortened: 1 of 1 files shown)


smkuls:
> can we make this so it does not have to be repeated every time we create a similar background thread?
We could create an abstraction (a method or a class) which allows us to create a background task so we don't do the `handle_result` every time.
> Is there any more local solution besides stopping the entire event loop?
Today, afaik we run the `program` on the main loop, if we were to run it as a task instead then I think we can then just `cancel` that task. In any case, we will have to abort/cancel the task/loop because it will never terminate since it expects a value which will never arrive.
> Also does this bubble up the exception?
Yes.

Issue #44, opened 2018-09-01
Using dataflow in futures like in Viff


amiller:
In Viff the Share objects are subclasses of Future. The operators + * etc. are defined so that if the value of the operands are not available yet, then instead there is a callback added so that when the operands are ready, the computation takes place and the result is set.

This feature makes it easy to write dataflow expressions, that get processed layer by layer, reducing the number of sequential round trips. Without this we have to do it more verbosely.

Consider this fragment for opening `xy` by consuming beaver triple `a`,`b`,`ab`. Because of the awaiting, it proceeds sequentially   
```
    D = await (x - a).open()
    E = await (y - b).open()
    return await (D*b + E*a + ab + D*E).open()
```
We could instead reduce this to just two rounds by explicitly using `gather`:
```
    D,E = await gather((x - a).open(), (y - b).open())
    return await (D*b + E*a + ab + D*E).open()
```

To do something similar to Viff so it's more implicit, we could have a new class:
```
class FieldFuture(Future):
   def __add__(self, other):
        res = FieldFuture()
        if not isWaitable(other):
            other = Future(other)
        gather(self, other).add_done_callback(lambda (a,b): res.set_result(a + b))
        return res
```
compare with https://github.com/amiller/viff/blob/master/viff/passive.py#L133

Then assuming `.open()` returns a FieldFuture, we could use it directly in expressions and get the pipelining/parallel behavior we'd expect.
```
    D = (x - a).open()
    E = (y - b).open()
    return (D*b + E*a + ab + D*E).open()
```

I'm not sure what all the downsides of this are. Here a few details to compare with Viff: 
- When looking at a piece of untyped code, it may be hard to tell if the operations are happening over Shares or FutureShares.
- Viff has every Share be a future by default, so there is no explicit difference of layers between them. 
- There is also not a distinction made between a Share (every party holds a different underlying value) and a field element (every party has the same value). Hence it could make sense to have FieldFuture and ShareFuture be separate objects.

smkuls:
If we introduce `ShareFuture`, then what are the cases wherein we will need just the `Share` object? Won't all `Shares` need to be opened and thus will be of type `ShareFuture` by default?

In case of `Field` values, when are we going to open `Field` elements? If we were to open a `Field` element, won't it be a `Share`?

Although, I agree that introducing callbacks will allows us to write cleaner code which won't be merely sequential.

amiller:
[...]

PR #46, opened 2018-09-03
Dataflow computations with ShareFuture and FieldFuture

amiller:
Added test to show failure in travis

Added a WIP standalone script trying to reproduce the bug in #44

Now with pipelining (but how can we check this?)

### DIFF
--- a/honeybadgermpc/passive.py
+++ b/honeybadgermpc/passive.py
@@ -1,4 +1,3 @@
-from __future__ import annotations
 import asyncio
 from asyncio import Future
 from .field import GF
--- a/honeybadgermpc/passive.py
+++ b/honeybadgermpc/passive.py
@@ -1,4 +1,6 @@
+from __future__ import annotations
 import asyncio
+from asyncio import Future
 from .field import GF
 from .polynomial import polynomialsOver
 from .router import simple_router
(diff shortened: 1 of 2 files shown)


honeybadgermpc/passive.py
```diff
@@ -168,6 +203,36 @@ def __mul__(self, other): raise NotImplemented
 
         def __str__(self): return '{%d}' % (self.v)
 
+    def _binopShare(fut, other, op):
```
smkuls: I see that the `_binopShare` above has an `isinstance` check for `ShareFuture`. Do we still need this method?

honeybadgermpc/passive.py
```diff
@@ -136,6 +138,32 @@ def write_shares(f, modulus, degree, myid, shares):
...
+
+        if isinstance(other, Future):
+            def cb(_): return res.set_result(op(fut.result(), other.result()))
+            asyncio.gather(fut, other).add_done_callback(cb)
+        else:
+            def cb(_): return res.set_result(op(fut.result(), other))
+            fut.add_done_callback(cb)
+        return res
+
+    class FieldFuture(Future):
```
smkuls: How about a class derived from `Future` like `BaseFuture` and then `FieldFuture` and `ShareFuture` derive from `BaseFuture`?

Will this help us avoid having the same code (the overloading of the operators) in two separate places?

honeybadgermpc/passive.py
```diff
@@ -136,6 +138,32 @@ def write_shares(f, modulus, degree, myid, shares):
 
 
 def shareInContext(context):
+
+    def _binopField(fut, other, op):
```
smkuls: What does `binopField` stand for? Can we name it to something better?

PR #50, opened 2018-09-06
Move `test-asyncio-simplerouter` under tests/ dir

sbellem:
**UPDATE**:
* move test module under `tests/` dir
* move some test components under `tests/conftest.py`
* make some test components into fixtures:
    * `simple_router`
    * `share`
    * `runtime`
* added a few checks to the original test to ensure the "simple router" works as expected
* fIx the error stemming from a standalone `await` statement by replacing with  `await asyncio.sleep(1)`.

~~Also added the `test-asyncio-simplerouter` to travis. This is temporary as it would be best to move the test under `tests/`. Some work has already been done towards that in https://github.com/sbellem/HoneyBadgerMPC/commit/0152cc40325a5ae3367ec2c6728772fcf248cf07#diff-484462fced51d1a06b1d93b4a44dd535.~~

~~If this PR does not need to be merged urgently, the above could easily be completed so that the simple router tests are under the `tests/` dir.~~ _done_


smkuls (approved):

smkuls:
@sbellem  Can you also add a description in the PR so as to why it was failing and what the fix is? It will make things easier if someone looks back later

tests/conftest.py
```diff
@@ -96,3 +97,92 @@ def triples_shares_files(request, GaloisField, triples_polys, triples_files_pref
...
+        for i in range(N):
+            sends[i] = make_send(i)
+            receives[i] = make_recv(i)
+        return (sends, receives)
+
+    return _simple_router
+
+
+@fixture
+def random_element():
```
smkuls, line 141: Is this being used anywhere?

tests/conftest.py
```diff
@@ -96,3 +97,92 @@ def triples_shares_files(request, GaloisField, triples_polys, triples_files_pref
     t = request.param['t']
     write_polys(
         triples_files_prefix, GaloisField.modulus, N, t, triples_polys)
+
+
+@fixture
+def simple_router():
```
smkuls, line 103: NYC: I am a little confused, we have a `router.py` which seems to have a similar router and that is what I am using for runnign the programs. If we want to test that then do we need to duplicate this code here?

tests/conftest.py
sbellem: Nope. I did not remove it as perhaps it may be of some use at some point. Better to ask @amiller.

tests/conftest.py
sbellem: Oh yes, I see. Are you using `router.py` in the tests or in the implementation code under `honeybadgermpc`?

PR #51, opened 2018-09-07
Add tests and refactor polynomial.py

smkuls:
* Refactoring and cleaning up polynomial.py
* Adding tests to verify the functionality of the methods


honeybadgermpc/polynomial.py
```diff
@@ -82,6 +80,25 @@ def random(cls, degree, y0=None):
...
+        @classmethod
+        def interp_extrap(cls, xs, omega):
+            """
+            Interpolates the polynomial based on the even points omega^2i
+            then evaluates at all points omega^i
+            """
+            n = len(xs)
+            assert n & (n-1) == 0, "n must be power of 2"
+            assert pow(omega, 2*n) == 1, "omega must be 2n'th root of unity"
+            assert pow(omega, n) != 1, "omega must be primitive 2n'th root of unity"
```
sbellem, line 92: If these `assert` statements are important beyond debugging purposes, we should explicitly raise an error. A `ValueError` would probably be fine.

See https://github.com/initc3/HoneyBadgerBFT-Python/issues/42#issue-353891694 for more information on why the usage of assert statements may be problematic.

honeybadgermpc/polynomial.py
```diff
@@ -135,86 +147,14 @@ def fft_helper(A, omega, field):
...
-    # Interpolate the polynomial up to degree n
-    poly = Poly.interpolate_fft(xs, omega**2)
-
-    # Evaluate the polynomial
-    xs2 = poly.evaluate_fft(omega, 2*n)
+def fft(poly, omega, n, seed=None):
+    assert n & n-1 == 0, "n must be a power of 2"
+    assert len(poly.coeffs) <= n
+    assert pow(omega, n) == 1
+    assert pow(omega, n//2) != 1
```
sbellem, line 154: See comments at https://github.com/initc3/HoneyBadgerMPC/pull/51/files#r216349389

tests/test_polynomial.py
```diff
@@ -0,0 +1,60 @@
+from honeybadgermpc.field import GF
+import honeybadgermpc.polynomial as polynomial
+from random import randint
+
+Field = GF(0x73eda753299d7d483339d80809a1d80553bda402fffe5bfeffffffff00000001)
+Poly = polynomial.polynomialsOver(Field)
```
sbellem: We can make use of the following fixtures: https://github.com/initc3/HoneyBadgerMPC/blob/dev/tests/conftest.py#L27-L38

sbellem: p.s.: If using the fixtures, you'll need to invoke them for each test function: e.g.:

```python
def test_poly_eval_at_k(GaloisField, Polynomial):
    poly1 = Polynomial([0, 1])  # y = x
    # ...
    coeffs = [randint(0, GaloisField.modulus-1) for i in range(d)]
    # ...
```

honeybadgermpc/polynomial.py
sbellem: Also, since this logic is reused elsewhere, it would make sense to put it in a small validation function or method, e.g.:

```python
def validate_omega(xs, omega):
    if n & (n-1) != 0:
        raise ValueError("n must be power of 2")
    if pow(omega, 2*n) != 1:
        raise ValueError("omega must be 2n'th root of unity")
    if pow(omega, n) == 1:
        raise ValueError("omega must be primitive 2n'th root of unity")
```

we could furthermore write a simple [decorator](https://docs.python.org/3/glossary.html#term-decorator) if we do invoke this validation often.

**useful resource on decorators**: https://realpython.com/primer-on-python-decorators/#decorating-classes

honeybadgermpc/polynomial.py
```diff
@@ -94,17 +111,15 @@ def get_omega(field, n, seed=None):
 
     This only makes sense if n is a power of 2!
     """
-    p = field.modulus
+    assert n & n-1 == 0, "n must be a power of 2"
```
sbellem, line 114: See comment at https://github.com/initc3/HoneyBadgerMPC/pull/51/files#r216349389

honeybadgermpc/polynomial.py
```diff
@@ -94,17 +111,15 @@ def get_omega(field, n, seed=None):
...
+    x = field(random.randint(0, field.modulus-1))
+    y = pow(x, (field.modulus-1)//n)
+    if y == 1 or pow(y, n//2) == 1:
         return get_omega(field, n)
-    if pow(y, n//2) == 1:
-        return get_omega(field, n)
-    assert pow(y, n) == 1
-    assert pow(y, n//2) != 1
+    assert pow(y, n) == 1, "omega must be 2n'th root of unity"
+    assert pow(y, n//2) != 1, "omega must be primitive 2n'th root of unity"
```
sbellem, line 122: Same comment as in https://github.com/initc3/HoneyBadgerMPC/pull/51/files#r216349389

honeybadgermpc/polynomial.py
```diff
@@ -117,15 +132,12 @@ def fft_helper(A, omega, field):
     list is of the form [a0, a1, ... , an].
     """
     n = len(A)
-
-    # Check if n is a power of 2.
-    assert not (n & (n-1))
+    assert not (n & (n-1)), "n must be a power of 2"
```
sbellem, line 135: See comment in https://github.com/initc3/HoneyBadgerMPC/pull/51/files#r216349389

amiller: I think all of these are actually only for debugging and are correct as asserts (they are not for validating inputs or messages from other parties)

smkuls: Resolving this and other comments on asserts.

PR #60, opened 2018-09-18
Bgtask cancel

amiller:
If the program does complete without exception, and the background task is cancelled, then calling `future.exception()` will throw an exception. So we should check that the future is not cancelled first before stopping the loop and bubbling up the exception


amiller:
Without this commit, notice that `python -m honeybadger.passive` shows that a `concurrent.futures._base.CancelledError` exception is thrown

smkuls:
Looks good to me!

smkuls (approved):

Issue #59, opened 2018-09-18
Define a serialization format, establish socket connections


amiller:
We still need to define how parties on different machines actually communicate - right now we've been using an abstract message passing style (placing data as python objects in queues). Now we need to serialize/deserialize them over sockets.

One approach is to use a generic serialization format like Pickle. This is the easiest and has the benefit of being easily extended. Otherwise we can pattern match on session-ID and run a custom decoder for each message type (this is also extensible). Another alternative is Protocol Buffers, which is also extensible but requires a bit more effort (it is also faster and portable across languages).

Using serialization and sockets from honeybadgerbft can be a good starting point. Tom has already built up an implementation in the AVSS implementation https://github.com/tyurek/asyncmix/blob/master/HelperClasses.py
 especially the `Sender`, `Listener` class, which associate an IP/port with each node and serialize/deserialize python objects. The `Config` class uses json but could be adapted to something else. Viff has some serialization but it is a bit more buried/coupled in the overall framework and rely heavily on Twisted. We would need to do something similar using sockets in asyncio.

This can be quite simple. For example, here's code I wrote for honeybadgerBFT that does all the serialization and socket establishing in just ~250 lines of code. 
https://github.com/initc3/HoneyBadgerBFT-Python/blob/demo3/run_fifo.py
The code from Tom could also be ported over fairly directly https://github.com/tyurek/asyncmix/blob/master/TestAvssMultipleProcesses.py
Something like zmq would also be fine.

Solving this problem in some way is the main obstacle for beginning benchmarking of online phase experiments (necessary for #30 among other things)

amiller:
Adding some more thoughts to implementing sockets using TLS for authentication....

Conceptually what we want to do is "mutual authentication". Each of the nodes will have a private key (a `certdata/player-$myid.key)`, and a self-signed certificate `certdata/player-$id.cer`. We all have to agree on this set of certificates, but not otherwise have a CA.
The `generate-certificates.py` module adapted from Viff is now setup to run, and by default puts 4 self-signed certificates in `certdata/`. This script uses pyopenssl, the alternative is just to use openssl from the comment line for this:

      openssl req -x509 -newkey rsa:2048 -keyout ./certdata/player-$myid.key -nodes \
                  -out ./certdata/player-$myid.cert -sha256 -days 1000

The python standard library has support for TLS authentication.
https://docs.python.org/3/library/ssl.html

[...]

PR #72, opened 2018-09-29
Serialization format and socket connections

smkuls:
Pending issues:

- If you want to run two MPC programs in the same python script then it will require the listener port to be changed. This is not a blocker now and can be taken later.

### DIFF
--- a/honeybadgermpc/field.py
+++ b/honeybadgermpc/field.py
@@ -24,7 +24,7 @@ arithmetic one would expect: addition, multiplication, etc.
 
 Defining a field:
 
->>> Zp = GF(19)
+>>> Zp = GF.get(19)
 
 Defining field elements:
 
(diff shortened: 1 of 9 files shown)


smkuls:
This closes #59.

amiller:
Are there instructions how to run this? How do you test it? Sylvain probably has ideas of how to integrate such tests in. I also want to get my tmux style script to work.

amiller (approved):
This looks great! I like seeing the sockets done this way directly with `loop.sock_recv` rather than using a larger package wrapped by asyncio. I'd like to record somewhere in the issues your thoughts on why not to use the asyncio stream server.

I am happy to merge it now, but a great question maybe @sbellem has ideas for is how we can make a test case out of running multiple processes.

amiller:
trying to reopen didn't really want to merge yet

amiller:
process_router is not well named, it's not a router. It just open sockets, routing is done by whatever is the external network. Let's give it a name that is more evocative of the design choices it makes, like that each node creates a a single listener on one port, for receiving one connection from each party and currently no authentication. How about:

sendrecv_from_socket
I also think it would be better if we listen on one different port for each client, because then in principle we could use VPN software to handle authentication without us having to add that to our codebase.

Right now, we don't have that option because we just rely on the sender of each message to identify themselves correctly.

It doesn't make sense to have "runProgramAsProcess" in passive.py, nor for passive.py to have to import IPC. Let's just move that into ipc.py

PR #74, opened 2018-09-29
Revert "Serialization format and socket connections"

amiller:
Reverts initc3/HoneyBadgerMPC#72


amiller:
It doesn't make sense to have "runProgramAsProcess" in passive.py, nor for passive.py to have to import IPC. Let's just move that into ipc.py

smkuls:
I thought it would be better to have both the `runAsTasks` and `runAsProcesses` in the same file but then adds this dependency which I agree isn't good. Will remove it.

Issue #75, opened 2018-10-01
Some commits have disappeared?


sbellem:
Wondering where are the commits from PRs #65, #67 and #68

PR #65 was "re-done" in #76 and the diff clearly shows that the commit is somehow gone.

amiller:
I assume they're just squashed when merging (i ususally did squash and merge)

sbellem:
Resolved via #76, #78, #79.

PR #76, opened 2018-10-01
Add early exit & verbose mode in sh scripts for CI

sbellem:
This a the same PR as #65. Not sure what happened to it (see #75). Re-doing the PR but perhaps there is a better way to avoid having duplicate PRs.

See original PR (#65) for comments.


amiller (approved):
OK

PR #78, opened 2018-10-04
Simplify the way the docs are built for travis

sbellem:
Re-doing #67 (see #75):

This was done so that it is a bit more consistent with the way it is done in the `docs/Makefile`. That is, the `docs/Makefile` has:

```make
@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
```

and the script for travis now has:

```yaml
sphinx-build -M html docs docs/_build -c docs -W
```

where `-c docs -W` are the sphinx options.

----

_Documention of the `sphinx-build` command is at http://www.sphinx-doc.org/en/master/man/sphinx-build.html_


amiller (approved):
OK

PR #79, opened 2018-10-04
Add sphinx options to docker-compose builddocs

sbellem:
Re-doing #68 (see #75):

This allows one to quickly test that the docs are being generated without warnings/errors as it is tested on travis.

The [`--keep-going`](http://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-keep-going) option is used so that the docs will be generated despite warnings. This way one can still see what the docs look like.


amiller (approved):
OK

PR #81, opened 2018-10-06
[dont merge] Revert gf

amiller:
This reverts the style of fields.py back to a function GF() that returns an inner object, instead of having an outer class GF. We switched from function to outer class to make it easier to use generic `pickle` when serializing field objects for sending on sockets. Using `__reduce__` it's possible to make this work either way, so I don't think it makes much difference. I have a slight aesthetic preference towards the original way, with a distinct class for each field.

Caveat: it is possible that there is a difference in performance. The __reduce__ method applied to instances of the inner class (Field elements) may be packaging up both the whole modulus each time, whereas __reduce__ applied to the outer class (the Field itself) only needs to be transmitted once.

Long term we will not want to use pickle for performance anyway, we should be able to do much faster encoding given that we know exactly the kinds of protocol messages we'll send.

### DIFF
--- a/honeybadgermpc/field.py
+++ b/honeybadgermpc/field.py
@@ -24,7 +24,7 @@ arithmetic one would expect: addition, multiplication, etc.
 
 Defining a field:
 
->>> Zp = GF.get(19)
+>>> Zp = GF(19)
 
 Defining field elements:
 
@@ -62,7 +62,7 @@ prime modulus (see :func:`GF` for more information):
 Field elements from different fields cannot be mixed, you will get a
 type error if you try:
 
->>> Zq = GF.get(17)
+>>> Zq = GF(17)
 >>> z = Zq(2)
 >>> x + z
 Traceback (most recent call last):
@@ -74,6 +74,8 @@ The reason for the slightly confusing error message is that ``x`` and
 """
 
 from gmpy import mpz
+
+_field_cache = {}
 
 
 class FieldsNotIdentical(Exception):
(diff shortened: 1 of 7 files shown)


honeybadgermpc/field.py
```diff
@@ -94,246 +96,268 @@ def __int__(self):
...
+        return _field_cache[modulus]
+
+    if not mpz(modulus).is_prime():
+        raise ValueError("%d is not a prime" % modulus)
+
+    # Define a new class representing the field. This class will be
+    # returned at the end of the function.
+    class GFElement(FieldElement):
+
+        def __reduce__(self):
```
smkuls, line 141: I had tried something similar, I was still getting the "cannot pickle local object" error so I assumed that a local class can't be reduced using `pickle`. oO

honeybadgermpc/passive.py
```diff
@@ -147,10 +147,10 @@ def write_shares(f, modulus, degree, myid, shares):
 def shareInContext(context):
 
     def _binopField(fut, other, op):
-        assert type(other) in [ShareFuture, GFElementFuture, Share, GFElement]
+        assert type(other) in [ShareFuture, GFElementFuture, Share, Field]
```
smkuls, line 150: You probably also want to rename `GFElementFuture` back to `FieldFuture`

smkuls (commented):
I agree that this is a much cleaner way and we should stick to this.

amiller:
Closing for now. Revisit in asyncmix-to-dev

PR #82, opened 2018-10-11
Ipc parametization and containerization

sbellem:
This PR proposes some changes to `ipc.py` which allows its execution in containerized environments.

The usage of a simple config file was also introduced in order to simply the ipc main command's arguments.

Using the localhost -based approach via `launch_ipc.sh` remains unchanged in terms of usage, except that the script was moved under `scripts/` i.e.: `sh scripts/launch_ipc.sh`,

To simplify the usage two `make` targets were introduced: `ipc` and `ipc-containers`:

Run the nodes on localhost:

```console
$ make ipc 
```

Run the nodes in separate containers:

```console
$ make ipc-containers
````

It is also possible to use `docker-compose` directly:

```console
$ docker-compose -f network.yml up
```

The above displays the logs of each container/node in a sequential manner in one shell.

### DIFF
--- a/honeybadgermpc/ipc.py
+++ b/honeybadgermpc/ipc.py
@@ -30,6 +30,9 @@ class Senders(object):
             while True:
                 msg = await q.get()
                 if msg is None:
+                    print('Close the connection')
+                    writer.close()
+                    await writer.wait_closed()
                     break
                 # print('SEND %8s [%2d -> %s]' % (msg[1][0], msg[0], recvid))
                 data = pickle.dumps(msg)
--- a/honeybadgermpc/ipc.py
+++ b/honeybadgermpc/ipc.py
@@ -1,7 +1,10 @@
+import os
 import pickle
 import asyncio
 import sys
 import struct
+import socket
+
 from .passive import PassiveMpc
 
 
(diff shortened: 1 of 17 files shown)


honeybadgermpc/ipc.py
```diff
@@ -12,24 +15,48 @@ def __init__(self, queues, config):
...
+        # it should not crash the whole thing, or more precisely prevent other nodes
+        # from participating in the protocol.
+        while True:
+            try:
+                addrinfo_list = [
+                    socket.getaddrinfo(self.config[i].ip, self.config[i].port)[0][4]
+                    for i in range(len(self.queues))
+                ]
+            except socket.gaierror:
+                continue
```
amiller, line 34: This should have exponential backoff or 1 second or something

honeybadgermpc/ipc.py
```diff
@@ -12,24 +15,48 @@ def __init__(self, queues, config):
...
+            else:
+                break
+        # XXX END temporary hack
+
         streams = [asyncio.open_connection(
-                self.config[i].ip,
-                self.config[i].port
+                addrinfo_list[i][0],
+                addrinfo_list[i][1],
```
amiller, line 41: Not block on a gather here until all connections are established. Instead the individual process loops should attempt the connection

amiller (approved):
Fantastic! This is ready to go

PR #88, opened 2018-10-17
Add changelog file

sbellem:
Add a file to keep track of important changes.

A good model on how to do this is documented at https://keepachangelog.com/.

Since there will probably not be any version changes for a while, another "checkpoint" can be used such as the name of a milestone that has been reached.

All in all this file can be useful to work with along with the roadmap.

- [ ] Show a concrete example.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,1 @@
+# Changelog


amiller:
Merging this to have the file, lets move adding an example to an issue or just get to it when we have something to write down

PR #91, opened 2018-10-23
Batch and robust reconstruction

amiller:
This implements the batch public reconstruction operation, enabling linear communication online phase MPC.

The main new additions are wb_interpolate, which implements malicious-secure decoding and robust decoding with error correction. The reconstruction is refactored into robust_reconstruction.py, which tries to reconstruct with varying numbers of erasures (attempting to correct errors). 

The batch_reconstruct method in batch_reconstruction.py implements the batch decoding method, in chunks of up to t+1 values at once. Only one message is sent and received for the entire batch.


amiller:
This should now be renamed to robust.py since passive is no longer accurate, but that will require a lot of cosmetic changes, so saving it to another issue

amiller:
We should merge this into dev, and merge into asynchromix as well. We'll want to use batch operations in asynchromix benchmarks. After doing the first benchmark without that (we'll want that as a baseline for non-batch mode mpc anyway), we should merge this in.

PR #92, opened 2018-10-24
Asynchromix

lu562:
1. I noticed that this pull request seems not comparing code in asynchromix with the code in dev branch. Anything wrong here?

2. Somehow relative import for python does not work in my vm, so I just deleted "."  of those import lines in passive.py and field.py. Maybe we can add them back.


smkuls:
> 1. I noticed that this pull request seems not comparing code in asynchromix with the code in dev branch. Anything wrong here?

Can you try to rebase your branch on top of `dev` and then raise the PR. It is very hard to review this PR since it is showing a lot of changes. This branch also has conflicts with the `dev` branch. Can you please resolve this as well?

Once you are done with that, can you also make sure that the CI passes (tests/lint).

honeybadgermpc/passive.py
```diff
@@ -351,13 +347,154 @@ def generate_test_randoms(prefix, k, N, t):
         assert s == 0
     print('[%d] Finished' % (context.myid,))
 
+async def powermix_phase1(context):
```
smkuls: Please move all power mixing code out of this file and import whatever functions you need to call from `passive.py` in the new file.

honeybadgermpc/passive.py
```diff
@@ -52,10 +52,10 @@ def _reconstruct(self, shareid):
 
         # print('[%d] reconstruct %s' % (self.myid, shareid,))
 
-        if not self._openings[shareid].done():
-            s = Poly.interpolate_at(shares)
-            # Set the result on the future representing this share
-            self._openings[shareid].set_result(s)
+        s = Poly.interpolate_at(shares)
```
smkuls: I did a rebase and I think you don't need this change and it is showing up since it probably was there when you branched out. Can you please take a look and undo all the changes that you have not made in `passive.py`?

smkuls:
* Please don't checkin binary files or temp files like `.o`. To keep things cleaner you can update the `.gitignore` to not include them.

smkuls (changes requested):
* Please don't checkin binary files or temp files like `.o`. To keep things cleaner you can update the `.gitignore` to not include them. 
* Avoid checking in the `.input` files and add code to generate and consume them.
* Move all the code out of the root directory to an `apps` folder.
* Please cleanup your code a little bit - remove commented code, meaningful variable names. We also have a flake8 linter which runs as a part of each PR, make sure that it passes.
* Please write some tests to validate the output and add it in the tests folder.

PR #99, opened 2018-10-28
[asyncmix-master] Asynchronous mixing implementation

smkuls:


tests/test_butterfly_network.py
```diff
@@ -0,0 +1,26 @@
...
+from math import log
+import random
+
+
+@mark.asyncio
+async def test_butterfly_network(sharedatadir):
+    import honeybadgermpc.apps.shuffle.butterfly_network as butterfly
+    from honeybadgermpc.passive import generate_test_triples, Field, runProgramAsTasks
+
+    async def verify_output(ctx, **kwargs):
```
amiller, line 12: This test scenario looks great. It uses some preshared inputs, preshared -1/1 shares, run the butterfly network, and check that the output is indeed a permutation (the sorted arrays match).

Dockerfile
```diff
@@ -19,5 +18,14 @@ COPY . /usr/src/HoneyBadgerMPC
 
 RUN pip install -e pairing/
 
+# This is needed otherwise the build for the power sum solver will fail.
+# This is a known issue in the version of libflint-dev in apt.
+# https://github.com/wbhart/flint2/issues/217
+# This has been fixed if we pull the latest code from the repo. However, we want
+# to avoid compiling the lib from the source since it adds 20 minutes to the build.
+RUN sed -i '30c #include "flint/flint.h"' /usr/include/flint/flintxx/flint_classes.h
```
amiller, line 26: ok good pragmatic save...
is there a way to make this assert that we're on the expected version of flint, would we notice if it that changed

amiller (approved):
ok all this looks fine

Dockerfile
smkuls: I have actually opened an issue on the flint repo - https://github.com/wbhart/flint2/issues/512

I will follow up with them regarding this.

PR #101, opened 2018-10-29
Re-work Dockerfile to improve build speed

sbellem:
* Consolidated the `Dockerfile and .ci/Dockerfile` into one `Dockerfile`.
* Change Debian-based Python image instead of Alpine.
* Do not build the pairing wheel, instead just `pip install pairing/`.
* Rust nightly is installed in a `RUN` command so that the layer will be cached. When invoked in a script (`build-wheel.sh`) it would not get cached and thus be re-built each time.
* Put `zfec` and `pycrypto` dependencies in `setup.py`
* Put AWS dependencies into their own extra section. They are installed when using `docker-compose` for development purposes but not installed on Travis CI when running the tests. To install oneself without `docker-compsoe` for dev purposes one must run `pip install -e .[dev,aws]`.


.travis.compose.yml
```diff
@@ -4,8 +4,23 @@ services:
   test-hbmpc:
     build:
       context: .
-      dockerfile: .ci/Dockerfile
+      dockerfile: Dockerfile
```
smkuls, line 7: So only one Dockerile now? YAY!

Dockerfile
```diff
@@ -1,21 +1,14 @@
...
+FROM python:3.7.1-stretch
 
-COPY ./pairing /usr/src/pairing
-#RUN git clone -b docker-build https://github.com/sbellem/pairing.git /usr/src/pairing
-WORKDIR /usr/src/pairing
-
-RUN sh scripts/build-wheel.sh
+ENV PYTHONUNBUFFERED=1
 
+RUN apt-get update && apt-get install -y vim tmux
```
smkuls, line 5: I know that the tests are passing but just curious - are `make` and `gcc` available by default in this image?

.travis.compose.yml
```diff
@@ -4,8 +4,23 @@ services:
...
       args:
         - BUILD
     volumes:
-      - .:/usr/src/HoneyBadgerMPC
+      - ./aws:/usr/src/HoneyBadgerMPC/aws
```
smkuls, line 11: I like this better since this wold keep the docker environment cleaner but we probably should document at some point that in case you add a folder at the root level then remember to update the dockerfile..

smkuls (approved):

Dockerfile
sbellem: yes

.travis.compose.yml
sbellem: This is important in order not to override the built `pypairing` binary.

PR #104, opened 2018-10-29
Make hbavss pass flake8 tests

amiller:
To merge when the checks are green


amiller:
OK just take this to remove the exclusions in flake8

PR #108, opened 2018-10-30
Introduce program id to run programs in parallel

smkuls:


amiller (approved):
- .pid should be renamed to .sid 
Otherwise this approach looks great for running multiple programs sharing a single socket stream

- This is a stopgap abstraction since it supports a simple workflow, 1. add a lot of programs to run, 2. begin running all those programs in parallel.

In the more general case, "arbitrary concurrent composition", such as when invoking MPC in response to blockchain events, we will need to add new programs on the fly, even while earlier programs are still running.

Regardless, approving it now because it's a good choice for our current benchmark needs (running many compute-powers's in parallel)

PR #114, opened 2018-10-30
AWS Benchmark for AVSS

hybridNeo:
will update this with instructions

reviewers : @amiller  @smkuls

### DIFF
--- a/aws/AWSConfig.py
+++ b/aws/AWSConfig.py
@@ -11,13 +11,14 @@ class RegionConfig(object):
 
 
 class MPCConfig(object):
-    def __init__(self, command, t, k, delta, port, numTriples):
+    def __init__(self, command, t, k, delta, port, numTriples, n):
         self.COMMAND = command
         self.T = t
         self.PORT = port
         self.NUM_TRIPLES = numTriples
         self.K = k
         self.DELTA = delta
+        self.n = n
 
 
 class AwsConfig:
@@ -30,7 +31,8 @@ class AwsConfig:
         mpcConfig["k"],
         mpcConfig["delta"],
         mpcConfig["port"],
-        mpcConfig["num_triples"]
+        mpcConfig["num_triples"],
+        mpcConfig["n"]
     )
 
     awsconfig = config["aws"]
--- a/aws/aws-config.json
+++ b/aws/aws-config.json
@@ -1,35 +1,30 @@
 {
     "mpc" : {
-        "command": "python -m honeybadgermpc.ipc",
+        "command": "python -m honeybadgermpc.secretshare_hbavsslight",
         "t": 1,
+        "n": 4,
         "k": 8,
         "port": 7000,
         "delta": -999,
-        "num_triples": 1000
+        "num_triples": 1000,
+        "skipPreProcessing" : true
     },
-
+    
     "aws":{
         "credentials_file_path": "./aws/credentials.secret.json",
         "setup_file_path": "./aws/setup.sh",
 
         "region":{
-            "us-east-1": {
-                "vm_count": 2,
-                "security_group_ids": ["sg-0faaf975"],
-                "image_id": "ami-0ac019f4fcb7cb7e6",
-                "key_file_path": "./aws/samarth.pem",
-                "key_name": "samarth"
-            },
             "us-east-2": {
-                "vm_count": 2,
+                "vm_count": 5,
                 "security_group_ids": ["sg-f211d79b"],
                 "image_id": "ami-0f65671a86f061fcd",
-                "key_file_path": "./aws/samarth-us-east-2.pem",
-                "key_name": "samarth-us-east-2"
+                "key_file_path": "./aws/rahul-us-east.pem",
+                "key_name": "rahul-us-east"
             }
         },
 
-        "vm_name": "samarth-vms",
+        "vm_name": "rahul-vms",
 
         "instance_type": "t2.micro",
         "instance_user_name": "ubuntu",
@@ -38,6 +33,6 @@
     },
 
     "docker": {
-        "image_path": "smkuls/honeybadgermpc:latest"
+        "image_path": "rahulsm2/honeybadgermpc:latest"
     }
 }
--- a/aws/run-on-ec2.py
+++ b/aws/run-on-ec2.py
@@ -37,6 +37,22 @@ def runCommandsOnInstances(
         thread.join()
 
 
+def getHbAVSSCommands(s3Manager, instanceIps, instanceIds):
+    N, t = AwsConfig.MPC_CONFIG.n, AwsConfig.MPC_CONFIG.T
+    port = AwsConfig.MPC_CONFIG.PORT
+    instanceConfig = getInstanceConfig(N, t, port, instanceIps)
+    print(f">>> Uploading config file to S3 in '{AwsConfig.BUCKET_NAME}'<<<")
+    instanceConfigUrl = s3Manager.uploadConfig(instanceConfig)
+    print(">>> Config file upload complete. <<<")
+    setupCommands = [[id, [
+            "sudo docker pull %s" % (AwsConfig.DOCKER_IMAGE_PATH),
+            "mkdir -p config",
+            "cd config; curl -sSO %s" % (instanceConfigUrl),
+            "mkdir -p benchmark",
+        ]] for i, id in enumerate(instanceIds)]
+    return AwsConfig.MPC_CONFIG.COMMAND, setupCommands
+
+
 def getIpcCommands(s3Manager, instanceIps, instanceIds):
     N, t = AwsConfig.TOTAL_VM_COUNT, AwsConfig.MPC_CONFIG.T
     port = AwsConfig.MPC_CONFIG.PORT
@@ -154,6 +170,10 @@ if __name__ == "__main__":
         )
     elif AwsConfig.MPC_CONFIG.COMMAND.endswith("butterfly_network"):
         mpcCommand, setupCommands = getButterflyNetworkCommands(
+            s3Manager, instanceIps, instanceIds
+        )
+    elif AwsConfig.MPC_CONFIG.COMMAND.endswith("secretshare_hbavsslight"):
+        mpcCommand, setupCommands = getHbAVSSCommands(
             s3Manager, instanceIps, instanceIds
         )
     else:
--- a/conf/hbavss.local.ini
+++ b/conf/hbavss.local.ini
@@ -3,6 +3,8 @@
 [general]
 N: 4
 t: 1
+k: 10
+delta: -9999
 skippreprocessing: True
 
 [peers]
--- a/honeybadgermpc/secretshare_hbavsslight.py
+++ b/honeybadgermpc/secretshare_hbavsslight.py
@@ -5,6 +5,7 @@ import os
 import pickle
 import asyncio
 from asyncio import Queue
+from .logger import BenchmarkLogger
 
 
 # secretshare uses reliable broadcast as a sub protocol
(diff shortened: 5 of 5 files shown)


aws/AWSConfig.py
```diff
@@ -18,6 +18,51 @@ def __init__(self, command, t, port, numTriples):
...
+class AVSSConfig(object):
+    def __init__(self, command, N, t, port, skipPreProcessing):
+        self.COMMAND = command
+        self.T = t
+        self.N = N
+        self.PORT = port
+        self.SKIPPREPROCESSING = skipPreProcessing
+
+
+class AwsHBAVSSConfig:
```
smkuls: I am not sure if having this as a separate config is a good idea, AVSS is the first step but then later we do want to run other MPC applications.

And I just realized that that's what you asked when you meant separate file - all I thought was you just wanted to run things. :/

smkuls: Nevermind, I just discussed this with Andrew and he said it is okay this way!.

smkuls: This looks like a lot of duplicated code to me when compared with the existing code. Can we move some of it to a base class and move both configs to a `config` folder?

aws/aws-hbavss-config.json
```diff
@@ -0,0 +1,35 @@
...
+        "skipPreProcessing" : true
+    },
+
+    "aws":{
+        "credentials_file_path": "./aws/credentials.secret.json",
+        "setup_file_path": "./aws/setup.sh",
+
+        "region":{
+            "us-east-2": {
+                "vm_count": 17,
```
smkuls: I would also suggest to change this to something low in the checked in version since if someone runs it by mistake it is going to create 17 VMs.

aws/run-on-ec2.py
```diff
@@ -35,12 +37,57 @@ def runCommandsOnInstances(
         thread.join()
 
 
-if __name__ == "__main__":
-    from honeybadgermpc.passive import generate_test_triples, generate_test_zeros
+def runHBAVSS():
```
smkuls: Can we refactor this and the `runMPC` method and move the common code out?

aws/run-on-ec2.py
```diff
@@ -92,3 +139,12 @@ def runCommandsOnInstances(
     os.makedirs(runId, exist_ok=True)
     runCommandsOnInstances(ec2Manager, logCollectionCommands, True, f"{runId}/benchmark")
     s3Manager.cleanup()
+
+
+if __name__ == "__main__":
+    if(len(sys.argv) < 2):
+        print("Invalid argument. specify job to run")
+    if(sys.argv[1] == 'mpc'):
```
smkuls: Can we change this to `if`, `elif` and `else` with the `else` printing `invalid argument`?

honeybadgermpc/ipc.py
```diff
@@ -79,7 +79,9 @@ def __init__(self, queues, config, nodeid):
     def close(self):
         for q in self.queues:
             q.put_nowait(None)
-        self.benchlogger.info("Total bytes sent out: %d", self.totalBytesSent)
+        nodeid = os.environ.get('HBMPC_NODE_ID')
```
smkuls: The latest code in `asyncmix-master` fixes this by having the logger as a member of the class by passing the nodeid in the constructor. An `os.environ` call will be more expensive.

honeybadgermpc/secretshare_hbavsslight.py
```diff
@@ -241,7 +242,14 @@ def __init__(self, publicparams, privateparams, send, recv):
                 str(sharedkeys[j]).encode('utf-8'), witnesses[j])
         message = pickle.dumps(
             (c, encryptedwitnesses, encryptedshares, crs[0] ** sk))
-        print("Dealer Time: " + str(os.times()[4] - time2[4]))
+
+        dealer_time = str(os.times()[4] - time2[4])
```
smkuls, line 248: I actually wasn't aware of `os.times()` until now. Does it have any advantage over `time.time()`?

I also came across this old issue - https://bytes.com/topic/python/answers/596767-q-result-os-times-different-time-command (could have been fixed by now)

NIT: Can we rename `time2` to `endtime`?

honeybadgermpc/secretshare_hbavsslight.py
```diff
@@ -263,6 +271,7 @@ def __init__(self, publicparams, privateparams, send, recv, reconstruction=True)
         (self.t, self.n, crs, self.participantids,
          self.participantkeys, self.dealerid, self.sid) = publicparams
         self.reconstruction = reconstruction
+        self.time2 = os.times()
```
smkuls, line 274: nit: can we rename this to `endtime` or `starttime` whichever it is?

honeybadgermpc/secretshare_hbavsslight.py
```diff
@@ -241,7 +242,14 @@ def __init__(self, publicparams, privateparams, send, recv):
                 str(sharedkeys[j]).encode('utf-8'), witnesses[j])
         message = pickle.dumps(
             (c, encryptedwitnesses, encryptedshares, crs[0] ** sk))
-        print("Dealer Time: " + str(os.times()[4] - time2[4]))
+
+        dealer_time = str(os.times()[4] - time2[4])
+        print("Dealer Time: " + dealer_time)
+        # benchmarking: time taken by dealer
+        nodeid = os.environ.get('HBMPC_NODE_ID')
```
smkuls: Can we pass this as a parameter to the constructor instead of reading it form the environment variable? I want to avoid reading environment variables in the inner layers.

honeybadgermpc/secretshare_hbavsslight.py
```diff
@@ -326,7 +335,16 @@ def receive_msg(self, sender, msg):
...
                 self.sharevalid = True
                 self.output = self.share
+
                 print('Output available', self.output)
+                recepient_time = str(os.times()[4] - self.time2[4])
+                print("Dealer Time: " + recepient_time)
+                # benchmarking: time taken by dealer
+
+                nodeid = os.environ.get('HBMPC_NODE_ID')
+                benchmarkLogger = BenchmarkLogger.get(nodeid)
```
smkuls: Same comment as above.

smkuls (dismissed):
I have suggested some code quality changes. It will be good to have them before we merge. :)

aws/aws-hbavss-config.json
hybridNeo: ok

smkuls:
Another thing that I forgot to mention. Can you make sure that `sh scripts/launch_hbavss.sh` works after your changes? I think when we run on local on multiple processes then we don't set the `HBMPC_NODE_ID` env variable.

aws/AWSConfig.py
hybridNeo: Traditionally every new application would have a Aws<app>Config . but yeah some sort of code reuse by inheritence is required

hybridNeo: I agree

aws/run-on-ec2.py
hybridNeo: I thought passive was being used only by runMPC

aws/run-on-ec2.py
hybridNeo: my bad

honeybadgermpc/secretshare_hbavsslight.py
hybridNeo: time2 was there in the previous code so I reused it

amiller:
I made an issue to discuss benchmarking of avss that can last longer than this pull request. 
#121  
But let me give a quick suggestion for now.

There are two main parts to `hbAVSSlight.` 
One is the cryptography computation (a compute heavy task). The other is reliable broadcast. (Where the communication occurs, and more likely to be latency bound).

So, could you insert compute time benchmarking, and even better, use cProfile enable / disable to get a summary of what parts of that computation take a while? Thanks @changxu2 for showing how to do this!
https://github.com/changxu2/HoneyBadgerMPC/blob/batch_bm/tests/open_secrets_bm.py#L42

```
cp = cProfile.Profile()
cp.enable()
.....  # Do the computational work, especially "verify_eval" and "decrypt" here in hbAVSS
cp.disable()
cp.print_stats() # Option 1. default
p = pstats.Stats(cp)   # Option 2. What we prefer in hbAVSS
p.strip_dirs().sort_stats(SortKey.CUMULATIVE, SortKey.TIME).print_stats(.5)
```

This is the line where the computing begins, 
https://github.com/hybridNeo/HoneyBadgerMPC/blob/asyncmix-master/honeybadgermpc/secretshare_hbavsslight.py#L309

Here is where the `broadcast` occurs. As far as I can tell, this task kicks off the reliable broadcast then waits for it to complete. So it would be to put start/stop timestamps around that.
https://github.com/hybridNeo/HoneyBadgerMPC/blob/asyncmix-master/honeybadgermpc/secretshare_hbavsslight.py#L443

honeybadgermpc/secretshare_hbavsslight.py
hybridNeo: time2 was there in the previous code so I reused it

aws/aws-config.json
```diff
@@ -1,35 +1,30 @@
...
-                "image_id": "ami-0ac019f4fcb7cb7e6",
-                "key_file_path": "./aws/samarth.pem",
-                "key_name": "samarth"
-            },
             "us-east-2": {
-                "vm_count": 2,
+                "vm_count": 5,
                 "security_group_ids": ["sg-f211d79b"],
                 "image_id": "ami-0f65671a86f061fcd",
-                "key_file_path": "./aws/samarth-us-east-2.pem",
```
amiller, line 27: so....... can we come up with a better way to do this that doesn't involve checking in our respective home folders?

amiller: Or at least, make clear instructions in documentation what it is we need to change? When we hand this over to someone else to run the benchmarks, what instructions do we give them?

conf/hbavss.local.ini
```diff
@@ -3,6 +3,8 @@
 [general]
 N: 4
 t: 1
+k: 10
+delta: -9999
```
amiller, line 7: what is delta? can the config file have a comment explaining it?

honeybadgermpc/secretshare_hbavsslight.py
```diff
@@ -326,7 +334,14 @@ def receive_msg(self, sender, msg):
             if not self.sharevalid and self.okcount >= 2*self.t + 1:
                 self.sharevalid = True
                 self.output = self.share
+
                 print('Output available', self.output)
+                recepient_time = str(os.times()[4] - self.time2[4])
+                print("Recepient Time: " + recepient_time)
+                # benchmarking: time taken by dealer
+                benchmarkLogger = BenchmarkLogger.get(nodeid)
+                benchmarkLogger.info("AVSS recepient time:  " + recepient_time)
```
amiller: spelling: should be "recipient" not "recepient"

honeybadgermpc/secretshare_hbavsslight.py
```diff
@@ -263,6 +270,7 @@ def __init__(self, publicparams, privateparams, send, recv, reconstruction=True)
         (self.t, self.n, crs, self.participantids,
          self.participantkeys, self.dealerid, self.sid) = publicparams
         self.reconstruction = reconstruction
+        self.time2 = os.times()
         (self.pid, self.sk) = privateparams
         assert self.pid != self.dealerid, "HbAvssRecipient, but pid is dealerid"
         (self.sharedkey, self.rbfinished, self.finished, self.sendrecs,
```
amiller, line 279: Can you insert a benchmark logging statement here, for "Begin decryption". Decryption is completed after "verify_eval" completes.

amiller (changes requested):
Main questions - needs benchmarking to isolate the *decryption time* for recipients.

aws/aws-config.json
hybridNeo: we can add more documentation in the readme under the aws folder. I will add that

conf/hbavss.local.ini
hybridNeo: delta is something I needed to add to make hbavss work locally, its something AWSConfig uses, @smkuls is probably using this in butterfly netowrk code. I am unsure of its use

amiller (approved):

PR #115, opened 2018-10-30
Fix dup event loop

sbellem:
:smile: 

A bit complicated to explain why ... will try to do so later.

But perhaps, at the risk of oversimplifying:

It is either pytest-asyncio all the way when async is on, or not at all (e.g.: managing the loop ourselves like it was done in some tests, especially the closing part).

### DIFF
--- a/tests/test_commonsubset_functionality.py
+++ b/tests/test_commonsubset_functionality.py
@@ -1,3 +1,7 @@
-def test_acs_ideal_():
-    from honeybadgermpc.commonsubset_functionality import test_acs_ideal
-    test_acs_ideal()
+from pytest import mark
+
+
+@mark.asyncio
+async def test_acs_ideal_():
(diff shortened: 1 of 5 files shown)


amiller (approved):
Ohhh. Ok that answers the question, I was very confused by it but this makes sense.

sbellem:
> Ohhh. Ok that answers the question, I was very confused by it but this makes sense.

Well, this is really not obvious at first sight and I do not know what is the "exact" mechanics that causes the problem. This may be a problem with `pytest-asyncio` itself (see https://github.com/pytest-dev/pytest-asyncio/issues/29#issuecomment-292230228). 

---

As a side note, it may be useful to know that the `event_loop` fixture provided by the `pytest-asyncio` plugin can be overridden (refs: https://github.com/pytest-dev/pytest-asyncio#pytestmarkasyncio, https://github.com/pytest-dev/pytest-asyncio#event_loop).

PR #122, opened 2018-11-01
Run asyncmix code on AWS

smkuls:


amiller (approved):
Alright! Looks fine by inspection. Having config files to control which things we run looks a little cumbersome. If we write down the workflow for adding a new application to test, it seems like it will involve a lot of places required to change the code. So we should still look for improvements.
Is runID the same as session id / Sid?

smkuls:
It does look cumbersome, I agree. We can definitely revisit this later but my answer to this would be that it depends heavily on the parameters which the application takes. I will try to explore if there are any better ways to provide the parameters and commands - which might imply that we have to do it using more files but might not require any changes in the code.

RunId is something that I have associated with AWS to avoid mixing up the configs. It doesn't map directly to an SID because it is only an AWS specific thing. I was using it in powermixing to avoid mixing multiple outputs or else it would end up summing the results from the previous `power.sums` file.

PR #124, opened 2018-11-09
Fix for hanging IPC code

smkuls:
* Remove sleep
* Display exception in handle_client method


amiller (changes requested):
+1 what a relief. 
We need to modify ipc.py and the other apps "launch_*.sh" with:
`rm sharedata/READY # NOTE: see powermixing.py wait_for_preprocessing`
and probably the launch scripts too

amiller (approved):
OK all looks good

PR #131, opened 2018-11-12
Implement batch butterfly

smkuls:


amiller:
Looks beautiful, and interface compatible already! Great

apps/shuffle/butterfly_network.py
```diff
@@ -35,54 +59,60 @@
...
-        layer1output2.append(t2)
+    n = len(inputs)//2
+    As, Bs, ABs, sbits = getNTriplesAndSbits(tripleshares, randshares, n)
+    Xs = [inputs[2*i] for i in range(n)]
+    Ys = [inputs[2*i + 1] for i in range(n)]
+    layer1output1, layer1output2 = await batchSwitch(ctx, Xs, Ys, sbits, As, Bs, ABs, n)
 
-    layer2outpu1 = await permutationNetwork(
-        ctx, layer1output1, k, num/2, d, randshares, tripleshares, level+1
+    layer2output1 = await permutationNetwork(
```
amiller, line 103: Each recursive call to `permutationNetwork` has a SMALLER input size by half. As a result, this is NOT linear because each batch will be smaller and smaller, and in particular the middle layers will have `N^2 k` communication. This needs t be made into ONE recursive call to be linear communication

PR #133, opened 2018-11-13
Performance improvements to hbAVSS-light

amiller:
This branch contains several performance improvements for hbAVSS-light
- Uses ThreadPoolExecutor and thread-unsafe operations for scalar multiplication in G1
- Does not verify each merkle branch in ReliableBroadcast, optimistically
- Polynomial evaluation is faster w/ inline horner evaluation

There are also fixes to the dockerfile
- Builds the local directory with `pip install --no-cache -e .[$BUILD]` so it can pick up dev changes
- Builds pairing library without `-e` so it is in release mode (faster)

### DIFF
--- a/honeybadgermpc/secretshare_hbavsslight.py
+++ b/honeybadgermpc/secretshare_hbavsslight.py
@@ -6,6 +6,8 @@ import pickle
 import asyncio
 from asyncio import Queue
 from .logger import BenchmarkLogger
+import concurrent.futures
+import psutil
 
 
 # secretshare uses reliable broadcast as a sub protocol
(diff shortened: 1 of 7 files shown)


honeybadgermpc/secretshare_hbavsslight.py
```diff
@@ -92,6 +103,21 @@ def f(poly, x):
     return y
 
 
+def f_horner(poly, x):
+    assert type(poly) is list
+    # previousb = 0
+    k = len(poly) - 1
+    b = ZR(0)
+    # for (i,coeff) in enumerate(poly):
```
smkuls: nit: remove commented code?

scripts/run_hbavss.py
```diff
@@ -38,4 +39,6 @@ def main():
 
 if __name__ == "__main__":
     debug = True
+    # cProfile.run("main()")
+    cProfile
```
smkuls, line 43: That's it? You just write the name of the module and it works? oO

amiller: mmm okay really that is playing the role of `# noqa: E{module imported but not used}`

smkuls: Why do you need to import it if you aren't using it? Can we not completely remove it?

amiller: it's just a script, it's useful to turn on and off that profile output

PR #134, opened 2018-11-14
Adding support for multi session AVSS

hybridNeo:


smkuls (approved):

smkuls:
* @hybridNeo Avoid merge commits. Use rebase instead.
* The PR is failing `flake8` checks, can you fix them?
* Can you rename the PR to something suitable?
* Can you keep only the desired number of commits in the PR? I see 3 (do we need 3?)

honeybadgermpc/ipc.py
```diff
@@ -85,12 +85,15 @@ def __init__(self, queues, config, nodeid):
...
+                # time2 = os.times()
                 data = pickle.dumps(msg)
+                # pickle_time = str(os.times()[4] - time2[4])
+                # print('pickle time ' + pickle_time)
                 padded_msg = struct.pack('>I', len(data)) + data
                 self.totalBytesSent += len(padded_msg)
                 writer.write(padded_msg)
                 await writer.drain()
-                await asyncio.sleep(0.001)
+                await asyncio.sleep(0.00000000001)
```
amiller, line 96: Can we add a comment here explaining why we think this is necessary and drawing attention to it, I think we have an issue about it but it looks like black magic

amiller:
I'm going to approve this for the sake of benchmarking, but lots of nits before this can go from asyncmaster-to-dev:
- "hbavss_multi.py" -> based on the naming convention so far, this file should have the name `secretshare_hbavsslight_multi`. The light is important because we have a different protocol for batch mode which is more expensive for the dealer (but linear for large batches)
- This creates a copy/paste of most of the operations in `secretshare_hbavsslight.py`, instead of copy/paste repeating them, we should import them from modules. Actually it looks like `HbAvssDealer` and `HbAvssRecipient` are the same, only a main function / driver function is new... if so this could go in `scripts/mpc` or just as a modification to `secretshare_hbavsslight.py`
- Actually `secretshare_hbavsslight` should itself be improved and refactored, we already have a polynomials.py, the difference between the operations here is that it makes use of the rust bindings, but we should have something like polynomials_rust.py or something like to address that

amiller:
This needs to be squashed/rebased before I can merge it without creating a merge commit....

hybridNeo:
> I'm going to approve this for the sake of benchmarking, but lots of nits before this can go from asyncmaster-to-dev:
> 
> * "hbavss_multi.py" -> based on the naming convention so far, this file should have the name `secretshare_hbavsslight_multi`. The light is important because we have a different protocol for batch mode which is more expensive for the dealer (but linear for large batches)
> * This creates a copy/paste of most of the operations in `secretshare_hbavsslight.py`, instead of copy/paste repeating them, we should import them from modules. Actually it looks like `HbAvssDealer` and `HbAvssRecipient` are the same, only a main function / driver function is new... if so this could go in `scripts/mpc` or just as a modification to `secretshare_hbavsslight.py`
> * Actually `secretshare_hbavsslight` should itself be improved and refactored, we already have a polynomials.py, the difference between the operations here is that it makes use of the rust bindings, but we should have something like polynomials_rust.py or something like to address that

Actually the only reason I copied stuff over now is because I have added custom print statements and time calculation statements . should not be present in the dev version though

PR #136, opened 2018-11-18
Parallelizing download and upload inputs on AWS.

lu562:


aws/run-on-ec2.py
```diff
@@ -115,6 +116,14 @@ def getButterflyNetworkCommands(s3Manager, instanceIps, instanceIds):
     return AwsConfig.MPC_CONFIG.COMMAND, setupCommands
 
 
+q = Queue()
```
smkuls: I don't think having a global variable is a good idea, unless it is a constant. Can you please move it within one of the functions and then pass it around?

aws/run-on-ec2.py
```diff
@@ -137,17 +146,30 @@ def getPowermixingCommands(runid, s3Manager, instanceIds, instanceIps):
...
+            "cp upload-script/Download_input.sh sharedata/Download_input.sh ",
             "mkdir -p benchmark",
+            "ulimit -n 10000",
         ]
+        threads = []
         for j in range(k):
             fname = f"{powersPrefix}_{runid}_{j}-{i}.share"
-            url = s3Manager.uploadFile(fname)
-            commands.append(f"cd sharedata; curl -sSO {url}")
+            threads.append(threading.Thread(target=uploadFile, args=(fname,)))
```
smkuls: You can get rid of the function and replace it with `lambda fname: q.put(s3Manager.uploadFile(fname))`.

aws/run-on-ec2.py
```diff
@@ -137,17 +146,30 @@ def getPowermixingCommands(runid, s3Manager, instanceIds, instanceIps):
...
             "mkdir -p config",
             "cd config; curl -sSO %s" % (instanceConfigUrl),
             "mkdir -p sharedata",
+            "git clone https://github.com/lu562/upload-script.git",
+            "cp upload-script/Download_input.sh sharedata/Download_input.sh ",
             "mkdir -p benchmark",
+            "ulimit -n 10000",
         ]
+        threads = []
         for j in range(k):
```
smkuls, line 160: nit: This loop can be changed to a list comprehension.

aws/run-on-ec2.py
smkuls: You can start the thread right here instead of iterating over the threads again and starting them. Then, add them to threads list.

aws/run-on-ec2.py
```diff
@@ -137,17 +146,30 @@ def getPowermixingCommands(runid, s3Manager, instanceIds, instanceIps):
...
+        for j in range(k):
+            threads[j].start()
+        for j in range(k):
+            threads[j].join()
+        with open('%s-%d-links' % (runid, i), 'w') as f:
+            while not q.empty():
+                print(q.get(), file=f)
+        fname = f"{runid}-{i}-links"
+        url = s3Manager.uploadFile(fname)
+        commands.append(f"cd sharedata; curl -sSO {url}; sh Download_input.sh {fname}")
```
smkuls: This command will have to be changed now, right?

scripts/aws/download_input.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+pids=""
+RESULT=0
+while IFS='' read -r line ; do
```
smkuls, line 13: Can you add comments in this file like we had discussed?

smkuls (dismissed):
After you make the changes that I have requested, can you also do a sanity run to make sure things actually run?

aws/run-on-ec2.py
lu562: so right now the script is in docker image but how do we get it outside docker? since the first step is that we need to prepare inputs outside the docker container and then pass inputs into container.

aws/run-on-ec2.py
lu562: could you explain this in more detail? Thanks :)

amiller: instead of having `def uploadFile(fname): ....` that is only used in one place, you can make an anonymous function with `lambda fname: ....`. Not a big difference, style mostly

aws/run-on-ec2.py
amiller: I am not sure, @smkuls  what did you have in mind

smkuls: If you look at the docker commands, we do a -v x:y which mounts the folder outside the container to inside the container.

amiller: I see, so we do not need to upload/download the url list via s3, we can just copy to `$x/sharedata` where `$x` is the host location of the mounted volume

aws/run-on-ec2.py
```diff
@@ -137,17 +143,29 @@ def getPowermixingCommands(runid, s3Manager, instanceIds, instanceIps):
     setupCommands = []
     for i, instanceId in enumerate(instanceIds):
         commands = [
-            "sudo docker pull %s" % (AwsConfig.DOCKER_IMAGE_PATH),
             "mkdir -p config",
             "cd config; curl -sSO %s" % (instanceConfigUrl),
             "mkdir -p sharedata",
+            "git clone https://github.com/lu562/upload-script.git",
```
amiller: This looks like it should not be there, can't this be a local file

aws/run-on-ec2.py
lu562: I guess we still need to first use s3 to upload inputs to ec2 first, then mount inputs into containers. My question was our script for downloading input is inside container and we need to extract it out so that we can download input on ec2. And smkuls recommends also uploading scripts using s3.

amiller (approved):

aws/run-on-ec2.py
lu562: this is outdated code, I don't know why it appears here.

amiller:
I'm approving and merging this despite a couple of reservations. Let's revisit this if it turns out that setup fails for some large values of `k`

PR #138, opened 2018-11-18
Disable merkle tree verification for ECHO message

smkuls:


amiller:
This is fine to check into asyncmix-master for benchmarking the optimistic case. Thanks for documenting it as a PR

amiller (approved):

PR #139, opened 2018-11-18
Reuse input across runs on AWS

smkuls:
* Move AWS credentials from file to environment variable.
* Provide options to run `run-on-ec2.py` in setup only, skip setup and
full mode. This allows the user to reuse the setup files which were
uploaded for a larger k for any consecutive runs with a smaller k.

### DIFF
--- a/apps/shuffle/butterfly_network.py
+++ b/apps/shuffle/butterfly_network.py
@@ -196,7 +196,7 @@ if __name__ == "__main__":
         for peerid, addrinfo in config_dict['peers'].items()
     }
 
-    inputs = [Field(i) for i in range(1, k+1)]
+    inputs = [Field(random.randint(0, Field.modulus-1)) for _ in range(k)]
 
     asyncio.set_event_loop(asyncio.new_event_loop())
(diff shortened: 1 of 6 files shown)


apps/shuffle/butterfly_network.py
```diff
@@ -196,7 +196,7 @@ def runButterlyNetworkInTasks():
         for peerid, addrinfo in config_dict['peers'].items()
     }
 
-    inputs = [Field(i) for i in range(1, k+1)]
+    inputs = [Field(random.randint(0, Field.modulus-1)) for _ in range(k)]
```
amiller, line 199: Phew

aws/AWSConfig.py
```diff
@@ -21,6 +23,15 @@ def __init__(self, command, t, k, delta, port, numTriples, n):
         self.n = n
 
 
+def read_environment_variable(key):
+    try:
+        value = os.environ[key]
+    except KeyError:
```
amiller, line 29: Good idea

amiller (approved):
Looks good

PR #141, opened 2018-11-24
Butterfly network and batch reconstruct fixes

smkuls:
Make butterfly network always switch a batch size of k//2. Earlier each
recursive call was getting a smaller input to operate on and thus the size of the
batch which was being switched was less than k//2 thereby incurring
additional communication overhead.

### DIFF
--- a/apps/shuffle/butterfly_network.py
+++ b/apps/shuffle/butterfly_network.py
@@ -1,7 +1,9 @@
 import random
 from math import log
+import os
 import asyncio
 from itertools import islice
+from operator import __mul__, __floordiv__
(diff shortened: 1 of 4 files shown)


amiller:
If this pass tests, it looks correct to me. 
It would also be nice to check that the bandwidth cost is n log n, but i guess we'll see that from the AWS experiment logs...
It's a bit hard to read with 3 nested loops, I wonder if there is some more intuitive way to write this. Something to leave for later or a different issue.
Finally - I'll make a separate issue for wording, but this is a Benes network (which consists of two butterfly networks)

amiller (approved):

Issue #150, opened 2018-11-29
Simulate latency in docker using `tc`


amiller:
The `latencybug-reproduce` branch 
https://github.com/amiller/HoneyBadgerMPC/tree/latencybug-reproduce

has some neat tooling that simulates latency in the docker dev environment using `tc`
https://github.com/initc3/HoneyBadgerMPC/commit/da4995a5677eaa8f7928b32912e0f26a628c68eb
. The only downside is that `docker-compose.yml` has to be set to allow NET_ADMIN to the container.

amiller:
Unassigning to rahul since he's focusing on #29 instead. Will leave this open to assign to a new student

amiller:
Resolved with #312

PR #151, opened 2018-11-29
uses traffic shaping to simulate latency in multi-process experiments

amiller:
This is for #150  
It uses traffic shaping to simulate latency. I found this useful for shaking out errors that occurred in AWS experiments, but did *not* occur in the loopback socket experiments without this latency.

The solution is not too complicated, but validating it was pretty hard-won. I had previously been using some random timeouts in the python code itself. But when it finally mattered, that wasn't any help. In a nutshell, we encountered "hanging" behavior in geo-distributed cloud experiments, that we could produce in a single region or in local experiments, *even* when adding simulated latency via random sleeps.
This bug is preserved in the branch here: https://github.com/initc3/HoneyBadgerMPC/tree/latencybug-reproduce

My main questions to address before merging is:
- [ ] is `add_caps: - NET_ADMIN` in docker-compose really necessary or can we do this some other way? is this too heavyweight? (fortunately `tc` was available without needing additional packages)
- [x] How can we document examples where this is helpful in finding real bugs? Except for this latencybug-reproduce branch, it seems too easy to mask otherwise


smkuls:
Closing this PR since `asyncmix-master` has been merged.

PR #155, opened 2019-01-17
Add logging module and remove print statments

### LINKED ISSUE #40: Use logging package instead of print statements
The `logging` package in python is pretty extensive and supports different levels of logging - DEBUG, INFO, WARNING etc.

With just one setting we can control the level of logs which are spit out on the console, this way we don't have to add/remove print statement again and again.

smkuls:
Closes #40


apps/shuffle/butterfly_network.py
```diff
@@ -75,7 +75,8 @@ def getNTriplesAndSbits(tripleshares, randshares, n):
     # k (log k)^2
     assert k == len(inputs)
     assert k & (k-1) == 0, "Size of input must be a power of 2"
-    benchLogger = BenchmarkLogger.get(ctx.myid)
+    benchLogger = logging.LoggerAdapter(
```
amiller, line 77: Sucks to have two-line + indent log statements, it's a discouragement from using them :( Can we simplify it with an import/def up top?

amiller (dismissed):
I imagine `logging.debug` creates a performance penalty if the serialized elements are large, compared with a print statement that we'd typically comment out for production. Just something to keep in mind

amiller (changes requested):
Make benchmark logger shorter / one line

apps/shuffle/butterfly_network.py
smkuls: As we discussed, this is only for targeted logging only when we want to log benchmark related information and since it needs the `nodeid` it needs to be initialized within the MPC context.

smkuls:
In terms of performance impact:
1. We don't have too many debug statements.
2. If we disable the logging of debug statements (can be done by just switching one flag in logging.yaml) then an additional operation that is performed compared to commented print statements is the evaluation of the format string. I am leaving this as is for now since there are some places where I don't know the type of the data being logged and `%s` might not work.
3. We can revisit this again when we are doing benchmarking to see the additional overhead. I have opened #159 for this.
4. There is another option as specified [here](https://stackoverflow.com/a/16595186/2744506) but this affects readability and makes logging complicated so I want to avoid this.

amiller (approved):

PR #157, opened 2019-01-17
Move subgroup to a constant

### LINKED ISSUE #118: Move the field modulus to a constant
Avoid having to copy the field modulus every time and put it in a separate file. Something like:

```
class FieldModulus:
    BLS = "<field_modulus>

Usage:
FieldModulus.BLS
```

smkuls:
Closes #118


amiller (approved):
Looks sane. Makes the meaning a lot more explicit, I like it

PR #158, opened 2019-01-18
Make names conform to pep8 requirements

smkuls:


apps/shuffle/butterfly_network.py
```diff
@@ -22,50 +22,51 @@
         await asyncio.sleep(1)
 
 
-async def batchBeaver(ctx, Xs, Ys, As, Bs, ABs):
-    Ds = await (Xs - As).open()  # noqa: W606
-    Es = await (Ys - Bs).open()  # noqa: W606
+async def batch_beaver(ctx, xs, ys, as_, bs_, abs_):
+    ds = await (xs - as_).open()  # noqa: W606
+    es = await (ys - bs_).open()  # noqa: W606
```
amiller, line 28: I think it's useful to have a way to distinguish between the "public" values of reconstructed secrets and the "shared" values of secrets themselves. It's used a bit inconsistently here, but the idea has been `D` is a public value while `a,b,x,y,ab` are all secrets. Is there any other convention we could use along these lines that's pep8 compatible? Having the underscore doesn't seem to connote much

amiller (approved):
Thanks. This clearly took a lot of time, but it's an extensive pep8-ization of our codebase. The main changes were getting rid of camelcase, splitting words by underscores, and having consistent capitalization (basically lower_case preferred everywhere). I'm approving it but I'd like to be able to introduce exceptions to the style in some places where it helps map between parameters and their on-paper representation (for example N vs n) and when it's desired to distinguish elements of different types (like G H for elliptic curve elements and x y for scalars in pairing module, or P for a polynomial vs x for an evaluation point in polynomial module, or D,E for public constants versus x,y,a,b,ab in beaver multiplication).

honeybadgermpc/secretshare_functionality.py
```diff
@@ -50,38 +50,38 @@ def __init__(self, sid, N, f, inputFromDealer=None):
...
-                self.inputFromDealer = F_SecretShare.inputFromDealer
+            if myid == dealer:
+                self.inputFromDealer = f_secret_share.inputFromDealer
             else:
                 self.inputFromDealer = None
 
             # A future representing the output is available
-            self.output = F_SecretShare.outputs[myid]
-    return SecretShare_IdealProtocol
+            self.output = f_secret_share.outputs[myid]
```
amiller: This doesn't look right - ideal functionalities in the paper are always indicated by a capital F. A lowercase f doesn't mean the same thing. If lowercase is essential, then it should be replaced by `functionality_` or `func_`

smkuls: pep8 doesn't permit uppercase, I will change it to `func`.

apps/shuffle/butterfly_network.py
smkuls: The reason you see the underscore is because `as` is a reserved keyword in python and as per pep8 a trailing underscore can be used to avoid the name collision.

I understand your point wrt the naming convention on paper, do you think it would be better to add a `#noqa` to ignore this and change it back to capitals?

PR #160, opened 2019-01-22
Integrate FFT in Batch Reconstruction

rahulgovind:


honeybadgermpc/batch_reconstruction.py
```diff
@@ -81,15 +81,30 @@ def attempt_reconstruct_batch(data, field, n, t, point):
...
+        Poly = polynomialsOver(field)
+        zs = [i for i in range(len(data)) if data[i] is not None]
+
+        assert len(zs) >= t + 1
+        zs = zs[:(t + 1)]
+        As, Ais = fnt_decode_step1(Poly, zs, point.omega2, point.order)
+        precomputed_data = (zs, As, Ais)
+    else:
+        precomputed_data = None
+
```
amiller, line 97: Good use of precomputed data here calling `fnt_decode_step1` only once for  batch

honeybadgermpc/polynomial.py
```diff
@@ -275,6 +275,124 @@ def fft(poly, omega, n, seed=None):
...
+class EvalPoint(object):
+    """Helper to generate evaluation points for polynomials between n parties
+
+    If FFT is being used:
+    omega is a root of unity s.t. order(omega) = (smallest power of 2 >= n)
+    i'th point (zero-indexed) = omega^(i + 1)
+
+    Without FFT:
+    i'th point (zero-indexed) = i + 1
+    """
```
amiller, line 369: This `EvalPoint` helper looks like a good design, it's clear to see how it works and it encapsulates having to pass `omega` around separately from `point`

tests/test_batch_reconstruction.py
```diff
@@ -96,6 +97,70 @@ def handle_async_exception(loop, ctx):
         for elem in r:
             assert type(elem) is GFElement
 
+
+@mark.asyncio
+async def test_fft():
```
amiller: this test looks good, a bit repetitive versus `test()` but makes it clear that the evaluation is using the fft eval ponit

amiller (approved):
- really nice first PR
- this will probably get merged after #158 so you should pre-emptively make the new parts pep8 compliant, basically get rid of `camelCaseStyle` preferring `under_score_style` instead

smkuls:
- You have some merge conflicts, can you please resolve those.

- There is a CI pipeline which runs after every PR, it looks like it didn't run for this PR - I think it is probably because there are conflicts - can you fix it and make sure the CI passes?

- I will review your code once the CI passes so please don't merge it until then.

honeybadgermpc/batch_reconstruction.py
```diff
@@ -81,15 +81,30 @@ def attempt_reconstruct_batch(data, field, n, t, point):
...
+        as_, ais_ = fnt_decode_step1(poly, zs, point.omega2, point.order)
+        precomputed_data = (zs, as_, ais_)
+    else:
+        precomputed_data = None
+
     for i in range(n_chunks):
         chunk = [d[i] if d is not None else None for d in data]
         try:
-            p, failures = attempt_reconstruct(chunk, field, n, t, point)
+            p, failures = attempt_reconstruct(chunk, field, n, t, point,
```
smkuls, line 101: code style: I think there are certain guidelines around how you make the split to the next line..
I follow this
```
attempt_reconstruct(
chunk, field, n, t, point, precomputed_data)
```

The other option is to move each parameter to a separate line. What you have used is inconsistent with the rest of the code, can you please change this?

honeybadgermpc/batch_reconstruction.py
```diff
@@ -111,7 +126,8 @@ def with_at_most_non_none(data, k):
             yield x
 
 
-async def batch_reconstruct(elem_batches, p, t, n, myid, send, recv, debug=False):
+async def batch_reconstruct(elem_batches, p, t, n, myid, send, recv, debug=False,
```
smkuls, line 129: same comment as above

honeybadgermpc/batch_reconstruction.py
```diff
@@ -134,7 +150,7 @@ def with_at_most_non_none(data, k):
     benchLogger = logging.LoggerAdapter(
         logging.getLogger("benchmark_logger"), {"node_id": myid})
 
-    def point(i): return Fp(i+1)  # TODO: make it use omega
+    point = EvalPoint(Fp, n, use_fft=use_fft)
```
smkuls: nit: Generally a bad practice to use the same variable name as the arg name `use_fft=use_fft`

tests/test_batch_reconstruction.py
rahulgovind: Should I separate out the code common to both tests?

honeybadgermpc/polynomial.py
```diff
@@ -277,6 +277,124 @@ def fft(poly, omega, n, seed=None):
...
+        Ai(xi) = prod( xi - xj ) for j != i
+    """
+    k = len(zs)
+    omega = omega2**2
+    xs = [omega**z for z in zs]
+
+    # Compute A(X)
+    a_ = poly([1])
+    for i in range(k):
+        print(i, poly([-xs[i], 1]))
```
smkuls: Please avoid using print statements anywhere. Use the logging framework instead. `logging.info` or `logging.debug` or `logging.warn`. Please take a pass over your code and get rid of all of them.

honeybadgermpc/polynomial.py
```diff
@@ -277,6 +277,124 @@ def fft(poly, omega, n, seed=None):
...
+
+    # Compute N'(x)
+    nis = [ys[i] / ais_[i] for i in range(k)]
+    ncoeffs = [0 for _ in range(n)]
+    for i in range(k):
+        ncoeffs[zs[i]] = nis[i]
+    n_ = poly(ncoeffs)
+
+    # Compute P/A(X)
+    # PoverA = -Poly([N(omega**(n-j-1)) for j in range(n)])
```
smkuls: what's `Pover`?

smkuls: Please remove any commented code.

honeybadgermpc/polynomial.py
```diff
@@ -277,6 +277,124 @@ def fft(poly, omega, n, seed=None):
...
+    def __init__(self, field, n, use_fft=False):
+        self.use_fft = use_fft
+        self.field = field
+
+        # Need an additional point where we evaluate polynomial to get secret
+        order = n
+        if use_fft:
+            self.order = order if (order & (order - 1) == 0) else 2 ** order.bit_length()
+
+            self.omega2 = get_omega(field, 2 * self.order, seed=0)
```
smkuls, line 381: any reason why we want to fix the seed?

honeybadgermpc/polynomial.py
```diff
@@ -277,6 +277,124 @@ def fft(poly, omega, n, seed=None):
...
+            self.omega = self.omega2 ** 2
+        else:
+            self.order = order
+            self.omega2 = None
+            self.omega = None
+
+    def __call__(self, i):
+        if self.use_fft:
+            return self.field(self.omega2.value ** (2 * i))
+        else:
```
smkuls, line 391: nit: else is not needed since you have a `return`

honeybadgermpc/robust_reconstruction.py
```diff
@@ -13,17 +13,17 @@
...
     assert sum(f is not None for f in encoded) >= 2*t + 1
 
     # raise ValueError("Sentinel bug")
 
     # interpolate with error correction to get f(j,y)
-    _, decode, _ = make_encoder_decoder(n, t+1, field.modulus)
+    _, decode, _ = make_encoder_decoder(n, t+1, field.modulus, point)
 
-    p = polynomials_over(field)(decode(encoded))
+    p = polynomials_over(field)(decode(encoded, precomputed_data=precomputed_data))
```
smkuls, line 26: nit: consider renaming the arg value and name

honeybadgermpc/wb_interpolate.py
```diff
@@ -142,7 +140,14 @@ def decode(encoded_msg, debug=False):
 
         if e == 0:
             # decode with no errors
-            p_ = poly.interpolate(enc_m)
+            if point.use_fft:
+                zs, as_, ais_ = precomputed_data
+                ys = [m for m in encoded_msg if m is not None]
+                ys = ys[:t + 1]
+                p_ = fnt_decode_step2(poly, zs, ys, as_, ais_,
```
smkuls, line 147: Same comment as above regarding splitting to next line

tests/test_batch_reconstruction.py
```diff
@@ -96,6 +97,70 @@ def handle_async_exception(loop, ctx):
...
+    n = 4
+    p = 97
```
smkuls: Can you change it to `elliptic_curve.BLS...`? It is a larger modulus and we have it in most places in the existing code.

tests/test_batch_reconstruction.py
```diff
@@ -96,6 +97,70 @@ def handle_async_exception(loop, ctx):
...
+    # After AVSS, poly1 = x + 2, poly2 = 3x + 4, secret1 = 2, secret2 = 4
+    # Hard code the shared secret value as input into batch_reconstruction function\
+    point = EvalPoint(fp, n, use_fft=True)
+    omega = point.omega
+    shared_secrets = [(omega ** 0 + 2, 3 * omega ** 0 + 4),
+                      (omega ** 1 + 2, 3 * omega ** 1 + 4),
+                      (omega ** 2 + 2, 3 * omega ** 2 + 4),
+                      (omega ** 3 + 2, 3 * omega ** 3 + 4)]
+    shared_secrets = [(a.value, b.value) for (a, b) in shared_secrets]
+    # shared_secrets = [()]
```
smkuls: remove commented code

tests/test_batch_reconstruction.py
```diff
@@ -96,6 +97,70 @@ def handle_async_exception(loop, ctx):
...
+    # Test 1: Correct decoding with all four points
+    sends, recvs = simple_router(n)
+    towait = []
+    for i in range(n):
+        ss = tuple(map(fp, shared_secrets[i]))
+        towait.append(batch_reconstruct(ss, p, t, n, i,
```
smkuls: Same comment as above regarding splitting to next line

tests/test_batch_reconstruction.py
```diff
@@ -96,6 +97,70 @@ def handle_async_exception(loop, ctx):
...
+        assert r == [2, 4]
+
+    # Test 2: Correct decoding with up to 1 error
+    sends, recvs = simple_router(n)
+    towait = []
+    for i in range(n):
+        ss = shared_secrets[i]
+        if i == 2:
+            ss = tuple(map(fp, (0, 0)))  # add an error
+        towait.append(batch_reconstruct(ss, p, t, n, i,
```
smkuls: Same comment as above regarding splitting to next line

tests/test_batch_reconstruction.py
```diff
@@ -96,6 +97,70 @@ def handle_async_exception(loop, ctx):
...
+    towait = []
+    for i in range(n):
+        ss = tuple(map(fp, shared_secrets[i]))
+        towait.append(batch_reconstruct(ss, p, t, n, i,
+                                        sends[i], recvs[i], True, use_fft=True))
+    results = await asyncio.gather(*towait)
+    for r in results:
+        assert r == [2, 4]
+
+    # Test 2: Correct decoding with up to 1 error
```
smkuls: Do you think it will be better to separate these to multiple tests if they are separate? It shouldn't happen that the first test fails and then the rest of the test isn't executed (though Python might actually execute it)

tests/test_batch_reconstruction.py
```diff
@@ -96,6 +97,70 @@ def handle_async_exception(loop, ctx):
...
+    # Test 3: If there is an error and one crashed node, it will time out
+    sends, recvs = simple_router(n)
+    towait = []
+    for i in range(n):
+        ss = tuple(map(fp, shared_secrets[i]))
+        if i == 2:
+            continue  # skip this node
+        if i == 3:
+            ss = tuple(map(fp, (0, 0)))  # add an error
+        towait.append(batch_reconstruct(ss, p, t, n, i,
```
smkuls: Same comment as above regarding splitting to next line

smkuls (changes requested):

honeybadgermpc/polynomial.py
rahulgovind: Sure

honeybadgermpc/polynomial.py
rahulgovind: All right

honeybadgermpc/batch_reconstruction.py
rahulgovind: Why so?

honeybadgermpc/batch_reconstruction.py
rahulgovind: There's actually a lot of code (at least in batch_reconstruction.py and polynomial.py) that follows this style. I believe it'll actually be harder to just split the arguments to a different line if a function has a lot of arguments.

honeybadgermpc/polynomial.py
rahulgovind: All parties must arrive at the same omega for the algorithm to work. Fixing the seed was the easiest way to do that that I could think of.

honeybadgermpc/robust_reconstruction.py
rahulgovind: Are you referring to precomputed_data?

honeybadgermpc/batch_reconstruction.py
smkuls: I guess it is okay. You can resolve comments related to these.

I checked PEP8 - this seems to be a permitted style.

tests/test_batch_reconstruction.py
rahulgovind: Sure

honeybadgermpc/batch_reconstruction.py
smkuls: Andrew and I had a discussion the other day where we wanted to actually avoid code like `x = open(x)` and this reminded me of that - that would ease debugging and improve readability but I guess for function args those arguments don't hold.

I guess it is okay to leave it as is.

honeybadgermpc/polynomial.py
rahulgovind: That's true. But in this case, the 'else' statement makes it easier to figure out that the second route is taken only when fnt is not being used right?

honeybadgermpc/batch_reconstruction.py
rahulgovind: All right.

honeybadgermpc/polynomial.py
smkuls: It is way obvious with the return and there's only one statement in the if. `else` serves no purpose.

honeybadgermpc/robust_reconstruction.py
smkuls: I think this is fine for function args. Closing this.

tests/test_batch_reconstruction.py
rahulgovind: I think so too. That makes it easier to figure out what test is actually being run. But it would be better to set up common test cases inside by subclassing pytest.TestCase, since most of these tests have common setup code.

rahulgovind: For now, I just looked at how the other batch reconstruction tests were being done and used the same format.

smkuls: Yea the existing code is far from perfect. Most of it was written when we were in a time crunch and it has been reviewed properly or written properly.

honeybadgermpc/polynomial.py
rahulgovind: I'll document this in the code.

tests/test_batch_reconstruction.py
rahulgovind: Should I refactor the tests?

smkuls: Yea, it would be better to refactor them.

smkuls (approved):

PR #168, opened 2019-01-29
Extract preprocessing code out to a seprate file

smkuls:
* Move preprocessing code to a separate file and get rid of all file IO in all the apps
* This breaks the AWS code which will be fixed later
* Includes some other naming fixes


honeybadgermpc/preprocessing.py
```diff
@@ -0,0 +1,157 @@
+import logging
```
amiller, line 1: Excellent - this file is a good idea and refactoring the rest of the code to make use of this will help.

Two things to keep in mind that might help focus how this evolves over time:
- there seems to be no universal set of preprocessing, instead it's likely we'll want all of the ones currently here and more
- In general there are a few different ways to get preprocessing, one is for testing from pre-generated files, the other is as a result of output from a prior conversion step. Clearly for testing we want to simplify the procedure for using pre-generated files, but it's only one of the modes.

If we want to support multiple concurrent applications that use different preprocessing, they may each need to "reserve" a portion of the preprocessing data. 

Also, not all applications need all the preprocessing data. So maybe "wait for preprocessing" should be parameterized by how many of which kinds of preprocessing data we need?

tests/test_passive.py
```diff
@@ -2,21 +2,19 @@
...
-@mark.usefixtures('zeros_shares_files')
-async def test_open_shares(zeros_files_prefix):
+async def test_open_shares():
     from honeybadgermpc.mpc import TaskProgramRunner
+    from honeybadgermpc.preprocessing import PreProcessedElements
+
     N, t = 3, 1
     number_of_secrets = 100
+    pp_elements = PreProcessedElements()
+    pp_elements.generate_zeros(100, N, t)
```
amiller: It seems like this approach means we have to generate the preprocess elements each time we run the test, as opposed to re-using the same pregenerated values across tests. Will this make running the tests significantly slower?

honeybadgermpc/preprocessing.py
smkuls: > there seems to be no universal set of preprocessing, instead it's likely we'll want all of the ones currently here and more

Agreed, I have structured the current code in a way that you can use a mix of them by generating and calling a get on the ones that you specifically need. The generation of course will be moved out later since that is expected to happen during the offline phase but the getters as in the current code should suffice.

> In general there are a few different ways to get preprocessing, one is for testing from pre-generated files, the other is as a result of output from a prior conversion step. Clearly for testing we want to simplify the procedure for using pre-generated files, but it's only one of the modes.

I agree, I completely forgot why we had a single source for generating the test preprocessed elements. I am going to fix that part so we reuse during tests.

> If we want to support multiple concurrent applications that use different preprocessing, they may each need to "reserve" a portion of the preprocessing data.

Haven't worked out on how this is going to happen - this would be one of the first steps for offline phase.

> Also, not all applications need all the preprocessing data. So maybe "wait for preprocessing" should be parameterized by how many of which kinds of preprocessing data we need?

With the current code, since we have to generate the preprocessed elements that we need, all we really need is one wait_for_processing when everything is ready, waiting for something specific doesn't seem to provide any benefit. In fact, with the actual offline phase implementation, this wait shouldn't be there since the preprocessed elements should be ready.

tests/test_passive.py
smkuls: Will fix this to have preprocessing run only once for tests.

PR #170, opened 2019-02-04
Extract preprocessing code out to a seprate file

smkuls:


amiller (approved):
OK. My main request is dealt with, tests reuse preprocessing, so I'm happy with this for now

PR #171, opened 2019-02-06
Multiple fixes

smkuls:
* Implement BeaverTriples and DoubleSharings as mixins
* Cleanup test code to remove unnecessary code and simplify some tests
* Rename test_passive.py to test_mpc.py and move relevant tests to
test_mixins.py
* Remove hard code Field and Poly from mpc.py and move to `Mpc`
* Add `generate_double_shares` to preprocessing.py. Instead of making
two separate calls to `generate_share` this allows to fetch them at one
shot avoiding the need for any share_id.


amiller (approved):
This looks really good!

PR #173, opened 2019-02-07
Implement batch versions of mixins

smkuls:
* Add tests to verify their implementation
* Move all code to use mixins including ProcessProgramRunner


amiller (approved):
Looks excellent.
This seems to address all the cases we want, being able to write programs that use multiplication but are agnostic to whether it uses triples or online-double reductions.

PR #174, opened 2019-02-08
Update AWS code to be in sync with the latest code

smkuls:
* Have verified that both powermixing and butterfly network run on AWS with the updated code

### DIFF
--- a/aws/README.md
+++ b/aws/README.md
@@ -38,7 +38,7 @@ There is a `BenchmarkLogger` which must be used when logging any benchmak data. 
 Use the following commands to push the image to dockerhub. Avoid building the image using `docker-compose` since it doesn't the ignore files specified in `.dockerignore`.
 ```
 cd /path/to/HoneyBadgerMPC
-docker build -t honeybadger .
+docker build -t honeybadger . --build-arg BUILD=dev
 docker tag honeybadger:latest smkuls/honeybadgermpc:latest # Replace with appropriate DockerHub location
 docker push smkuls/honeybadgermpc:latest # Replace with appropriate DockerHub location
 
--- a/aws/aws-config.json
+++ b/aws/aws-config.json
@@ -1,30 +1,29 @@
 {
     "mpc" : {
-        "command": "python -m honeybadgermpc.hbavss_multi",
+        "command": "python -m honeybadgermpc.ipc",
         "t": 1,
         "n": 4,
-        "k": 1000,
+        "k": 8,
         "port": 7000,
         "delta": -999,
         "num_triples": 1000,
         "skipPreProcessing" : true
     },
-    
+
     "aws":{
-        "credentials_file_path": "./aws/credentials.secret.json",
         "setup_file_path": "./aws/setup.sh",
 
         "region":{
-            "us-east-2": {
-                "vm_count": 5,
-                "security_group_ids": ["sg-f211d79b"],
-                "image_id": "ami-0f65671a86f061fcd",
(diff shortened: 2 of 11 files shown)


aws/README.md
```diff
@@ -38,7 +38,7 @@ There is a `BenchmarkLogger` which must be used when logging any benchmak data.
 Use the following commands to push the image to dockerhub. Avoid building the image using `docker-compose` since it doesn't the ignore files specified in `.dockerignore`.
 ```
 cd /path/to/HoneyBadgerMPC
-docker build -t honeybadger .
+docker build -t honeybadger . --build-arg BUILD=dev
```
amiller, line 41: what's that for?

aws/aws-config.json
```diff
@@ -1,30 +1,29 @@
...
-                "security_group_ids": ["sg-f211d79b"],
-                "image_id": "ami-0f65671a86f061fcd",
-                "key_file_path": "./aws/rahul-us-east.pem",
-                "key_name": "rahul-us-east"
+            "us-east-1": {
+                "vm_count": 4,
+                "security_group_ids": ["sg-0faaf975"],
+                "image_id": "ami-0ac019f4fcb7cb7e6",
+                "key_file_path": "./aws/samarth-us-east1.pem",
+                "key_name": "samarth-us-east1"
```
amiller: Now that we're working in `dev`, we should try to avoid this. AWS tests should use a per-user config, or even an AWS_KEY_NAME environment variable

aws/aws-config.json
```diff
@@ -33,6 +32,6 @@
     },
 
     "docker": {
-        "image_path": "rahulsm2/honeybadgermpc:latest"
+        "image_path": "smkuls/honeybadgermpc:latest"
```
amiller: Similarly if dockerhub is an essential part of our AWS usage we should document that specifically

amiller (changes requested):
This looks fine to me, as far as I can tell. Possibly we shouldn't approve this merge until someone else (@lu562) confirms they can run it themselves too. 
I'm requesting changes on the parts that explicitly refer to a specific key file in your home directory or a particular user name at the dockerhub service

aws/aws-config.json
smkuls: I will change these to generic values instead of my own and leave this file to be used only for reference.

aws/aws-config.json
smkuls: This has already been documented in the AWS Readme. I can create a HoneyBadgerMPC repo on Dockerhub and remove my username from this.

aws/README.md
smkuls: The Dockerfile has an argument to pip install which works differently for tests, docs and dev. This needs to be passed to build the correct image.

lu562:
I confirmed that I can run this code for ec2 tests of both powermixing and butterfly network.

aws/aws-config.json
smkuls: I have created a `dsluiuc` organization on DockerHub. We can use that from now on.

aws/aws-config.json
smkuls: I have changed these to generic values. Since we support multiple geos having an environment variable won't work since we will need more than one.

I have left this config to be used for reference and any changes to it won't be checked in as I have added it to `.gitignore`.

amiller (approved):

amiller:
OK, I acknowledge my requested changes were implemented

PR #176, opened 2019-02-09
Speed up batch reconstruction using Cython and NTL

rahulgovind:
The batch reconstruction has been refactored to split it into two steps

- an optimistic part where we assume that no malicious parties are present

- a pessimistic part where malicious parties are present

The pessimistic and slow part is called only if the first 2t+1 parties do not
agree on the same polynomial.

The following Cython/NTL-based functions are exposed in Python and are
significantly faster than their Python counterparts
- interpolate (Langrange interpolation)
- vandermonde_batch_interpolate (Interpolate multiple polynomials using
vandermonde matrices)
- vandermonde_batch_evaluate (Evaluate multiple polynomials using Vandermonde
matrices)
- fft_interpolate (Interpolate function using FNT decoding)
- fft_batch_interpolate
- fft (Converts polynomial to fft representation / evaluates polynomial at
points omega^0, omega^1 ...)

The functions take integers and lists of integers as inputs and return lists of
integers. No special field objects!


honeybadgermpc/batch_reconstruction.py
```diff
@@ -126,6 +128,106 @@ def with_at_most_non_none(data, k):
...
+
+
+def merge_lists(lists):
+    result = []
+    for l in lists:
+        result += l
+    return result
+
+
+async def batch_interpolate(data_receivers, field, point, n, t):
```
amiller, line 189: Thanks --- this is a much improved refactor already just in the python code. Having it previous part of the "wb_interpolate" world was the wrong choice, so this is better

honeybadgermpc/ntl/ccobject.h
```diff
@@ -0,0 +1,57 @@
+/*
+ * Credits to sagemath
```
amiller: Add slightly more detail - just in case we want to track this down and make a more formal reference?

honeybadgermpc/ntl/helpers.h
```diff
@@ -0,0 +1,211 @@
...
+
+    int d = deg(A);
+    // Differentiate polynomial A
+    vec_ZZ_p Ad_coeffs;
+    Ad_coeffs.SetLength(d);
+    for (int i=0; i < d; i++) {
+        Ad_coeffs[i] = (i + 1) * coeff(A, i + 1);
+    }
+
+    // Evaluate derivative at x0, x1, ..., x_(n-1). Then cherry pick ones present in zs
```
amiller, line 162: This is interesting. Your version of the `fnt_decode` seems to follow the original paper more closely than what I ended up writing in python, to compare: https://github.com/initc3/HoneyBadgerMPC/commit/1e540825cedf8e4f344c84cb9be9360d41fe2af5#diff-740f99d2f265e8c8531acd2645b990eaR278 
Did you work through their derivation and understand? Do you have any idea where the difference ends up?

tests/test_ntl.py
```diff
@@ -0,0 +1,119 @@
...
+    n = 4
+
+    # When
+    fft_rep = fft(coeffs, omega, p, n)
+
+    # Then
+    assert fft_rep == [1, 5, 12, 8]
+
+
+def test_fft_big(galois_field, galois_field_roots):
```
amiller, line 59: Solid tests

amiller (approved):
This looks solid to me. Amazing second PR. I'm giving an approving review but @smkuls is welcome to review too

amiller:
Really impressed with this PR, nice coding :1st_place_medal:

honeybadgermpc/ntl/ccobject.h
rahulgovind: Added a link to the original file. Should I add anything else?

.travis.compose.yml
```diff
@@ -24,3 +24,4 @@ services:
       - ./pairing/benches:/usr/src/HoneyBadgerMPC/pairing/benches
       - ./pairing/Cargo.toml:/usr/src/HoneyBadgerMPC/pairing/Cargo.toml
       - ./pairing/setup.py:/usr/src/HoneyBadgerMPC/pairing/setup.py
+      - /usr/src/HoneyBadgerMPC/honeybadgermpc/ntl  # Directory _not_ mounted from host
```
smkuls, line 27: We should document this to avoid having to debug this kind of issue again.

Can you open an issue and take that up in a separate PR?

honeybadgermpc/batch_reconstruction.py
```diff
@@ -145,17 +247,14 @@ def with_at_most_non_none(data, k):
 
     Reconstruction takes places in chunks of t+1 values
     """
-    fp = field = GF.get(p)
+    fp = GF.get(p)
     poly = polynomials_over(fp)
-    bench_logger = logging.LoggerAdapter(
```
smkuls, line 150: This was added to benchmark the time taken to finish the two rounds. We'll need that information when running experiments. Have you moved this somewhere else?

honeybadgermpc/ntl/helpers.h
rahulgovind: Yeah. I followed through their derivation and everything looks right. However, I still don't understand why they are using polynomial multiplication instead of FFT to directly evaluate the polynomial at the omega^i points. 

The changes I made were primarily to make step 1 of the FNT decode O(N log N) too. This is because while step 1 is only done once per batch, the batches need to be exceedingly large for large n. (For n=512, batch size needs to be greater than 30000 to amortize an O(N^2) cost). 

So right now, the entire FNT decode process has an O(N log N) complexity.

rahulgovind:
> Really impressed with this PR, nice coding 🥇

Thanks! 😄

setup.py
```diff
@@ -16,7 +18,7 @@
     'zfec',
     'pycrypto',
     'cffi',
-    'psutil',
```
smkuls: nit: it is a common practice to leave the comma at the end of the last element so when you add something new later in a separate PR it only shows one line as diff.

smkuls (commented):
I took a brief look to verify sanity. The PR is in a great shape. Great job!

setup.py
rahulgovind: Oh! That makes perfect sense. I'll add it back. I removed it by mistake when I was trying to add Cython to the build requirements before.

.travis.compose.yml
rahulgovind: Will do.

honeybadgermpc/batch_reconstruction.py
rahulgovind: Looks like I removed it somehow. I added the benchmark logging back now.

.travis.compose.yml
rahulgovind: Referenced this issue in #180

PR #181, opened 2019-02-18
Shuhong Gao's Reed-Solomon decoding algorithm and refactoring of reconstruction

rahulgovind:
This is just a sort of proof-of-concept implementation of Shuhong Gao's decoding algorithm (http://www.math.clemson.edu/~sgao/papers/RS.pdf) to replace the Welch-Berlekamp algorithm HBMPC is using right now. 

Some advantages of the algorithm
- Can integrate FFT
- Better time complexity
   Berlekamp-Welch is `O(N^3)`. The paper does not note the overall time complexity. Although it looks like it would be `O(N^2)` without FFT and `O(N poly(log N))` with FFT

I ran some benchmarks using  https://gist.github.com/rahulgovind/02955d65b341e158bcbaf3001ddb5b5fdc26c23b and compared the performance of the new decoding implementation vs the old implementation. Here are the results -
img

Note that part of the improvement is due to using NTL/Cython vs pure python. However, part of the speedup is due to the better time complexity. The O(N^2) complexity of Gao's algorithm without FFT is evident in the second graph as the speedup increases linearly with N. Also, the O(N poly(log N))) complexity of Gao's algorithm with FFT is evident in the third graph.


amiller:
It's great to see a clear performance improvement and n^3 to n^2 could easily make the difference of something we'd want to show off in out implementation (itll determine the worst case latency especially at very large N). 

IMO it's useful to still have the simple pure python WB algorithm around and tested as an alternative, but this already looks like it's a clear better path. Awesome work

docker-compose.yml
```diff
@@ -25,6 +25,7 @@ services:
       - ./pairing/benches:/usr/src/HoneyBadgerMPC/pairing/benches
       - ./pairing/Cargo.toml:/usr/src/HoneyBadgerMPC/pairing/Cargo.toml
       - ./pairing/setup.py:/usr/src/HoneyBadgerMPC/pairing/setup.py
+      - /usr/src/HoneyBadgerMPC/honeybadgermpc/ntl  # Directory _not_ mounted from host
```
smkuls: why is this showing as a diff? wasn't this included in your previous change?

amiller (approved):

smkuls (changes requested):
Gao's algo, lagrange interpolation, FFT based interpolation, and wb_interpolate need to be moved to a config so that things can be easily changed without having to rebuild the code every time.

docker-compose.yml
smkuls: The fix for this is already in. Closing this.

amiller (commented):
So the problem is just that wb_interpolate is expecting a degree-1 or greater polynomial, and all 0s is treated as degree-0? I think if this is legitimately a special case, we can just check for it

rahulgovind:
> So the problem is just that wb_interpolate is expecting a degree-1 or greater polynomial, and all 0s is treated as degree-0? I think if this is legitimately a special case, we can just check for it

It does look like  `wb_interpolate` is expecting a degree-0 or greater polynomial. I also think that `wb_interpolate` will not be able to handle cases where the evaluation point is `0`. However, this situation will most probably not occur in our use cases.

honeybadgermpc/batch_reconstruction.py
```diff
@@ -220,16 +220,75 @@ def merge_lists(lists):
         data = await wait_for(data_receivers, n_available)
         data = tuple(with_at_most_non_none(data, n_available))
         logging.debug(f'data R1: {data} nAvailable: {n_available}')
-        data = tuple([None if d is None else tuple(map(field, d)) for d in data])
-        reconstructed_poly = attempt_reconstruct_batch(data, field, n, t, point)
+        if config is not None and config.decoding_algorithm == "welch-berlekamp":
+            logging.info("[BatchReconstruction] Using welch-berlekamp decoding")
```
smkuls: It'll be better to change this to `debug` or else this is going to pollute the logs.

honeybadgermpc/batch_reconstruction.py
```diff
@@ -220,16 +220,75 @@ def merge_lists(lists):
...
+            data = tuple([None if d is None else tuple(map(field, d)) for d in data])
+            reconstructed_poly = attempt_reconstruct_batch(data, field, n, t, point)
```
smkuls: Can you rename this and gao's caller methods to include the algo's name? It is confusing to read. If I decide to use either on my own for some other purpose, I should be able to just look at the method name and call the appropriate algorithm.

smkuls: Follow a consistent naming convention.

honeybadgermpc/batch_reconstruction.py
```diff
@@ -220,16 +220,75 @@ def merge_lists(lists):
...
             # TODO: return partial success, so we can skip these next turn
             continue
 
-        return reconstructed_poly
+        return merge_lists(reconstructed_poly)
     return None
 
 
+def robust_batch_interpolate(data, field, point, n, t, n_available):
+    logging.debug(f"Attempting robust interpolation with {n_available} points")
```
smkuls: I know other places in the code have this, but we should use `logging.debug("Message: %d", n_available)`. This has the benefit that the format string is not invoked if you are not logging the `DEBUG` level.

honeybadgermpc/batch_reconstruction.py
```diff
@@ -220,16 +220,75 @@ def merge_lists(lists):
...
+    n_chunks = max(len(d) for d in data if d is not None)
+    x = []  # Evaluation points
+    z = []  # Index of parties we are considering here
```
smkuls: Avoid single letter variable names. You should never require a comment to describe the intention of a variable.

honeybadgermpc/batch_reconstruction.py
```diff
@@ -220,16 +220,75 @@ def merge_lists(lists):
...
+            break
+
+    assert len(x) == n_available
+
+    p = field.modulus
+
+    polynomial_solutions = []
+    for chunk in chunks:
+        if point.use_fft:
+            logging.debug("Using Gao's algorithm with FFT")
```
smkuls: nit: this can be moved before `if` -> `logging.debug("Using Gao's algo. FFT = %s", point.use_fft)`

honeybadgermpc/batch_reconstruction.py
```diff
@@ -220,16 +220,75 @@ def merge_lists(lists):
...
+            polynomial_solution, err_polynomial = gao_interpolate(x, chunk, t + 1, p,
+                                                                  z, point.omega.value,
+                                                                  point.order,
+                                                                  use_fft=True)
+        else:
+            logging.debug("Using Gao's algorithm without FFT")
+            polynomial_solution, err_polynomial = gao_interpolate(x, chunk, t + 1, p)
+
+        if polynomial_solution is None:
+            logging.debug("Robust interpolation failed")
```
smkuls: Is this an error? Should you raise an `exception` here? Use, `logging.error` or `logging.warning` for failure/warning/error logs.

honeybadgermpc/batch_reconstruction.py
```diff
@@ -220,16 +220,75 @@ def merge_lists(lists):
...
+            return None
+
+        errors_detected = len(err_polynomial) - 1  # Num errors = deg of error polynomial
+        num_not_erroneous = n_available - errors_detected
+        if num_not_erroneous < 2 * t + 1:
+            # Need 2t + 1 parties to agree on same polynomial
+            logging.debug(f"Robust interpolation partially failed."
```
smkuls: Same comment here, use -> %s, %d in the logging string instead of `f{}` like strings.

honeybadgermpc/config.py
```diff
@@ -26,8 +25,27 @@ class ConfigVars(object):
...
+    def __init__(self, induce_faults, decoding_algorithm):
         self.induce_faults = induce_faults
+        self.decoding_algorithm = decoding_algorithm
+
+    @classmethod
+    def default(cls):
+        return cls(induce_faults=False, decoding_algorithm="gao")
+
+    @classmethod
+    def from_json(cls, json_config):
```
smkuls, line 38: Neat!

tests/test_ntl.py
```diff
@@ -1,5 +1,6 @@
 from honeybadgermpc.ntl.helpers import interpolate, batch_vandermonde_interpolate, \
-    batch_vandermonde_evaluate, fft, fft_interpolate, fft_batch_interpolate
+    batch_vandermonde_evaluate, fft, fft_interpolate, fft_batch_interpolate, \
```
smkuls: Consider renaming these methods to maintain consistency. `lagrange_evaluate` `fft_evaluate` etc.

tests/test_ntl.py
```diff
@@ -117,3 +118,146 @@ def test_fft_batch_interpolate(galois_field, galois_field_roots):
...
+    x = list(range(n))
+    encoded = [sum(int_msg[j] * pow(x[i], j, p) for j in range(k)) % p
+               for i in range(n)]
+
+    # Check decoding with no errors
+    decoded, _ = gao_interpolate(x, encoded, k, p)
+    assert (decoded == int_msg)
+
+    # Corrupt with maximum number of erasures:
+    cmax = n - 2 * t - 1
```
smkuls, line 154: Aren't these better served as separate test cases? I would advice against checking multiple functionalities in one test since the first failure masks all others.

smkuls: Same comment for the test cases below.

tests/test_wb_interpolate.py
```diff
@@ -40,16 +40,57 @@ def test_decoding():
...
+    p = 53  # prime
+    t = k - 1  # degree of polynomial
+
+    enc, dec, _ = make_encoder_decoder(n, k, p)
+    encoded = enc(int_msg)
+    # print("plain message is: %r" % (integerMessage,))
+    # print("encoded message is: %r" % (encoded,))  # cleaner output
+
+    # Check decoding with no errors
+    with pytest.raises(IndexError):
```
smkuls, line 52: Nice! Add a link to the bug here so we know which bug it is in case someone fixes this later.

smkuls (changes requested):
The code can definitely be structured better. 

You should create one `class` per algorithm. All algorithms should be derived from an abstract base class whose methods tell us what all functionalities need to be supported if someone wants to add a new algorithm later (python doesn't have interfaces sadly).

You should only pass the base class object around which points to one of the algorithms. This base class should be instantiated in a factory which should be the only place reading the config in the whole code. Also, if the code is growing too much in one file, move it to separate files and encapsulate within a folder.

I am more that open to suggestions for better structuring of this code. Let me know if you want to discuss more.

* Rename the PR appropriately.
* Remove the word POC from the commit. I am okay with multiple commits.
* Update the sample config to include the values for all the options that you have added. This will make it much easier for someone to figure out what all is supported.
* Add a file called `test_batch_reconstruction_modes.py`. Write tests cases which run complete MPC programs using the TaskProgramRunner and take in the config. You can use `test_mixins.py` as a reference.
* Andrew had commented something on slack that some functionality is missing in some versions. Can you include those? Having an abstract base class will make sure that we don't miss these later on.

smkuls:
I also noticed that there's no option to pick between lagrange/fft. Can you add that as well?

Dockerfile
```diff
@@ -33,6 +33,7 @@ RUN ln -sf bash /bin/sh
 RUN git clone https://github.com/JHUISI/charm.git
 RUN cd charm && git reset --hard be9587ccdd4d61c591fb50728ebf2a4690a2064f && ./configure.sh && make install
 
+
```
smkuls: nit: why? avoid touching files which aren't related to the PR, bloats up the diff and unnecessarily adds your name to `git blame`

honeybadgermpc/ntl/helpers.pyx
```diff
@@ -318,3 +338,60 @@ cpdef SetNTLNumThreads(x):
...
+
+        for i in range(res_vec.length()):
+            result[i] = int(ccrepr(res_vec[i]))
+        for i in range(err_vec.length()):
+            error_poly[i] = int(ccrepr(err_vec[i]))
+        return result, error_poly
+
+    return None, None
+
+def sqrt_mod(a, n):
```
amiller, line 366: Any tests for this?

tests/test_ntl.py
amiller: still not fully consistent... consider changing `fft_batch_interpolate` to `batch_fft_interpolate`

honeybadgermpc/ntl/helpers.h
```diff
@@ -208,4 +208,144 @@ void fnt_decode_step2(vec_ZZ_p &P_coeffs, ZZ_pX &A, vec_ZZ_p &Ad_evals,
     for (int i=0; i < k; i++) {
         P_coeffs[i] = coeff(P, i);
     }
```
amiller, line 210: If anything... `helpers.h` has a too generic name. It's actually chock full of some interesting algorithms of independent interest implemented using NTL. What if we rename it as `rsdecoder_impl.h` or something like that, just so it is a bit more descriptive what's in there?

amiller (approved):
This looks solid to me - I'm marking it approved but wait to see if smkuls acknowledges his requested changes are met

honeybadgermpc/ntl/helpers.pyx
rahulgovind: Added tests for sqrt_mod in d96fcc92a393e7e9e1ad09ff016a675ffbc53f6c

Dockerfile
smkuls: nit: avoid touching unnecessary lines to introduce new lines as well especially in bulky PRs

honeybadgermpc/reed_solomon.py
```diff
@@ -0,0 +1,381 @@
+from honeybadgermpc.ntl.helpers import vandermonde_batch_evaluate, \
+    vandermonde_batch_interpolate
+from honeybadgermpc.ntl.helpers import gao_interpolate
+from honeybadgermpc.ntl.helpers import fft, fft_interpolate, fft_batch_interpolate
+from honeybadgermpc.wb_interpolate import make_encoder_decoder
```
smkuls: what is this `make_encoder_decoder`? If this is related to only `wb_interpolate` can you please rename this to something appropriate to avoid confusion with the encoder decoders that you have added?

honeybadgermpc/reed_solomon.py
```diff
@@ -0,0 +1,381 @@
+from honeybadgermpc.ntl.helpers import vandermonde_batch_evaluate, \
+    vandermonde_batch_interpolate
+from honeybadgermpc.ntl.helpers import gao_interpolate
+from honeybadgermpc.ntl.helpers import fft, fft_interpolate, fft_batch_interpolate
```
smkuls, line 4: rename `fft` to `fft_evaluate`?

honeybadgermpc/reed_solomon.py
```diff
@@ -0,0 +1,381 @@
...
+from honeybadgermpc.wb_interpolate import make_encoder_decoder
+
+
+class Encoder(object):
+    def encode(self, data):
+        if type(data[0]) is list or type(data[0]) is tuple:
```
smkuls: nit: `type(data[0]) in [list, tuple]`

honeybadgermpc/reed_solomon.py
```diff
@@ -0,0 +1,381 @@
...
+            return self.encode_batch(data)
+        else:
```
smkuls: Please remove. This `else` is redundant. This is one extra line of code which is not needed.

honeybadgermpc/reed_solomon.py
```diff
@@ -0,0 +1,381 @@
...
+        :return: Encoded values
+        """
+        raise NotImplementedError
+
+
+class Decoder(object):
+    def decode(self, z, encoded):
+        if type(encoded[0]) is list:
+            return self.decode_batch(z, encoded)
+        else:
```
smkuls: same here, redundant `else`

honeybadgermpc/reed_solomon.py
```diff
@@ -0,0 +1,381 @@
...
+        """
+        :type data: list of list of integers
+        :return: Encoded values
+        """
+        raise NotImplementedError
+
+
+class Decoder(object):
+    def decode(self, z, encoded):
+        if type(encoded[0]) is list:
```
smkuls: do you not want to check for a `tuple` here as well?

honeybadgermpc/reed_solomon.py
```diff
@@ -0,0 +1,381 @@
+from honeybadgermpc.ntl.helpers import vandermonde_batch_evaluate, \
+    vandermonde_batch_interpolate
+from honeybadgermpc.ntl.helpers import gao_interpolate
+from honeybadgermpc.ntl.helpers import fft, fft_interpolate, fft_batch_interpolate
+from honeybadgermpc.wb_interpolate import make_encoder_decoder
+
+
+class Encoder(object):
```
smkuls, line 10: Since these are base classes, can you please add a short docstring so as to what each one does, only for `Encoder` `Decoder`.

honeybadgermpc/reed_solomon.py
```diff
@@ -0,0 +1,381 @@
...
+            if decoded is not None:
+                errors = []
+                if len(error_poly) > 1:
+                    err_eval = fft(error_poly, self.point.omega.value,
+                                   self.modulus, self.point.order)[:self.point.n]
+                    errors = [i for i in range(self.point.n)
+                              if err_eval[i] == 0]
+                return decoded, errors
+            return None, None
+        else:
```
smkuls, line 160: again, not needed

honeybadgermpc/reed_solomon.py
```diff
@@ -0,0 +1,381 @@
...
+    pass
+
+
+class IncrementalDecoder(object):
+    """
+    Incremental decoder helps process new data incrementally and aims to make the
+    case where no error is present extremely fast.
+
+    1) Validate that the data is indeed correct.
+    2) If at least d + 1 points are available, then we can use a non-robust decoder
```
smkuls: what is `d`? is `d` degree or same as `t`, can you please clarify this in the comments?

honeybadgermpc/reed_solomon.py
```diff
@@ -0,0 +1,381 @@
...
+            poly_eval = vandermonde_batch_evaluate(x, [[c.value for c in coeffs]],
+                                                   self.modulus)[0]
+            errors = [i for i in range(self.point.n)
+                      if enc_extended[i] is not None and
+                      enc_extended[i].value != poly_eval[i]]
+            return coeffs, errors
+        return None, None
+
+
+class DecodeValidationError(Exception):
```
smkuls: derive this from `HoneyBadgerMPCError` in `exception.py`

honeybadgermpc/reed_solomon.py
```diff
@@ -0,0 +1,381 @@
...
+                    all_match = False
+                    break
+
+            if all_match is False:
+                # Guess was incorrect
+                self._guess_decoded = None
+                self._guess_encoded = None
+                self._optimistic = False
+
+        if success and len(self._available_points) >= self._min_points_required():
```
smkuls, line 300: you never reassign `success`, it is always `True` at this point..

honeybadgermpc/reed_solomon.py
```diff
@@ -0,0 +1,381 @@
...
+            # Decoding successful
+            self._result = self._guess_decoded
+
+        return success
+
+    def _robust_update(self):
+        while self._num_decoded < self.batch_size:
+            print(f"len(z): {len(self._z)} len(data): {len(self._available_data[0])}")
```
smkuls: Change this to `logging.info`

honeybadgermpc/reed_solomon.py
```diff
@@ -0,0 +1,381 @@
...
+        self.degree = degree
+        self.batch_size = batch_size
+        self.max_errors = max_errors
+        self.validator = validator
+
+        self._confirmed_errors = set()
+        if confirmed_errors is not None:
+            self._confirmed_errors = confirmed_errors
+        self._available_points = set()
+        self._z = []
```
smkuls, line 246: what is `z`? are you following some convention defined in an algo that you have ported from the paper, if not, can you please rename this to something more descriptive

honeybadgermpc/reed_solomon.py
smkuls: or maybe `logging.debug` since you are doing this in a loop

honeybadgermpc/reed_solomon.py
```diff
@@ -0,0 +1,381 @@
...
+            self._result = self._partial_result
+
+    # Public API
+    def add(self, idx, data):
+        if self.done() or idx in self._available_points or \
+                idx in self._confirmed_errors:
+            return
+
+        if self._validate(data) is False:
+            return
```
smkuls: if validation fails, do we want to raise an error?

honeybadgermpc/reed_solomon.py
```diff
@@ -0,0 +1,381 @@
...
+        return self._result is not None
+
+    def get_results(self):
+        if self._result is not None:
+            return self._result, self._confirmed_errors
+        return None, None
+
+
+def get_rs_encoder(point, algorithm='vandermonde'):
+    if algorithm == 'vandermonde':
```
smkuls: Since you are using this string in multiple files, I would strongly recommend you to define this and the other algos as constants.

Consider a scenario, that there's a typo in one of the string and now you have to do a CTRL-F in the entire codebase, whereas if you had it as a constant then you have to edit in one place.

honeybadgermpc/reed_solomon.py
```diff
@@ -0,0 +1,381 @@
...
+        return VandermondeEncoder(point)
+    elif algorithm == 'fft':
+        return FFTEncoder(point)
+    raise ValueError("Incorrect algorithm. "
+                     "Supported algorithms are [vandermonde, fft]")
+
+
+def get_rs_decoder(point, algorithm='vandermonde'):
```
smkuls: Since you have these two methods as separate, I can make the mistake of using an FFTEncoder with a Vandermonde one or if I put it differently you are allowing me to use one with the other

Consider adding a factory:

```
class EncoderDecoderFactory():

     @staticmethod
    def get(self, point, algorithm=None):
        if algorithm is None:
            algorithm = "default"
        if algorithm == Algorithm.FFT:
            return FFTEncoder(point), FFTDecoder(point)
        # and so on
```

honeybadgermpc/reed_solomon.py
```diff
@@ -0,0 +1,381 @@
...
+    if algorithm == 'vandermonde':
+        return VandermondeDecoder(point)
+    elif algorithm == 'fft':
+        return FFTDecoder(point)
+
+    raise ValueError("Invalid algorithm. "
```
smkuls: This list should be defined as a `constant`. If tomorrow, I have to add a new algo I don't want to do a CTRL-F in the code

honeybadgermpc/reed_solomon.py
```diff
@@ -0,0 +1,381 @@
...
+                     "Supported algorithms are [vandermonde, fft]")
+
+
+def get_rs_robust_decoder(t, point, algorithm='gao'):
+    if algorithm == 'gao':
+        return GaoRobustDecoder(t, point)
+    raise ValueError("Invalid algorithm. "
```
smkuls: Same here

honeybadgermpc/reed_solomon.py
```diff
@@ -0,0 +1,381 @@
...
+    if algorithm == 'vandermonde':
+        return VandermondeDecoder(point)
+    elif algorithm == 'fft':
+        return FFTDecoder(point)
+
+    raise ValueError("Invalid algorithm. "
+                     "Supported algorithms are [vandermonde, fft]")
+
+
+def get_rs_robust_decoder(t, point, algorithm='gao'):
```
smkuls: what happened to wb_decoder?

honeybadgermpc/robust_reconstruction.py
```diff
@@ -1,66 +1,29 @@
...
-            return p, failures
-        except ValueError as e:
-            logging.debug(f'ValueError: {e}')
-            if str(e) in ("Wrong degree", "no divisors found"):
-                continue
-            else:
-                raise e
-    assert False, "shouldn't reach the end here"
+    use_fft = point.use_fft
+    enc = get_rs_encoder(point, 'fft' if use_fft else 'vandermonde')
```
smkuls: I would advise against doing an `if` `else` here. Shouldn't this pickup what is defined in the config? 

Unless I am missing something, with the changes that you have made, I don't see how I can run an MPC program both from **tasks** and **processes** and _easily switch_ between using either `fft` or `vandermonde`.

honeybadgermpc/robust_reconstruction.py
```diff
@@ -1,66 +1,29 @@
...
+    dec = get_rs_decoder(point, 'fft' if use_fft else 'vandermonde')
+    robust_dec = get_rs_robust_decoder(t, point)
+    online_decoder = IncrementalDecoder(enc, dec, robust_dec, t, 1, t)
```
smkuls: be consistent, `online_dec`

tests/test_ntl.py
```diff
@@ -1,5 +1,8 @@
-from honeybadgermpc.ntl.helpers import interpolate, batch_vandermonde_interpolate, \
-    batch_vandermonde_evaluate, fft, fft_interpolate, fft_batch_interpolate
+from honeybadgermpc.ntl.helpers import lagrange_interpolate, \
```
smkuls, line 1: nit: it looks neater if you do imports in multiple lines but I don't know if that follows the convention

smkuls: You can club vandermonde ones in one line, fft ones in another line etc.

tests/test_wb_interpolate.py
```diff
@@ -1,5 +1,5 @@
 import random
```
smkuls, line 1: the tests are spread across these files...
test_ntl, test_reed_solomon and this file, should we move this test to reed_solomon since you already have the gao test there, or do you want to move gao test here and then call this `test_robust_decoder.py`?

tests/test_reed_solomon.py
```diff
@@ -0,0 +1,124 @@
+import pytest
```
smkuls, line 1: No tests for IncrementalDecoder? You can take it in a separate PR if you'd like since this is growing large and we should merge it before this blows up.

smkuls (changes requested):
This is some badass refactoring!! I really like the new structure of the code. Great work!

I am requesting a few changes in the way the config values are used and some other minor changes.

Just so we have a track, things which still need to be added later and can be taken up in a separate PR:
1. Update `polynomial.py` to use code from NTL
2. Remove all traces of `fft` implementation from `polynomial.py`
3. Tests for IncrementalDecoder

tests/test_reed_solomon.py
amiller: Good catch on missing tests, we shouldn't add new content without minimal test coverage. I can't think of an easy way to recommend breaking this into multiple PRs, do you have a suggestion?

smkuls: I would actually suggest @rahulgovind to take the test up immediately after this PR and not add any more new code with this PR. There is a lot of new code in this and we shouldn't add more..

amiller: OK so let's merge this after the cosmetic fixes (no unnecessary file/line changes etc.), even without the tests for `IncrementalDecoder`, with an issue to add such tests in a subsequent PR

honeybadgermpc/reed_solomon.py
rahulgovind: Can't think of a good name in this case. It also serves the same function but has a different interface. I can mark it as a sort of private/obsolete API by prepending it with underscores. `make_encoder_decoder` -> `_make_encoder_decoder`.

honeybadgermpc/reed_solomon.py
rahulgovind: Semantics do not match very well. I don't think this is a good idea. `fft` results is equivalent to a sort of `fft_evaluate` only if `n` is a power of two. 

Also, the encoder / decoder interface is supposed to replace most of the `xxx_evaluate` stuff.

honeybadgermpc/reed_solomon.py
smkuls: If you make it private are you still going to keep the import it in this file?

honeybadgermpc/reed_solomon.py
rahulgovind: Yup. That's definitely better and easier to read.

honeybadgermpc/reed_solomon.py
rahulgovind: Fixed

honeybadgermpc/reed_solomon.py
rahulgovind: Yes. Not sure if it's a great idea but I've seen it in other places in the Python standard library.

smkuls: Lets rename it to `make_wb_encoder_decoder` to avoid the confusion

tests/test_wb_interpolate.py
rahulgovind: Not sure if this is a good idea. `test_ntl.py` tests lower level NTL-based implementations and interface to NTL code which seems like a good place to keep tests for `gao_interpolate`. 

`wb_interpolate` has a very different interface from the rest of the code and I think it's better to keep its code and tests separate.

honeybadgermpc/robust_reconstruction.py
rahulgovind: I'll need to think about this. FFT is not specific to reconstruction and needs to be set for other parts of HBMPC too. Therefore, ReconstructionConfig might not be the best place to keep it.

For now, `batch_reconstruct` needs to be called with `use_fft=True` for FFT, and `robust_reconstruct` needs to be called with `point` initialized with `use_fft=True` for FFT.

rahulgovind:
03c0088d2ffb1910752765aaa507e344298640a5 fixes a bug in `wb_interpolate` where interpolation would succeed even in cases where it should have actually failed. (It interpolated a higher degree polynomial than specified).

For example, with `t=1` and `n=4`, calling `decode([1, 4, 9, None])` should fail since the interpolation is a degree-2 polynomial (x^2). However, this previously succeeded since `max_errors` was set to `n - c - 2 * t - 1` where `c` is the number of erasures. Since `max_errors=0` for the given example, `decode` would mistakenly interpolate using non-robust interpolate on the three given points, and end up generating the degree-2 polynomial instead.

This was previously caught by https://github.com/initc3/HoneyBadgerMPC/blob/8ee6a3b56be244e437346556e85b988ac352af55/honeybadgermpc/robust_reconstruction.py#L31.

smkuls (approved):

smkuls:
I am approving this PR. I am still not convinced with the argument to not have FFT in config but that can be taken up separately. This PR has blown up way too much.

PR #184, opened 2019-02-21
Port BinaryAgreement algo form HoneyBadgerBFT

smkuls:
* This port required moving the code from gevent to asyncio
* As part of this move, all code related to protocols which have been
lifted from HoneyBadgerBFT have been moved to a protocols directory
* Remove global `TERMINATOR` object from TaskPool


amiller (approved):
Looks good. I reviewed just by inspection, not side-by-side with the gevent version. The unit tests and entire protocol test look as expected. Cool!

PR #185, opened 2019-02-22
Make BinaryAgreement and CommonCoin run on multiple processes

smkuls:
* Add a broadcast method to ProcessProgramRunner
* Add script to run commoncoin.py across multiple processes
* Modify boldyreva.py to add pickle and unpickle methods to private and
public keys, also added methods to compare public and private keys.
Added a helper method to generate serialized keys to be used with the
multiple process setup.


honeybadgermpc/ipc.py
```diff
@@ -222,26 +222,32 @@ def get_send_and_recv(self, sid):
...
                 return (i, o)
             return _recv
 
-        return make_send(self.nodeid, sid), make_recv(self.nodeid, sid)
+        send, broadcast = make_send(self.nodeid, sid)
+        return send, make_recv(self.nodeid, sid), broadcast
 
     def add(self, sid, program, **kwargs):
-        send, recv = self.get_send_and_recv(sid)
+        send, recv, _ = self.get_send_and_recv(sid)
```
amiller: I don't think we should modify `get_send_and_recv` to return `broadcast`, that can just be locally defined by whoever wants to use it. It's really better called "multicast" anyway.

honeybadgermpc/protocols/commoncoin.py
```diff
@@ -92,7 +93,85 @@ def hash(x):
...
+
+    if not pbk_string:
+        raise ConfigurationError('Environment variable `HBMPC_PB_KEY` must be set'
+                                 ' or a config file must be given as first argument.')
+
+    config_dict = load_config(configfile)
+    n = config_dict['N']
+    t = config_dict['t']
+    k = config_dict['k']
+    pbk = pickle.loads(codecs.decode(pbk_string.encode(), "base64"))
```
amiller: this looks weird, we seem to be encoding and decoding in the same line, base64 and pickle are involved, can this be simplified?

amiller (approved):
This looks good, I only noticed a couple cosmetic things to ask about

honeybadgermpc/protocols/commoncoin.py
smkuls: I will change this to `pickle.loads(base64.b64decode(pbk_string))`. The base64 encoding makes it convenient to store the pickled string in the shell script. Otherwise, it is a pain to escape the characters within the pickled string to store them in the shell script.

honeybadgermpc/ipc.py
smkuls: Changed.

honeybadgermpc/protocols/commoncoin.py
smkuls: This has also been updated.

amiller:
Looks good!

PR #187, opened 2019-02-25
add __xor__ in mpc.py

hanyunx:


amiller:
I don't mean to discourage making a PR, but I think it will not make sense to add "xor" as an operation on fields/shares. It makes sense for binary extension fields `GF(2^k)` but not for `GF(p)` for a large prime `p` like we're using here. Xor isn't a linear operation here so it won't work if parties just compute it locally.

PR #192, opened 2019-02-28
Run HoneyBadgerMPC tests only and skip charm tests

smkuls:
After adding `charm` as a dependency `pytest -v` also triggers charm's
test cases which prolong the CI time. This PR skips the charm's tests
and runs only the tests written within the `tests/` directory.


smkuls:
Found a better way to fix this.

PR #193, opened 2019-02-28
Skip charm tests and run only HoneyBadgerMPC tests

smkuls:
After adding `charm` as a dependency `pytest -v` also triggers `charm's` test cases which causes the CI to take longer thereby increasing the PR time. This change prevents `charm's` test cases to be run with `pytest -v`.


amiller (approved):
ok

PR #197, opened 2019-03-01
Refactor config and scripts

smkuls:
**Remove code related to containers in the same n/w**

* This configuration is never used. We don't create multiple containers on
the same network with different hostnames so we don't need this code. We
instead create multiple containers which interact with each other via
socket communication.

**Refactoring config and scripts**

* Added a global `HbmpcConfig` which is a per node config and contains
all the parameters which are required for running an MPC application.
* This config is loaded in `__init__.py` and thus can be used directly in
any part of the code.
* Added`-d` and `-f` flags to indicate distributed mode and the config
file path. By default all programs run in non-distributed setting. As a
result of adding the ArgParser, the user can now see help by running
the command with a `-h` tag.
* Added a generic script within `scripts/mpc/` folder which launches any
MPC program. This script takes the python file which needs to be run and
the config file prefix. The local config files have also been added
within the `conf/local/`folder. Each node has its own config file.
* The previous step has been repeated for the AVSS code as well since
that needs a 5 processes instead of 4.
* Made /bin/sh point to /bin/bash within the container since the `split`
command doesn't work with `sh` in the `launch_mpc.sh` script.
* The scripts folder has also been cleaned up by deleting all unncessary
scripts and moving the AWS `download_input.sh` script to the AWS folder.


amiller (approved):
Excellent, solid refactoring of config system...

PR #198, opened 2019-03-01
Add "induce_faults" to the config

smkuls:
**Move config from cfgparse to json**

Json gives much more flexibility. Most importantly it supports
subsections which are not supported by cfgparse. It is also easier to
serialize/deserialize json objects directly to Python objects.

**Add "induce_faults" to the config**

Add a config value to induce faults during reconsutrction. This would
cause the code to go the non optimistic path. This will be helpful
during benchmarking since this would prevent us from having to
rebuild/redeploy the code again and again.


amiller (changes requested):
So.... the only thing I'd say is that "induce_faults" is pretty low granularity. If we want to be able to exercise failure handlers of different kinds we may have to induce different kinds of faults.

smkuls:
Yes I agree, that's why I have added it specifically under `reconstruction` section in the config. We will need different kinds of faults and we'd like to not have faults in some part vs the other. So for that, I'd suggest introducing a new section in the config and adding `induce_faults` to that section and modifying the code appropriately to introduce the faults for that particular section of the code.

amiller (approved):

PR #199, opened 2019-03-02
Revert "More feature to betterpairing added and some refactoring perf…

nqyy:
…ormed "

This reverts commit b65e7a414238c2bedfb93d302c26d5546b561a51.

The joint commit of @tyurek and me broke the ``launch_hbavsslight.sh`` script. Details are specifed in https://github.com/initc3/HoneyBadgerMPC/issues/191 . We might need Tom to take a look at the issue. Right now, he is pretty busy so our best approach now is to revert it and get it re-commit later.

Issue #191, opened 2019-02-28
launch_hbavsslight broken after #177


smkuls:
#177 breaks the `launch_hbavsslight.sh` script. The script seems to work fine when this PR is reverted.

smkuls:
@chitianhao

nqyy:
The changes in betterpairing.py in #177 caused this issue. To be specific, in secretshare_hbavsslight.py line 347 to 361:
```
            if await _run_in_thread(self.pc.verify_eval, self.commit,
                                    self.pid+1, self.share, self.witness):
                decryption_time = str(os.times()[4] - decrypt_start_time[4])
                logging.info("decryption time : " + decryption_time)
                self.benchmarkLogger.info(
                    "decryption time :  " + decryption_time)

                self.send_ok_msgs()
                self.sendrecs = True
            else:
                logging.info("verifyeval failed")
                self.send_implicate_msgs()
            while not self.queues["hbavss"].empty():
                (i, o) = self.queues["hbavss"].get_nowait()
                await self.receive_msg(i, o)
````
After the changing, it goes to the "else" which contains an unimplemented method ``send_implicate_msgs()``. Will need @tyurek to take a look.

smkuls:
Another important thing: Why did the tests not catch this? We should first modify the tests to make sure this is caught by them and then propose a fix.

amiller:
changing `asyncio.wait` to `asyncio.gather`  at 
https://github.com/initc3/HoneyBadgerMPC/blob/a0abe5a1ab8094c73fc50919c3797988b5a71d56/tests/test_hbavss.py#L33
makes the test show this error

smkuls:
Closed by #201

PR #200, opened 2019-03-02
Revert because of #191

smkuls:
This reverts commit b65e7a414238c2bedfb93d302c26d5546b561a51.

The joint commit of @tyurek and @chitianhao broke the launch_hbavsslight.sh script.
Details are specifed in #191. We need Tom to take a look at the issue.
Right now, he is pretty busy so our best approach now is to revert it and get it
re-commit later.

This is same as #199 but it avoids the extra merge commit. I am creating this PR on my own since @chitianhao is busy dealing with AVSS stuff.


smkuls:
No need for this. This has been fixed in #201.

PR #201, opened 2019-03-02
Fix pairing

amiller:
This fixes two problems:
- first, the test in `test_hbavss.py` uses `asyncio.wait` rather than `asyncio.gather`, as a result the test passes even when the hbavss routine fails, which masked the error checked in with #177 
- second, the problem checked in with #177 resulted in `G1.one()` and `G2.one()` not actually returning an identity element


smkuls (approved):
Looks good to me!

nqyy:
This is great! Thanks @amiller

PR #203, opened 2019-03-03
Some fixes to AWS code

smkuls:
* Update AWS code to use the new json config.
* Add the `induce_faults` option to `aws-config.json` so the code can
actually be benchmarked in failure mode without rebuild/redeploy.
* Upload files/configs in parallel by default.
* Measure time taken to execute different operations within the AWS code
so when running the experiments later in case something is extremely
slow then that step can be optimized.
* Optimize the preprocessing code to not generate power in every
iteration instead generate the powers once. This reduces the time taken
to generate random powers to 0.33x.
* Replace all print statements with logging module. Since the logger logs
the current time, it also makes it helpful while running the experiments
to estimate how much time has elapsed.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -8,3 +8,4 @@ Dockerfile
 **/*.pyc
 sharedata/**
 aws/
+conf/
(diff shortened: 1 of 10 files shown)


amiller (approved):
All of this looks very good as AWS cleaning up and optimizations.
Let's merge this now, though optimizing preprocessing generation using vandermonde and/or fft will be better still. We should do that too before handing off to run the benchmarking IMO

PR #204, opened 2019-03-03
Optimize preprocessing and fix AWS __init__.py

smkuls:
**Add `benchmark` folder to aws/__init__.py**
This is needed to be able to load the logging config without any errors.

**Optimize evaluation of polynomials by using ntl**
Switch to vandermonde evaluate exposed via NTL for evaluating the
polynomial at a range of values. This would bring down the amount of
time required to create the preprocessing files.

### DIFF
--- a/aws/__init__.py
+++ b/aws/__init__.py
@@ -1,8 +1,10 @@
 import logging.config
 import yaml
+import os
 
 
 with open('honeybadgermpc/logging.yaml', 'r') as f:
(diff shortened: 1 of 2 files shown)


amiller (approved):
Works great. I ran it myself with a custom script in `scripts/benchmark_preprocessing.py` which just ran `PreProcessedElements().generate_rands(4096, 61, 20)` and confirmed the improvement in running time (we should figure out a better benchmarking approach tho)

PR #205, opened 2019-03-04
 Add a way to benchmark existing code

smkuls:
* Added `pytest-benchmark` as a dependency, this allows writing test
cases which basically benchmark the code.
* The documentation can be read here: https://pytest-benchmark.readthedocs.io
* It spits out a report after running the tests which describes the
amount of time it took to benchmark a function.
* There is also an option to generate a histogram and compare with
previous results.

image

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -129,3 +129,6 @@ aws/aws-config.json
 
 *~
 sharedata/
+
+# Graphs
+*.svg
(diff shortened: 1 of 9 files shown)


amiller (approved):
Works great, this use of pytest benchmark will make it easy to add benchmarks for any of our compute tasks, should be very helpful in keeping track of optimizations.

PR #206, opened 2019-03-04
[DO NOT MERGE] Use FFT for batch_reconstruction by default

smkuls:
THIS COMMIT SHOULD NOT GO TO DEV AS IS.

This is just a temporary way to unblock for benchmarking. This commit
evaluates the polynomial at powers of omega instead of [1, n] so that
FFT can be used for batch reconstruction.

This code will break share.open() because `robust_reconstruct` expects
the points at [1, n] and not at [omega^0, omega^(n-1)].


amiller:
Can we discard this?

smkuls:
Yes.

PR #207, opened 2019-03-05
square root function update

hanyunx:
- update `sqrt()` in field.py for `p == 1 mod 4 `case
- add test function in test/test_field.py

### DIFF
--- a/honeybadgermpc/field.py
+++ b/honeybadgermpc/field.py
@@ -243,20 +243,43 @@ class GFElement(FieldElement):
 
     def sqrt(self):
         """Square root.
-
         No attempt is made the to return the positive square root.
-
-        Computing square roots is only possible when the modulus
-        is a Blum prime (congruent to 3 mod 4).
         """
-        assert self.modulus % 4 == 3, "Cannot compute square " \
-            "root of %s with modulus %s" % (self, self.modulus)
+        assert self.modulus % 2 == 1, "Modulus must be odd"
+        assert pow(self, (self.modulus-1)//2) == 1
 
-        # Because we assert that the modulus is a Blum prime
-        # (congruent to 3 mod 4), there will be no reminder in the
-        # division below.
-        root = pow(self.value, (self.modulus+1)//4, self.modulus)
-        return GFElement(root, self.field)
+        if self.modulus % 4 == 3:
+            # The case that the modulus is a Blum prime
+            # (congruent to 3 mod 4), there will be no remainder in the
+            # division below.
+            root = pow(self.value, (self.modulus+1)//4)
+            return GFElement(root, self.field)
+        else:
+            # The case that self.modulus % 4 == 1
+            # Cipolla’s Algorithm
+            # http://people.math.gatech.edu/~mbaker/pdf/cipolla2011.pdf
+            t = u = 0
+            for i in range(1, self.modulus):
+                u = i*i - self
+                if pow(u, (self.modulus-1)//2) == self.modulus - 1:
+                    t = i
+                    break
+
+            def cipolla_mult(a, b, w):
+                return ((a[0]*b[0] + a[1]*b[1]*w), (a[0]*b[1] + a[1]*b[0]))
+
+            exp = (self.modulus+1)//2
+            exp_bin = bin(exp)[2:]
+            x1 = (t, 1)
+            x2 = cipolla_mult(x1, x1, u)
+            for i in range(1, len(exp_bin)):
+                if(exp_bin[i] == "0"):
+                    x2 = cipolla_mult(x2, x1, u)
+                    x1 = cipolla_mult(x1, x1, u)
+                else:
+                    x1 = cipolla_mult(x1, x2, u)
+                    x2 = cipolla_mult(x2, x2, u)
+            return x1[0]
 
     def bit(self, index):
         """Extract a bit (index is counted from zero)."""
--- a/tests/test_field.py
+++ b/tests/test_field.py
@@ -1,5 +1,6 @@
 import pytest
 import operator
+from pytest import raises
 
 
 def test_multiple_fields():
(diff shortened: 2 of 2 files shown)


smkuls:
This PR should be raise against `dev` and not `master` branch.

tests/test_field.py
```diff
@@ -25,3 +25,12 @@ def test_invalid_operations_on_fields():
     for op in operators:
         with pytest.raises(FieldsNotIdentical):
             op(field1(2), field2(3))
+
+
+def test_sqrt():
```
amiller: I'd rather see sqrt on a few random tests (say 50? should still be fast) than just one hardcoded one

amiller (approved):
This looks good!
- The travis ci is failing just because our `charm` dependency checked in a build-breaking change. We can pin it to a previous working version to work around it
- Requested more test cases

honeybadgermpc/field.py
```diff
@@ -239,20 +239,42 @@ def __rdiv__(self, other):
...
-        # Because we assert that the modulus is a Blum prime
-        # (congruent to 3 mod 4), there will be no reminder in the
-        # division below.
-        root = pow(self.value, (self.modulus+1)//4, self.modulus)
-        return GFElement(root, self.field)
+        assert self.modulus % 2 == 1
+
+        if self.modulus % 4 == 3:
+            # The case that the modulus is a Blum prime
+            # (congruent to 3 mod 4), there will be no reminder in the
```
amiller: reminder -> remainder

tests/test_field.py
hanyunx: Will do.

honeybadgermpc/field.py
```diff
@@ -239,20 +239,42 @@ def __rdiv__(self, other):
...
+            # division below.
+            root = pow(self.value, (self.modulus+1)//4)
+            return GFElement(root, self.field)
+        else:
+            # The case that self.modulus % 4 == 1
+            # Cipolla’s Algorithm
+            # http://people.math.gatech.edu/~mbaker/pdf/cipolla2011.pdf
+            t = u = 0
+            for i in range(1, self.modulus):
```
smkuls, line 262: This is going to be a really long running loop...

We should benchmark this.

honeybadgermpc/field.py
```diff
@@ -239,20 +239,42 @@ def __rdiv__(self, other):
...
+                u = i*i - self
```
smkuls, line 263: can you add test cases which cover both `if` and `else` part? This code is kinda strange although I can see that it works. Can you replace this with `self.value`?

honeybadgermpc/field.py
```diff
@@ -239,20 +239,42 @@ def __rdiv__(self, other):
...
+                if pow(u, (self.modulus-1)//2) == self.modulus - 1:
+                    t = i
+                    break
+
+            def cipolla_mult(a, b, w):
+                return ((a[0]*b[0] + a[1]*b[1]*w), (a[0]*b[1] + a[1]*b[0]))
+
+            exp = (self.modulus+1)//2
+            exp = [int(x) for x in bin(exp)[2:]]
```
smkuls: You don't need to necessarily convert this to an int. You can do `exp_bin = bin(exp)[2:]` and the check for `exp[i] == "0"` later. I would also request you to rename this variable to something else since you are using the same name just before this line.

tests/test_field.py
smkuls: We should write test cases to cover both `if` and `else` cases and then a few random ones.

honeybadgermpc/field.py
smkuls: Can you add a way to benchmark this for a few random variables in `benchmark/`? There's already a file there which shows how to write a sample benchmark. Let me know if you need any help.

smkuls (changes requested):

honeybadgermpc/field.py
hanyunx: `self.value` doesn't work actually.

tests/test_field.py
smkuls: use the `galois_field` fixture instead of redefining the field, look at `test_phase1` in `test_powermixing.py`.

honeybadgermpc/field.py
amiller: It was making me wonder.... the algorithm is guaranteed to work quickly when each attempt is random, rather than counting up from 1. But it's clearly simpler just to count up from 1. However, I wonder if it's possible then to grind for a worst-case scenario?

amiller:
@hanyunxu  if you merge from `dev` now you will pick up the change to the dockerfile that makes the tests pass again; ask if you need help w/ such merge

tests/test_field.py
```diff
@@ -27,6 +28,13 @@ def test_invalid_operations_on_fields():
             op(field1(2), field2(3))
 
 
-def test_random(galois_field):
-    from honeybadgermpc.field import GFElement
-    assert type(galois_field.random()) is GFElement
+def test_sqrt(galois_field):
+    from honeybadgermpc.elliptic_curve import Subgroup
```
smkuls: This import is not needed since it can be obtained from `galois_field.modulus`

tests/test_field.py
```diff
@@ -27,6 +28,13 @@ def test_invalid_operations_on_fields():
...
+    field = galois_field
+
+    for i in range(100):
+        num = field(random.randint(0, Subgroup.BLS12_381))
```
smkuls: you can infact do `galois_field.random()` without needing the modulus

tests/test_field.py
```diff
@@ -27,6 +28,13 @@ def test_invalid_operations_on_fields():
...
+        while pow(num, (Subgroup.BLS12_381-1)//2) != 1:
+            num = field(random.randint(0, Subgroup.BLS12_381))
```
smkuls: same here

tests/test_field.py
```diff
@@ -27,6 +28,13 @@ def test_invalid_operations_on_fields():
...
-def test_random(galois_field):
-    from honeybadgermpc.field import GFElement
-    assert type(galois_field.random()) is GFElement
+def test_sqrt(galois_field):
+    from honeybadgermpc.elliptic_curve import Subgroup
+    field = galois_field
+
+    for i in range(100):
+        num = field(random.randint(0, Subgroup.BLS12_381))
+        while pow(num, (Subgroup.BLS12_381-1)//2) != 1:
```
smkuls: can you add a comment here so as to what this is doing? I am guessing you are trying to find a random number which has a root? What happens when the number does not have a root and you call `sqrt` does it error out? can you please add a test cast to test that?

smkuls (commented):
I have given minor feedback. Once you update the PR with these changes or respond to the comments I will merge the code.

smkuls (approved):

PR #208, opened 2019-03-05
Work with a specific commit of charm

smkuls:
Current `charm` branch does not build. In order to get around that, we
can stick to a specific commit when things were OK. This commit is from
March 1.


amiller (approved):
I tried this and it works. Pinning is just as well in this case.

PR #210, opened 2019-03-05
Add a method to generate a random value in a field

smkuls:
Remove all traces of `randint(0, modulus-1)` and add a method to
generate a random element in a field. This method is added in the GF
class and returns a random field element.


amiller (approved):
ok, uncontroversial

PR #212, opened 2019-03-07
AVSS refactoring

smkuls:
**Change missing randints to field.random()**

**Cleanup AVSS code**
* Move redundant polynomial code from AVSS code and reuse code in
`polynomial.py`. This required modifying some code in `betterpairing.py`
in the `ZR` class to support certain operations.
* Move symmetric cryptography code out from AVSS to a separate file
and add tests for it.
* Move polynomial commitment code out from AVSS to a separate file and
add tests for it.
* Change `ZR.rand()` to `ZR.random()` to be consistent with
`field.random()` so both can be used seamlessly in the  `Polynomial`
class.

**Reimplementing HbAVSS light**
HbAVSS light has been reimplemented for the purpose of offline phase.
This also does not assume that the dealer is outside of the parties that are
participating which is the case in the offline phase. The client mode is
also supported where the dealer is not one of the parties
participating in the MPC as was the case with the previous version.


honeybadgermpc/hbavss_light.py
```diff
@@ -0,0 +1,135 @@
...
+from honeybadgermpc.symmetric_crypto import SymmetricCrypto
+from honeybadgermpc.exceptions import HoneyBadgerMPCError
+from honeybadgermpc.protocols.reliablebroadcast import reliablebroadcast
+
+
+class HbAVSSMessageType:
+    OK = "OK"
+
+
+class HbAvss(object):
```
amiller: name this HbAvss light to make room for the batch version?

honeybadgermpc/hbavss_light.py
```diff
@@ -0,0 +1,135 @@
...
+        # If the same sends, recvs are reused then the recv_loops of the two objects
+        # can get values belonging to the other object.
+        self.send, self.recv = send, recv
+        self.field = ZR
+        self.poly = polynomials_over(self.field)
+
+        # When running multiple AVSSes, the messages are forwarded to the correct
+        # AVSS using the queue. The key is the AVSS ID which is sent along with
+        # each AVSS message.
+        self.queues = defaultdict(asyncio.Queue)
```
amiller: So a single instance of this `HbAvss` object supports multiple AVSS's from the same dealer, right?

honeybadgermpc/hbavss_light.py
```diff
@@ -0,0 +1,135 @@
...
+        logging.debug("N: %d %d", n, self.n)
+
+        avss_msg = await reliablebroadcast(
+            avss_id,
+            self.my_id,
+            n,
+            self.t,
+            dealer_id,
+            broadcast_msg,
+            self.queues[avss_id].get,
```
amiller: There is a potential message ordering problem here. You are using the same read channel `self.queues[avss_id].get` for both the reliable broadcast subprotocol and for the AVSS-specific OK messages. It is possible that these two kinds of messages can be interleaved. These will need to be partitioned somehow into, for example, `self.RBCqueues` and `self.AVSSqueues` based on a tag

amiller (changes requested):
see note about potential race condition

honeybadgermpc/hbavss_light.py
smkuls: Yes, this is demonstrated in the tests as well.

honeybadgermpc/hbavss_light.py
smkuls: Makes sense. If someone finishes RBC and starts sending AVSS OK messages then that will screw things up. Will split this into avss and rbc queues.

smkuls: This has been fixed.

honeybadgermpc/hbavss_light.py
smkuls: Done.

honeybadgermpc/hbavss_multi.py
```diff
@@ -249,24 +52,20 @@ def __init__(self, publicparams, privateparams, send, recv):
...
-        (poly, polyhat, sharedkeys, shares, encryptedshares,
-         witnesses, encryptedwitnesses) = ([], [], {}, {}, {}, {}, {})
-        for i in range(t+1):
-            poly.append(ZR.rand())
-            polyhat.append(ZR.rand())
-        poly[0] = ZR(secret)
-        sk = ZR.rand()
+        (sharedkeys, shares, encryptedshares,
+         witnesses, encryptedwitnesses) = ({}, {}, {}, {}, {})
+        poly = polynomials_over(use_rust=True).random(t, secret)
```
amiller: doesn't `polynomials_over` have a required first argument?

amiller (changes requested):
I think what's happening is that there are no test cases for `hbavss_multi.py`, yet hbavss_multi was also changed. As part of this refactoring we should probably get rid of `hbavss_multi` or at least refactor it not to be redundant

amiller (approved):

honeybadgermpc/hbavss_multi.py
smkuls: I have removed it. I have also added an `avss_parallel` method in `HbAvssLight` which effectively replaces `hbavss_multi.py` and added a test case for the same.

This is also how I expect the batch interface to be.

PR #213, opened 2019-03-08
fix run-on-ec2 to support batch_openning benchmark

lu562:


smkuls (changes requested):
Can you please squash the two commits to one? Can you also write a proper commit message and a description? This will help us when looking at logs later to see what changes are a part of the commit.

smkuls (approved):
Looks great!

PR #214, opened 2019-03-09
Implement AVSSValueProcessor

smkuls:
This is the component of the offline phase which is responsible for
agreeing upon AVSS values from different nodes. The output of this
component is an `asyncio.Future` which eventually resolves to the value
which was AVSSed. The output is obtianed from a queue which maintains
the order across all AVSS parties for the shared values i.e. ensuring
that the values in the queue corresponding to the shares of the same
values across all parties.

### DIFF
--- a/tests/test_avss_value_processor.py
+++ b/tests/test_avss_value_processor.py
@@ -1,0 +1,321 @@
+import asyncio
+from pytest import mark
+from contextlib import ExitStack
+from pickle import dumps
+from honeybadgermpc.avss_value_processor import AvssValueProcessor
+from honeybadgermpc.protocols.crypto.boldyreva import dealer
+
+
+@mark.asyncio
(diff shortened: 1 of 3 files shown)


tests/test_avss_value_processor.py
```diff
@@ -0,0 +1,321 @@
...
+        dumps([0, 4, 0, 0]),
+    )),
+    # When t+1 parties other than the current node receive at least k values.
+    (4, 1, [0, 6, 0, 0], (
+        dumps([0, 4, 0, 0]),
+        dumps([0, 5, 0, 0]),
+        dumps([0, 6, 0, 0]),
+        dumps([0, 7, 0, 0]),
+    )),
+    (4, 1, [3, 5, 3, 5], (
```
amiller, line 179: neat test

amiller (approved):
This looks great, the tests cover the interface behavior that's required

PR #215, opened 2019-03-09
Add a benchmark for batch_opening

smkuls:
* Move all fixutres in conftest.py to fixtures.py so that they can be
reused between `benchmark` and `tests`
* Add a benchmark for `batch_opening`


amiller (approved):
Looks fine

PR #216, opened 2019-03-13
Tweak HbAVSSLight & AVSSValueProcessor for offline

smkuls:
**Tweak HbAVSSLight & AVSSValueProcessor for offline**

_* HbAVSSLight:_
* Add a notion of avss_id used to sequence the AVSSes per dealer.
* Switch to splitting send/recv channels for RBC and AVSS instead of
having one receive loop and introducing new message types.
* Add an output queue to which the share is added the moment it receives
enough acks.
* Update documentation to be clearer.
* Update tests to reflect the changes.

_* AVSSValueProcessor:_
* Run ACS periodically instead of having to run on demand.
* Add code to sequence the AVSS values per dealer based on the AVSS Id.
* Update tests to reflect the change.

_* Sequencer:_
* Added a sequencer module which exposes an interface to sequence values
from 0.
* Added tests for the same.

**Add node_id to logs generated in protocol/**

### DIFF
--- a/honeybadgermpc/avss_value_processor.py
+++ b/honeybadgermpc/avss_value_processor.py
@@ -1,30 +1,37 @@
 import asyncio
 import logging
 from pickle import dumps, loads
+from collections import defaultdict
 from honeybadgermpc.protocols.commonsubset import run_common_subset
-from honeybadgermpc.exceptions import HoneyBadgerMPCError
-
-
-class NotEnoughAVSSValuesError(HoneyBadgerMPCError):
-    pass
+from honeybadgermpc.sequencer import Sequencer
 
 
 class AvssValueProcessor(object):
+    # How often to wait before checking if a value became available?
+    WAIT_PERIOD_IN_SECONDS = 0.1
+    # How often to wait before running another instance of ACS?
+    ACS_PERIOD_IN_SECONDS = 1
 
     def __init__(self, pk, sk, n, t, my_id, send, recv, get_input):
-
         # This stores the AVSSed values which have been received from each dealer.
         self.inputs_per_dealer = [list() for _ in range(n)]
 
         # This stores the AVSSed values for each dealer which have been agreed.
         # This is a list of Futures which are marked done when the value is received.
         # The Future gurantees that the AVSS value has been agreed by at least `t+1`
-        # parties and will resolve to the value once it is received by this node.
+        # parties and will resolve to a share once it is received by this node.
         self.outputs_per_dealer = [list() for _ in range(n)]
 
-        # This stores a list of the idxes of the next AVSS value to be returned
+        # This stores a list of the indices of the next AVSS value to be returned
         # when a consumer requests a value dealt from a particular dealer.
         self.next_idx_to_return_per_dealer = [0]*n
+
+        # The input to ACS is the count of values received at this node dealt by all
+        # the nodes. In order to get the share of the same element at each node the
+        # ordering of messages per dealer needs to be ensured. This is ensured by a
+        # using an instance of a sequencer per dealer. The sequencer buffers any values
+        # received out of order and delivers them in an order based on the AVSS Id.
+        self.sequencers = defaultdict(Sequencer)
 
         self.pk, self.sk = pk, sk
         self.n, self.t, self.my_id = n, t, my_id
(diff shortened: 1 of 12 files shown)


honeybadgermpc/avss_value_processor.py
```diff
@@ -1,85 +1,104 @@
...
-
-
-class NotEnoughAVSSValuesError(HoneyBadgerMPCError):
-    pass
+from honeybadgermpc.sequencer import Sequencer
 
 
 class AvssValueProcessor(object):
+    # How often to wait before checking if a value became available?
+    WAIT_PERIOD_IN_SECONDS = 0.1
```
amiller, line 11: asynchronous purists will identify this as a potential red flag - timeout parameters will draw attention. It's OK, because we do not actually rely on these for safety OR for liveness. We should explicitly mark these as such, and just keep track of the proof obligation we need to explain

honeybadgermpc/protocols/crypto/boldyreva.py
```diff
@@ -173,7 +173,10 @@ def __setstate__(self, d):
 def dealer(players=10, k=5, seed=None):
     """ """
     # Random polynomial coefficients
-    a = group.random(ZR, count=k, seed=seed)
+    if seed is not None:
+        a = [group.random(ZR, seed=seed+i) for i in range(k)]
```
amiller, line 177: can we file a bug report to charm about this

honeybadgermpc/sequencer.py
```diff
@@ -0,0 +1,25 @@
+import heapq
+
+
+class Sequencer(object):
```
amiller, line 4: Nice, this makes sense. It's an abstract data structure implemented just on top of a priority queue. It holds items of the form `(i, val)`. Items can be enqueued in any order, but `get` will always return them in strict counting order, `(1,_),(2,_),...` so values added out of order are buffered until the next counting number is available

amiller (approved):
Looks great to me

honeybadgermpc/avss_value_processor.py
```diff
@@ -1,85 +1,104 @@
...
+                # So we need to set the result of that future to be equal to this value.
+                idx = len(self.inputs_per_dealer[dealer_id])-1
+                if idx < len(self.outputs_per_dealer[dealer_id]):
+                    assert not self.outputs_per_dealer[dealer_id][idx].done()
+                    self.outputs_per_dealer[dealer_id][idx].set_result(avss_value)
+
+    async def _acs_runner(self):
+        logging.debug("[%d] Starting ACS runner", self.my_id)
+        acs_counter = 0
+        while True:
```
amiller, line 95: I'd like to iterate a bit on the policy for how we decide when to start the ACS next. Having a timeout is great. But another policy can be to wait until some new completed values have been added to the AVSS queues. I think there is not a perfect answer here, just heuristics, but we can discuss some options

honeybadgermpc/sequencer.py
```diff
@@ -0,0 +1,25 @@
...
+        assert self.is_next_available()
+        value = heapq.heappop(self.heap)
+        self.values.remove(value[0])
+        self.next += 1
+        return value
+
+    def is_next_available(self):
+        return len(self.heap) > 0 and self.heap[0][0] == self.next
+
+    def add(self, value):
```
rahulgovind, line 36: It seems like the interface `add(self, sequence_number, value)` would be a good fit here?

smkuls: I had thought about it but this was inspired by the `heapq.heappush` method which follows the same design. I would have to create a tuple to club the values anyway to push them on to the `heap`. Having it as a tuple, also means that I can push the AVSSed value directly without having to extract the sequence number from it first and then insert it.

honeybadgermpc/avss_value_processor.py
smkuls: I can think of a way to get rid of it and make the get a _pure async_ method.

honeybadgermpc/protocols/crypto/boldyreva.py
smkuls: I will get rid of charm eventually and replace it with a suitable `betterpairing` version. I will still file a bug on them. I have opened #217 (assigned to me) to track this.

honeybadgermpc/avss_value_processor.py
smkuls: Opened #218 to track this.

honeybadgermpc/avss_value_processor.py
smkuls: Sure, we can definitely be smarter about how we do this. In fact, we can have both - a timeout along with doing an ACS only when we have some available values or more than `k` values.

Opened #218 to track this.

Issue #202, opened 2019-03-02
Implement hbAVSS (batch)


amiller:
We currently have an implementation of hbAVSS-light, an AVSS scheme with `O(N^2)` communication cost to share a single value. With `hbavss-multi` we also have a script for running a bunch of these concurrently, though still with `O(N^2)` overhead. The hbAVSS protocol described here achieves `O(N)` total communication overhead for a large batch. Implementing this is among our top priorities since it's our main novel algorithm. https://arxiv.org/abs/1902.06095

amiller:
To close once the implicate phase & constant time polycommit are integrated

amiller:
Closed with #227 #257

PR #219, opened 2019-03-14
First patch of Batch HBAVSS

nqyy:
This is the first patch. ``reliablebroadcast`` is used here instead of the ``disperse``, which will be added in later. This only contains the basic implementation without Implication Handling and Share Recovery. A simple test is added to test if we can broadcast the shares and reconstruct the batch.

I combined the light version and the batch version together using inheritance and renamed the file as ``hbavss.py``. The tests are also combined and the file is renamed to ``test_hbavss.py``

This also fixes a potential issue that a malicious node may send an OK or READY message multiple times.

https://github.com/initc3/HoneyBadgerMPC/issues/202

P.S. ``z`` is passed as a flattened 1-d list. Don't know why it doesn't work with 2d. Maybe there is a small bug somewhere that took me and Samarth long time without finding it. I'll just put it this way first.

### DIFF
--- a/honeybadgermpc/hbavss.py
+++ b/honeybadgermpc/hbavss.py
@@ -20,9 +20,12 @@ logger.setLevel(logging.ERROR)
 
 class HbAVSSMessageType:
     OK = "OK"
+    # IMPLICATE = "IMPLICATE"
+    READY = "READY"
 
 
-class HbAvssLight(object):
+class HbAvss(object):
+    # base class of HbAvss containing light and batch version
     def __init__(self, public_keys, private_key, g, h, n, t, my_id, send, recv):
         self.public_keys, self.private_key = public_keys, private_key
         self.n, self.t, self.my_id = n, t, my_id
@@ -50,6 +53,8 @@ class HbAvssLight(object):
     def __exit__(self, type, value, traceback):
         self.subscribe_recv_task.cancel()
 
+
+class HbAvssLight(HbAvss):
     async def _process_avss_msg(self, avss_id, dealer_id, avss_msg):
         tag = f"{dealer_id}-{avss_id}-AVSS"
         send, recv = self.get_send(tag), self.subscribe_recv(tag)
@@ -68,11 +73,13 @@ class HbAvssLight(object):
             logging.error("PolyCommit verification failed.")
             raise HoneyBadgerMPCError("PolyCommit verification failed.")
 
-        oks_recvd = 0
-        while oks_recvd < 2*self.t + 1:
-            _, avss_msg = await recv()  # First value is the `sid`
-            if avss_msg == HbAVSSMessageType.OK:
-                oks_recvd += 1
+        # to handle the case that a (malicious) node sends
+        # an OK message multiple times
+        ok_set = set()
+        while len(ok_set) < 2 * self.t + 1:
+            sender, avss_msg = await recv()  # First value is the `sid`
+            if avss_msg == HbAVSSMessageType.OK and sender not in ok_set:
+                ok_set.add(sender)
 
         # Output the share as an integer so it is not tied to a type like ZR/GFElement
         share_int = int(share)
@@ -138,7 +145,7 @@ class HbAvssLight(object):
             assert dealer_id == self.n
         assert type(avss_id) is int
 
-        logger.debug("[%d] Starting AVSS. Id: %s, Dealer Id: %d, Client Mode: %s",
+        logger.debug("[%d] Starting Light AVSS. Id: %s, Dealer Id: %d, Client Mode: %s",
                      self.my_id, avss_id, dealer_id, client_mode)
 
         broadcast_msg = None if self.my_id != dealer_id else self._get_dealer_msg(value)
@@ -182,3 +189,141 @@ class HbAvssLight(object):
             v = None if values is None else values[i]
             avss_tasks[i] = asyncio.create_task(self.avss(k*avss_id+i, v, dealer_id))
         return await asyncio.gather(*avss_tasks)
+
+
+class HbAvssBatch(HbAvss):
+    async def _process_avss_msg(self, avss_id, dealer_id, dispersal_msg):
+        tag = f"{dealer_id}-{avss_id}-B-AVSS"
+        send, recv = self.get_send(tag), self.subscribe_recv(tag)
+
+        def multicast(msg):
+            for i in range(self.n):
+                send(i, msg)
+
+        commitments, ephemeral_public_key, encrypted_witnesses = loads(
+            dispersal_msg)
+        secret_size = len(commitments)
+
+        # all_encrypted_witnesses: n
+        shared_key = pow(ephemeral_public_key, self.private_key)
+
+        shares = [None] * secret_size
+        witnesses = [None] * secret_size
+        # Decrypt
+        for k in range(secret_size):
+            shares[k], witnesses[k] = SymmetricCrypto.decrypt(
+                str(shared_key).encode("utf-8"),
+                encrypted_witnesses[self.my_id * secret_size + k])
+
+        # verify & send all ok
+        for i in range(secret_size):
+            if not self.poly_commit.verify_eval(
+                    commitments[i], self.my_id+1, shares[i], witnesses[i]):
+                # will be replaced by sending out IMPLICATE message later
+                # multicast(HbAVSSMessageType.IMPLICATE)
+                logging.error("PolyCommit verification failed.")
+                raise HoneyBadgerMPCError("PolyCommit verification failed.")
+
+        multicast(HbAVSSMessageType.OK)
+
+        # Bracha-style agreement
+        # to handle the case that a (malicious) node sends
+        # an OK or READY message multiple times
+        ok_set = set()
+        ready_set = set()
+        # if 2t+1 OK or t+1 READY -> send all ready
+        while len(ok_set) < 2 * self.t + 1 and len(ready_set) < self.t + 1:
+            sender, avss_msg = await recv()  # First value is the `sid`
+            if avss_msg == HbAVSSMessageType.OK and sender not in ok_set:
+                ok_set.add(sender)
+            elif avss_msg == HbAVSSMessageType.READY and sender not in ready_set:
+                ready_set.add(sender)
+
+        multicast(HbAVSSMessageType.READY)
+
+        # if 2t+1 ready -> output shares
+        while len(ready_set) < 2 * self.t + 1:
+            sender, avss_msg = await recv()  # First value is the `sid`
+            if avss_msg == HbAVSSMessageType.READY and sender not in ready_set:
+                ready_set.add(sender)
+
+        return shares
+
+    def _get_dealer_msg(self, values):
+        # Sample a random degree-(t,t) bivariate polynomial φ(·,·)
+        # such that each φ(0,k) = sk and φ(i,k) is Pi’s share of sk
+        secret_size = len(values)
+        phi = [None] * secret_size
+        commitments = [None] * secret_size
+        aux_poly = [None] * secret_size
+        # for k ∈ [t+1]
+        #   Ck, auxk <- PolyCommit(SP,φ(·,k))
+        for k in range(secret_size):
+            phi[k] = self.poly.random(self.t, values[k])
+            commitments[k], aux_poly[k] = self.poly_commit.commit(phi[k])
+
+        ephemeral_secret_key = self.field.random()
+        ephemeral_public_key = pow(self.g, ephemeral_secret_key)
+        # for each party Pi and each k ∈ [t+1]
+        #   1. w[i][k] <- CreateWitnesss(Ck,auxk,i)
+        #   2. z[i][k] <- EncPKi(φ(i,k), w[i][k])
+        z = [None] * self.n * secret_size
+        for i in range(self.n):
+            shared_key = pow(self.public_keys[i], ephemeral_secret_key)
+            for k in range(secret_size):
+                witness = self.poly_commit.create_witness(aux_poly[k], i+1)
+                z[i * secret_size + k] = SymmetricCrypto.encrypt(
+                    str(shared_key).encode("utf-8"), (phi[k](i+1), witness))
+
+        return dumps((commitments, ephemeral_public_key, z))
+
+    async def avss(self, avss_id, values=None, dealer_id=None, client_mode=False):
+        """
+        A batched version of avss similar to the one in light version
+        """
+        # If `values` is passed then the node is a 'Sender'
+        # `dealer_id` must be equal to `self.my_id`
+        if values is not None:
+            if dealer_id is None:
+                dealer_id = self.my_id
+            assert dealer_id == self.my_id, "Only dealer can share values."
+        # If `values` is not passed then the node is a 'Recipient'
+        # Verify that the `dealer_id` is not the same as `self.my_id`
+        elif dealer_id is not None:
+            assert dealer_id != self.my_id
+        if client_mode:
+            assert dealer_id is not None
+            assert dealer_id == self.n
+        assert type(avss_id) is int
+
+        logger.debug("[%d] Starting Batch AVSS. Id: %s, Dealer Id: %d, Client Mode: %s",
+                     self.my_id, avss_id, dealer_id, client_mode)
+
+        broadcast_msg = None if self.my_id != dealer_id else self._get_dealer_msg(values)
+        # In the client_mode, the dealer is the last node
+        n = self.n if not client_mode else self.n+1
+
+        tag = f"{dealer_id}-{avss_id}-B-RBC"
+        send, recv = self.get_send(tag), self.subscribe_recv(tag)
+
+        # this will be replaced by Disperse
+        avss_msg = await reliablebroadcast(
+            tag,
+            self.my_id,
+            n,
+            self.t,
+            dealer_id,
+            broadcast_msg,
+            recv,
+            send
+        )
+
+        if client_mode and self.my_id == dealer_id:
+            # In client_mode, the dealer is not supposed to do
+            # anything after sending the initial value.
+            return
+
+        logger.debug("[%d] Dispersal completed.", self.my_id)
+        share = await self._process_avss_msg(avss_id, dealer_id, avss_msg)
+        logger.debug("[%d] Batch AVSS [%s] completed.", self.my_id, avss_id)
+        return share
--- a/tests/test_hbavss.py
+++ b/tests/test_hbavss.py
@@ -3,7 +3,7 @@ from random import randint
 from contextlib import ExitStack
 from honeybadgermpc.polynomial import polynomials_over
 from honeybadgermpc.betterpairing import G1, ZR
-from honeybadgermpc.hbavss_light import HbAvssLight
+from honeybadgermpc.hbavss import HbAvssLight, HbAvssBatch
 from honeybadgermpc.mpc import TaskProgramRunner
 import asyncio
 
@@ -18,7 +18,7 @@ def get_avss_params(n, t):
 
 
 @mark.asyncio
-async def test_hbavss(test_router):
+async def test_hbavss_light(test_router):
     t = 2
     n = 3*t + 1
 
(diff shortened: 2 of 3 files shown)


honeybadgermpc/hbavss_batch.py
```diff
@@ -0,0 +1,235 @@
...
+            # In client_mode, the dealer is not supposed to do
+            # anything after sending the initial value.
+            return
+
+        # logging.debug("[%d] RBC completed.", self.my_id)
+        share = await self._process_avss_msg(avss_id, dealer_id, avss_msg)
+        # logging.debug("[%d] AVSS [%s] completed.", self.my_id, avss_id)
+        return share
+
+    async def avss_parallel(self, avss_id, k, values=None, dealer_id=None):
```
smkuls: This shouldn't be here anymore

honeybadgermpc/hbavss_batch.py
```diff
@@ -0,0 +1,235 @@
...
+from honeybadgermpc.poly_commit import PolyCommit
+from honeybadgermpc.symmetric_crypto import SymmetricCrypto
+from honeybadgermpc.exceptions import HoneyBadgerMPCError
+from honeybadgermpc.protocols.reliablebroadcast import reliablebroadcast
+
+# TODO: Move these to a separate file instead of using it from batch_reconstruction.py
+from honeybadgermpc.batch_reconstruction import subscribe_recv, wrap_send
+
+
+class HbAVSSMessageType:
```
smkuls: I know that you are going to merge the two files later so remember to keep only class.

honeybadgermpc/hbavss_batch.py
```diff
@@ -0,0 +1,235 @@
...
+    OK = "OK"
+    IMPLICATE = "IMPLICATE"
+    READY = "READY"
+
+
+class HbAvssBatch(object):
```
smkuls: I would appreciate if you could actually raise a PR with the inheritance that we discussed earlier. This way I'll know what code you have actually added. Right now there's a lot of duplicated code.

honeybadgermpc/hbavss_batch.py
```diff
@@ -0,0 +1,235 @@
...
+        #     shared_key = pow(self.public_keys[i], ephemeral_secret_key)
+        #     for k in range(self.t + 1):
+        #         witness = self.poly_commit.create_witness(aux_poly[k], i+1)
+        #         z[i][k] = SymmetricCrypto.encrypt(
+        #             str(shared_key).encode("utf-8"), (phi[k](i+1), witness))
+        ############################################################################
+
+        return dumps((commitments, ephemeral_public_key, z))
+
+    async def avss(self, avss_id, value=None, dealer_id=None, client_mode=False):
```
smkuls: it should be `values`

tests/test_hbavss_batch.py
```diff
@@ -0,0 +1,48 @@
...
+from random import randint
+from contextlib import ExitStack
+from honeybadgermpc.polynomial import polynomials_over
+from honeybadgermpc.betterpairing import G1, ZR
+from honeybadgermpc.hbavss_batch import HbAvssBatch
+from honeybadgermpc.mpc import TaskProgramRunner
+import asyncio
+
+
+def get_avss_params(n, t):
```
smkuls: This should be moved to `fixtures` and reused

honeybadgermpc/hbavss_batch.py
```diff
@@ -0,0 +1,235 @@
...
+            dispersal_msg)
+
+        # # all_encrypted_witnesses: n
+        shared_key = pow(ephemeral_public_key, self.private_key)
+
+        shares = [None] * (self.t+1)
+        witnesses = [None] * (self.t+1)
+        # Decrypt
+        for k in range(self.t+1):
+            # print(self.my_id, k, len(encrypted_witnesses[self.my_id * (self.t+1) + k]))
```
smkuls: Please remove commented code.

honeybadgermpc/hbavss_batch.py
```diff
@@ -0,0 +1,235 @@
...
+        ##################### Don't know why this doesn't work #####################
+        # # Decrypt
+        # for k in range(self.t+1):
+        #     shares[k], witnesses[k] = SymmetricCrypto.decrypt(
+        #         str(shared_key).encode("utf-8"), encrypted_witnesses[self.my_id][k])
+        ############################################################################
+
+        # verify & send all ok
+        for i in range(self.t+1):
+            if self.poly_commit.verify_eval(commitments[i], self.my_id+1, shares[i], witnesses[i]):
```
smkuls: nit: You should keep the method call in a separate line since sometimes we want to log the output of the call

honeybadgermpc/hbavss_batch.py
```diff
@@ -0,0 +1,235 @@
...
+                multicast(HbAVSSMessageType.IMPLICATE)
+
+        oks_recvd = 0
+        readys_recvd = 0
+        # if 2t+1 OK or t+1 READY -> send all ready
+        while oks_recvd < 2 * self.t + 1 and readys_recvd < self.t + 1:
+            _, avss_msg = await recv()  # First value is the `sid`
+            if avss_msg == HbAVSSMessageType.OK:
+                oks_recvd += 1
+            if avss_msg == HbAVSSMessageType.READY:
```
smkuls: `if` -> `elif`

honeybadgermpc/hbavss_batch.py
```diff
@@ -0,0 +1,235 @@
...
+                readys_recvd += 1
+
+        # logging.debug("[%d][HbAvssBatch] 2t+1 OKs or t+1 READYs received.", self.my_id)
```
smkuls: Is there any reason why you want to comment the logs? Please either uncomment them or delete them?

honeybadgermpc/hbavss_batch.py
```diff
@@ -0,0 +1,235 @@
...
+        # for k in range(self.t+1):
+        #     shares[k], witnesses[k] = SymmetricCrypto.decrypt(
+        #         str(shared_key).encode("utf-8"), encrypted_witnesses[self.my_id][k])
+        ############################################################################
+
+        # verify & send all ok
+        for i in range(self.t+1):
+            if self.poly_commit.verify_eval(commitments[i], self.my_id+1, shares[i], witnesses[i]):
+                multicast(HbAVSSMessageType.OK)
+            else:
```
smkuls: Since nobody is processing implicate message right now -> it is better to raise an `error` instead of sending the message. This will make debugging easier if something is actually failing.

honeybadgermpc/hbavss_batch.py
```diff
@@ -0,0 +1,235 @@
...
+        ephemeral_secret_key = self.field.random()
+        ephemeral_public_key = pow(self.g, ephemeral_secret_key)
+
+        # construct the payload z
+        z_size = self.n * (self.t + 1)
+        z = [None] * z_size
+        for k in range(z_size):
+            i = k // (self.t + 1)
+            j = k % (self.t + 1)
+            shared_key = pow(self.public_keys[i], ephemeral_secret_key)
```
smkuls: Do not recompute this, you can run a loop outside to compute this and the reuse the keys. You need this value per row and you are computing it for every cell.

smkuls (changes requested):

honeybadgermpc/hbavss_batch.py
nqyy: I don't think this is necessary. The output of ``verify_eval`` is just a boolean.

honeybadgermpc/hbavss_batch.py
nqyy: Removed.

honeybadgermpc/hbavss_batch.py
nqyy: Merged in ``hbavss.py`` with inheritance.

honeybadgermpc/hbavss_batch.py
nqyy: Merged in ``hbavss.py`` with inheritance

honeybadgermpc/hbavss_batch.py
nqyy: Resolved.

tests/test_hbavss_batch.py
nqyy: Files merged. No need for the move.

honeybadgermpc/hbavss_batch.py
nqyy: Resolved.

honeybadgermpc/hbavss_batch.py
nqyy: Resolved.

honeybadgermpc/hbavss_batch.py
nqyy: Resolved.

honeybadgermpc/hbavss_batch.py
nqyy: Resolved.

honeybadgermpc/hbavss_batch.py
nqyy: Resolved.

honeybadgermpc/hbavss.py
```diff
@@ -176,3 +181,152 @@ def _get_dealer_msg(self, value):
...
+        #         z[i][k] = SymmetricCrypto.encrypt(
+        #             str(shared_key).encode("utf-8"), (phi[k](i+1), witness))
+
+        return dumps((commitments, ephemeral_public_key, z))
+
+    async def avss(self, avss_id, values=None, dealer_id=None, client_mode=False):
+        """
+        A batched version of avss similar to the one in light version
+        """
+        # If `value` is passed then the node is a 'Sender'
```
amiller: `value` -> `values`

honeybadgermpc/hbavss.py
```diff
@@ -176,3 +181,152 @@ def _get_dealer_msg(self, value):
...
+            _, avss_msg = await recv()  # First value is the `sid`
+            if avss_msg == HbAVSSMessageType.READY:
+                readys_recvd += 1
+
+        return shares
+
+    def _get_dealer_msg(self, value):
+        # Sample a random degree-(t,t) bivariate polynomial φ(·,·)
+        # such that each φ(0,k) = sk and φ(i,k) is Pi’s share of sk
+        phi = [None] * (self.t+1)
```
amiller: is there an assumption here that `value` (shoudl be renamed `values`) contains exactly `t+1` elements? We should make this more general because the efficiency will be poor if it is only `t+1` at a time. As explained in the hbAVSS writeup it's necessary to amortize some of the work (like the reliable broadcast overhead) against somewhat larger batches

amiller: And at a minimum, if it only works for `t+1` values, that should be checked rather than just omitting the additional values

honeybadgermpc/hbavss.py
```diff
@@ -176,3 +181,152 @@ def _get_dealer_msg(self, value):
...
+        #   2. z[i][k] <- EncPKi(φ(i,k), w[i][k])
+        z = [None] * self.n * (self.t + 1)
+        for i in range(self.n):
+            shared_key = pow(self.public_keys[i], ephemeral_secret_key)
+            for k in range(self.t + 1):
+                witness = self.poly_commit.create_witness(aux_poly[k], i+1)
+                z[i * (self.t+1) + k] = SymmetricCrypto.encrypt(
+                    str(shared_key).encode("utf-8"), (phi[k](i+1), witness))
+
+        # Don't know why this doesn't work
```
amiller: What's this remaining commented out code, can it be removed?

honeybadgermpc/hbavss.py
```diff
@@ -14,9 +14,12 @@
 
 class HbAVSSMessageType:
     OK = "OK"
+    IMPLICATE = "IMPLICATE"
+    READY = "READY"
 
 
-class HbAvssLight(object):
```
amiller, line 25: We want to keep `HbAvssLight` available as well. It seems like this patch is getting rid of HbAvssLight altogether, but we still want both. My preference will be to have this as a separate file `hbavss.py` and `hbavss_light.py` both available

amiller (changes requested):

honeybadgermpc/hbavss.py
nqyy: I followed the paper and I thought that is exactly t+1. I can definitely make that change.

honeybadgermpc/hbavss.py
nqyy: I'll remove them.

honeybadgermpc/hbavss.py
nqyy: It is not removed. The light version is still there. I made a base class containing both classes and they are all in the same file ``hbavss.py``

amiller: OK I misread, marking as resolved

honeybadgermpc/hbavss.py
nqyy: Resolved

honeybadgermpc/hbavss.py
nqyy: Resolved

honeybadgermpc/hbavss.py
nqyy: Resolved

amiller (approved):

nqyy:
Conflict resolved and rebased.

honeybadgermpc/hbavss.py
```diff
@@ -182,3 +187,139 @@ def _get_dealer_msg(self, value):
...
+        # Bracha-style agreement
+        oks_recvd = 0
+        readys_recvd = 0
+        # if 2t+1 OK or t+1 READY -> send all ready
+        while oks_recvd < 2 * self.t + 1 and readys_recvd < self.t + 1:
+            _, avss_msg = await recv()  # First value is the `sid`
+            if avss_msg == HbAVSSMessageType.OK:
+                oks_recvd += 1
+            elif avss_msg == HbAVSSMessageType.READY:
+                readys_recvd += 1
```
rahulgovind: This would break the protocol if any (malicious) node sends an `OK` or `READY` message multiple times

honeybadgermpc/hbavss.py
```diff
@@ -182,3 +187,139 @@ def _get_dealer_msg(self, value):
...
+        # Decrypt
+        for k in range(secret_size):
+            shares[k], witnesses[k] = SymmetricCrypto.decrypt(
+                str(shared_key).encode("utf-8"),
+                encrypted_witnesses[self.my_id * secret_size + k])
+
+        # verify & send all ok
+        for i in range(secret_size):
+            if self.poly_commit.verify_eval(
+                    commitments[i], self.my_id+1, shares[i], witnesses[i]):
```
rahulgovind, line 221: Are the `self.my_id + 1`'s the evaluation points for the polynomial? If so, it might make sense to use `honeybadgermpc.polynomial.EvalPoint` instead.

honeybadgermpc/hbavss.py
smkuls: Nice catch! This needs to be fixed for AVSS light as well.

nqyy: Good point! & Resolved.

honeybadgermpc/hbavss.py
nqyy: This is a nice one. I think I'll leave it like this for consistency. But I found a bug here that I sent out many duplicate OK messages in the for loop and I need to fix it.

honeybadgermpc/hbavss.py
```diff
@@ -182,3 +191,146 @@ def _get_dealer_msg(self, value):
...
+                # will be replaced by sending out IMPLICATE message later
+                # multicast(HbAVSSMessageType.IMPLICATE)
+                logging.error("PolyCommit verification failed.")
+                raise HoneyBadgerMPCError("PolyCommit verification failed.")
+
+        multicast(HbAVSSMessageType.OK)
+
+        # Bracha-style agreement
+        oks_recvd = 0
```
smkuls: you don't need these two variables if you are using a `set`, just call `len(ok_set)`?

honeybadgermpc/hbavss.py
```diff
@@ -20,9 +20,12 @@
 
 class HbAVSSMessageType:
     OK = "OK"
+    IMPLICATE = "IMPLICATE"
```
smkuls: Please remove this, you don't seem to be using it anywhere.

honeybadgermpc/hbavss.py
```diff
@@ -69,10 +74,14 @@ def multicast(msg):
             raise HoneyBadgerMPCError("PolyCommit verification failed.")
 
         oks_recvd = 0
-        while oks_recvd < 2*self.t + 1:
-            _, avss_msg = await recv()  # First value is the `sid`
-            if avss_msg == HbAVSSMessageType.OK:
+        # to handle the case that a (malicious) node sends
+        # an OKmessage multiple times
+        ok_set = set()
```
smkuls, line 78: Same here, you don't need the `oks_recvd` anymore. Thanks for fixing this! :D

honeybadgermpc/hbavss.py
```diff
@@ -69,10 +74,14 @@ def multicast(msg):
             raise HoneyBadgerMPCError("PolyCommit verification failed.")
 
         oks_recvd = 0
-        while oks_recvd < 2*self.t + 1:
-            _, avss_msg = await recv()  # First value is the `sid`
-            if avss_msg == HbAVSSMessageType.OK:
+        # to handle the case that a (malicious) node sends
+        # an OKmessage multiple times
```
smkuls: nit: space after OK

honeybadgermpc/hbavss.py
```diff
@@ -182,3 +191,146 @@ def _get_dealer_msg(self, value):
             v = None if values is None else values[i]
             avss_tasks[i] = asyncio.create_task(self.avss(k*avss_id+i, v, dealer_id))
         return await asyncio.gather(*avss_tasks)
+
+
+class HbAvssBatch(HbAvss):
+    async def _process_avss_msg(self, avss_id, dealer_id, dispersal_msg):
+        tag = f"{dealer_id}-{avss_id}-AVSS"
```
smkuls: can you give it a separate tag prefix? `B-AVSS`?

honeybadgermpc/hbavss.py
```diff
@@ -182,3 +191,146 @@ def _get_dealer_msg(self, value):
...
+        # If `values` is not passed then the node is a 'Recipient'
+        # Verify that the `dealer_id` is not the same as `self.my_id`
+        elif dealer_id is not None:
+            assert dealer_id != self.my_id
+        if client_mode:
+            assert dealer_id is not None
+            assert dealer_id == self.n
+        assert type(avss_id) is int
+
+        logger.debug("[%d] Starting AVSS. Id: %s, Dealer Id: %d, Client Mode: %s",
```
smkuls: Can you update the log to say "Starting Batch AVSS"?

honeybadgermpc/hbavss.py
```diff
@@ -182,3 +191,146 @@ def _get_dealer_msg(self, value):
...
+                     self.my_id, avss_id, dealer_id, client_mode)
+
+        broadcast_msg = None if self.my_id != dealer_id else self._get_dealer_msg(values)
+        # In the client_mode, the dealer is the last node
+        n = self.n if not client_mode else self.n+1
+
+        tag = f"{dealer_id}-{avss_id}-RBC"
```
smkuls: update this tag prefix as well to `B-RBC`?

honeybadgermpc/hbavss.py
```diff
@@ -182,3 +191,146 @@ def _get_dealer_msg(self, value):
...
+        )
+
+        if client_mode and self.my_id == dealer_id:
+            # In client_mode, the dealer is not supposed to do
+            # anything after sending the initial value.
+            return
+
+        logger.debug("[%d] Dispersal completed.", self.my_id)
+        share = await self._process_avss_msg(avss_id, dealer_id, avss_msg)
+        logger.debug("[%d] AVSS [%s] completed.", self.my_id, avss_id)
```
smkuls: Please update this log as well

tests/test_hbavss.py
```diff
@@ -43,7 +43,38 @@ def get_avss_params(n, t):
 
 
 @mark.asyncio
-async def test_hbavss_client_mode(test_router):
+async def test_hbavss_batch(test_router):
```
smkuls, line 46: Can you create a test case which does multiple batch AVSSes from one node in parallel? Only one dealer node, which does two avsses in parallel. This will ensure that there are no tag conflicts or any other id conflicts and that messages are getting routed to the correct AVSS instance.

smkuls (commented):
This looks good to me. I have left some more comments. I have requested one additional test case, once you implement that I will merge the code.

Great work! :)

honeybadgermpc/hbavss.py
nqyy: Resolved.

honeybadgermpc/hbavss.py
nqyy: I'll use it later. I commented it out for now.

honeybadgermpc/hbavss.py
nqyy: Resolved.

honeybadgermpc/hbavss.py
nqyy: Resolved.

honeybadgermpc/hbavss.py
nqyy: Resolved.

honeybadgermpc/hbavss.py
nqyy: Resolved.

honeybadgermpc/hbavss.py
nqyy: Resolved.

honeybadgermpc/hbavss.py
nqyy: Resolved.

tests/test_hbavss.py
nqyy: Will do it later.

smkuls: Sure!

smkuls (approved):

smkuls:
Approving this since @chitianhao will be taking the other changes separately.

PR #220, opened 2019-03-14
Make AVSSValueProcessor output values in an order

smkuls:
* This change ensures that we pick values dealt from at least `n-t` nodes
for each batch.
* More tests have been added to test this functionality and the current
set of tests have been updated.

### DIFF
--- a/honeybadgermpc/avss_value_processor.py
+++ b/honeybadgermpc/avss_value_processor.py
@@ -7,9 +7,7 @@ from honeybadgermpc.sequencer import Sequencer
 
 
 class AvssValueProcessor(object):
-    # How often to wait before checking if a value became available?
-    WAIT_PERIOD_IN_SECONDS = 0.1
-    # How often to wait before running another instance of ACS?
+    # How long to wait before running another instance of ACS?
     ACS_PERIOD_IN_SECONDS = 1
 
     def __init__(self, pk, sk, n, t, my_id, send, recv, get_input):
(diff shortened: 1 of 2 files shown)


honeybadgermpc/avss_value_processor.py
```diff
@@ -33,35 +31,27 @@ def __init__(self, pk, sk, n, t, my_id, send, recv, get_input):
         # received out of order and delivers them in an order based on the AVSS Id.
         self.sequencers = defaultdict(Sequencer)
 
+        # This queue contains values AVSSed from all nodes such that at least `n-t`
```
amiller, line 34: nice explanation

honeybadgermpc/avss_value_processor.py
```diff
@@ -149,6 +133,37 @@ def _process_acs_output(self, acs_outputs):
...
+        pending_values = [None]*self.n
+        pending_counts = [0]*self.n
+        for i in range(self.n):
+            s, e = self.next_idx_to_return_per_dealer[i], len(self.outputs_per_dealer[i])
+            assert e - s >= 0
+            pending_values[i] = list(self.outputs_per_dealer[i][s:])
+            pending_counts[i] = len(pending_values[i])
+
+        pending_counts.sort()
+        # The t_th index represents the maximum values which at least `n-t` nodes have
```
amiller, line 154: Is the sort ascending or descending? The reason I'm asking.... we only want to need to ensure that at least `t+1` nodes have received the value. So it would be OK to take the `t`'th *largest* count, not just the `t` smallest

tests/test_avss_value_processor.py
```diff
@@ -319,3 +323,42 @@
 
         # This is set by another method and should not have been updated.
         assert all(proc.next_idx_to_return_per_dealer[i] == 0 for i in range(n))
+
+
+@mark.parametrize(
+        "n, t, input_next_ids, output_next_ids, per_dealer_input, output_queue_vals", (
```
amiller, line 329: Can you include a guide of how to read this? It looks daunting

amiller (changes requested):
Only change requested - more comments, and `t`'th smallest should actually be `t`'th largest (or if I'm missing something explain why)

smkuls:
@amiller Here's why I took the `t_th` smallest element and not the largest:
```
|  0  |  1  |  2  |  3  |
-------------------------
| 00  | 10  | 20  | 30  |  => Each row is a batch.
| 01  | 11  | 21  | 31  |
|     | 12  | 22  | 32  |
|     |     | 23  | 33  |
|     |     |     | 34  |
|     |     |     | 35  |

Counts => [2, 3, 4, 6]

Sorted in ascending order, t_th idx value = 3
Batch 1 => 00, 10, 20, 30
Batch 2 => 01, 11, 21, 31
Batch 3 => 12, 22, 32

As per my understanding, you want values from at least `n-t` nodes in one batch.

AVSS Value Proessor Output Order => 00, 10, 20, 30, 01, 11, 21, 31, 12, 22, 32

If you sort them in descending order and then take the t_th idx value = 4 then there will be an extra batch:
Batch 4 => 23, 33
```

tests/test_avss_value_processor.py
smkuls: Done

honeybadgermpc/avss_value_processor.py
smkuls: Here's why I took the `t_th` smallest element and not the largest:
```
|  0  |  1  |  2  |  3  |
-------------------------
| 00  | 10  | 20  | 30  |  => Each row is a batch.
| 01  | 11  | 21  | 31  |
|     | 12  | 22  | 32  |
|     |     | 23  | 33  |
|     |     |     | 34  |
|     |     |     | 35  |

Counts => [2, 3, 4, 6]

Sorted in ascending order, t_th idx value = 3
Batch 1 => 00, 10, 20, 30
Batch 2 => 01, 11, 21, 31
Batch 3 => 12, 22, 32

As per my understanding, you want values from at least `n-t` nodes in one batch.

AVSS Value Proessor Output Order => 00, 10, 20, 30, 01, 11, 21, 31, 12, 22, 32

If you sort them in descending order and then take the t_th idx value = 4 then there will be an extra batch:
Batch 4 => 23, 33
```

amiller (approved):
Ok, as best I can tell for now this makes sense.

PR #221, opened 2019-03-15
Implementation of offline phase

smkuls:
**Implement get random, get triple for offline phase**
* This uses AVSS to share random values, refines them and then return
them in an order.
* This is implemented as a blocking method which returns whenever a
random number becomes available. In the background we continuously AVSS
and process these values.
* This was tested by opening both a single as well a batch in an MPC
program.
* The same has been repeated for triples as well where a set of triples
are shared by each parties and then returned.

PENDING: Triple refinement before returning the triples.

**Remove duplicated code**
* Removed duplicated code in batch_reconstruction and
robust_reconstruction
* Remove unnecessary invocation of asyncio.ensure_future

### DIFF
--- a/honeybadgermpc/offline.py
+++ b/honeybadgermpc/offline.py
@@ -1,0 +1,176 @@
+import asyncio
+import logging
+from honeybadgermpc.hbavss_light import HbAvssLight
+from honeybadgermpc.avss_value_processor import AvssValueProcessor
+from honeybadgermpc.protocols.crypto.boldyreva import dealer
+from honeybadgermpc.betterpairing import G1, ZR
+from honeybadgermpc.polynomial import get_omega, polynomials_over
+from honeybadgermpc.field import GF
+from honeybadgermpc.elliptic_curve import Subgroup
+from honeybadgermpc.batch_reconstruction import subscribe_recv, wrap_send
+from abc import ABC, abstractmethod
+
+
+def get_avss_params(n, t, my_id):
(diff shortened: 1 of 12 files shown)


honeybadgermpc/offline.py
```diff
@@ -0,0 +1,166 @@
...
+        send, recv = self.get_send_recv(f'{self.tag}-AVSS')
+        g, h, pks, sk = get_avss_params(n, t, my_id)
+        self.avss_instance = HbAvssLight(pks, sk, g, h, n, t, my_id, send, recv)
+        self.avss_instance.__enter__()
+        self.tasks.append(asyncio.create_task(self._runner()))
+
+        send, recv = self.get_send_recv(f'{self.tag}-AVSS_VALUE_PROCESSOR')
+        pk, sks = dealer(n, t+1, seed=17)
+        self.avss_vale_processor = AvssValueProcessor(
+            pk, sks[my_id],
```
amiller, line 108: Vale -> value?

smkuls: Changed!

honeybadgermpc/offline.py
```diff
@@ -0,0 +1,176 @@
...
+        return self
+
+    def __exit__(self, *args):
+        for task in self.tasks:
+            task.cancel()
+        self.avss_instance.__exit__(*args)
+        self.avss_value_processor.__exit__(*args)
+
+
+class RandomGenerator(PreProcessingBase):
```
amiller, line 124: This is particularly elegant - it exactly matches the collect inputs -> extract randomness description from the papers

amiller (approved):
Everything looks good

PR #224, opened 2019-03-30
Create Progs folder for example MPC applications

Drake-Eidukas:
This PR creates a `prog` folder for storing example MPC applications, and starts by moving the existing `triple_refinement.py` to `prog/triple_refinement.py`

### DIFF
--- a/progs/triple_refinement.py
+++ b/progs/triple_refinement.py
@@ -1,4 +1,4 @@
-from .polynomial import get_omega
+from honeybadgermpc.polynomial import get_omega
 import asyncio
 import itertools
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,6 +9,7 @@ services:
         BUILD: dev,aws
     volumes:
       - ./apps:/usr/src/HoneyBadgerMPC/apps
+      - ./progs:/usr/src/HoneyBadgerMPC/progs
       - ./benchmark:/usr/src/HoneyBadgerMPC/benchmark
       - ./aws:/usr/src/HoneyBadgerMPC/aws
       - ./conf:/usr/src/HoneyBadgerMPC/conf
--- a/tests/progs/conftest.py
+++ b/tests/progs/conftest.py
@@ -1,0 +1,1 @@
+from tests.fixtures import *  # noqa: F403 F401
--- a/tests/progs/test_triple_refinement.py
+++ b/tests/progs/test_triple_refinement.py
@@ -5,7 +5,7 @@ import asyncio
 @mark.asyncio
 async def test_triple_refinement(test_preprocessing):
     from honeybadgermpc.mpc import TaskProgramRunner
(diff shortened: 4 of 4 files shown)


progs/triple_refinement.py
```diff
@@ -1,4 +1,4 @@
-from .polynomial import get_omega
+from honeybadgermpc.polynomial import get_omega
```
smkuls, line 1: There should be a test for this file. You will probably need to modify the test too. Run `pytest -v` from the container to see that all tests are passing. Anyway, the CI runs tests and it will fail if even a single test fails.

Drake-Eidukas: should the test be moved over to the `progs/` folder too, or should I leave it in test?

smkuls: you can create a `prog` folder in `tests` but then you will have to add a `conftest.py` as well in the `progs` folder which should import `from tests.fixtures import *`. You can basically follow the way things are in the `honeybadgermpc/benchmark` folder.

Drake-Eidukas:
PR _should_ be good to go now

prog/triple_refinement.py
```diff
@@ -60,3 +60,6 @@ def get_extrapolated_values(poly, a, b, d, omega):
     q_shares = map(context.Share, q[:num_valid_triples])
     pq_shares = map(context.Share, pq[:num_valid_triples])
     return p_shares, q_shares, pq_shares
+
+if __name__ == "__main__":
```
smkuls: why do you need this?

tests/prog/test_triple_refinement.py
```diff
@@ -5,7 +5,7 @@
 @mark.asyncio
 async def test_triple_refinement(test_preprocessing):
     from honeybadgermpc.mpc import TaskProgramRunner
-    from honeybadgermpc.triple_refinement import refine_triples
+    import prog.triple_refinement as triple_refinement
```
smkuls: oops,  I had a typo I meant `progs` and not `prog`, can you please fix this?

smkuls (changes requested):
Please rebase your branch so that this whole PR goes only as 1 commit. You can take up the minor changes that I requested and after that lets merge it.

tests/prog/test_triple_refinement.py
smkuls: Why do you need to import the whole file?  `from prog.triple_refinement import refine_triples` should also work, right?

prog/triple_refinement.py
Drake-Eidukas: Oh, forgot to remove this

tests/prog/test_triple_refinement.py
Drake-Eidukas: I don't need to import everything-- fixed both of these.

smkuls (approved):

PR #227, opened 2019-04-03
Implement AVID for Batch HBAVSS

nqyy:
This is the implementation of AVID (Asynchronous Verifiable Information Dispersal) for Batch HBAVSS. It contains the ``disperse`` and ``retrieve`` methods with disperse running in the background. We can retrieve individual item independently in HBAVSS through this class.

### DIFF
--- a/honeybadgermpc/hbavss.py
+++ b/honeybadgermpc/hbavss.py
@@ -7,6 +7,7 @@ from honeybadgermpc.poly_commit import PolyCommit
 from honeybadgermpc.symmetric_crypto import SymmetricCrypto
 from honeybadgermpc.exceptions import HoneyBadgerMPCError
 from honeybadgermpc.protocols.reliablebroadcast import reliablebroadcast
+from honeybadgermpc.protocols.avid import AVID
 
 # TODO: Move these to a separate file instead of using it from batch_reconstruction.py
 from honeybadgermpc.batch_reconstruction import subscribe_recv, wrap_send
@@ -192,6 +193,28 @@ class HbAvssLight(HbAvss):
 
 
 class HbAvssBatch(HbAvss):
+    def __init__(self, public_keys, private_key, g, h, n, t, my_id, send, recv):
+        super(HbAvssBatch, self).__init__(public_keys,
+                                          private_key, g, h, n, t, my_id, send, recv)
+        self.avid_msg_queue = asyncio.Queue()
+        self.tasks = []
+
+    async def _recv_loop(self, q):
+        while True:
+            avid, tag, dispersal_msg_list = await q.get()
+            self.tasks.append(asyncio.create_task(
+                avid.disperse(tag, self.my_id, dispersal_msg_list)))
+
+    def __enter__(self):
+        self._recv_task = asyncio.create_task(self._recv_loop(self.avid_msg_queue))
+        return self
+
+    def __exit__(self, typ, value, traceback):
+        self._recv_task.cancel()
+        for task in self.tasks:
+            task.cancel()
+        super(HbAvssBatch, self).__exit__(typ, value, traceback)
+
     async def _process_avss_msg(self, avss_id, dealer_id, dispersal_msg):
         tag = f"{dealer_id}-{avss_id}-B-AVSS"
         send, recv = self.get_send(tag), self.subscribe_recv(tag)
@@ -212,8 +235,7 @@ class HbAvssBatch(HbAvss):
         # Decrypt
         for k in range(secret_size):
             shares[k], witnesses[k] = SymmetricCrypto.decrypt(
-                str(shared_key).encode("utf-8"),
-                encrypted_witnesses[self.my_id * secret_size + k])
+                str(shared_key).encode("utf-8"), encrypted_witnesses[k])
 
         # verify & send all ok
         for i in range(secret_size):
@@ -267,15 +289,17 @@ class HbAvssBatch(HbAvss):
         # for each party Pi and each k ∈ [t+1]
         #   1. w[i][k] <- CreateWitnesss(Ck,auxk,i)
         #   2. z[i][k] <- EncPKi(φ(i,k), w[i][k])
-        z = [None] * self.n * secret_size
+        dealer_msg_list = [None] * self.n
         for i in range(self.n):
             shared_key = pow(self.public_keys[i], ephemeral_secret_key)
+            z = [None] * secret_size
             for k in range(secret_size):
                 witness = self.poly_commit.create_witness(aux_poly[k], i+1)
-                z[i * secret_size + k] = SymmetricCrypto.encrypt(
+                z[k] = SymmetricCrypto.encrypt(
                     str(shared_key).encode("utf-8"), (phi[k](i+1), witness))
+            dealer_msg_list[i] = dumps((commitments, ephemeral_public_key, z))
 
-        return dumps((commitments, ephemeral_public_key, z))
+        return dealer_msg_list
 
     async def avss(self, avss_id, values=None, dealer_id=None, client_mode=False):
         """
@@ -299,24 +323,21 @@ class HbAvssBatch(HbAvss):
         logger.debug("[%d] Starting Batch AVSS. Id: %s, Dealer Id: %d, Client Mode: %s",
                      self.my_id, avss_id, dealer_id, client_mode)
 
-        broadcast_msg = None if self.my_id != dealer_id else self._get_dealer_msg(values)
+        dispersal_msg_list = None
+        if self.my_id == dealer_id:
+            dispersal_msg_list = self._get_dealer_msg(values)
+
         # In the client_mode, the dealer is the last node
         n = self.n if not client_mode else self.n+1
 
-        tag = f"{dealer_id}-{avss_id}-B-RBC"
+        tag = f"{dealer_id}-{avss_id}-B-AVID"
         send, recv = self.get_send(tag), self.subscribe_recv(tag)
 
-        # this will be replaced by Disperse
-        avss_msg = await reliablebroadcast(
-            tag,
-            self.my_id,
-            n,
-            self.t,
-            dealer_id,
-            broadcast_msg,
-            recv,
-            send
-        )
+        avid = AVID(n, self.t, dealer_id, recv, send, n)
+        # start disperse in the background
+        self.avid_msg_queue.put_nowait((avid, tag, dispersal_msg_list))
+        # message can be retrieved individually
+        avss_msg = await avid.retrieve(tag, self.my_id)
 
         if client_mode and self.my_id == dealer_id:
             # In client_mode, the dealer is not supposed to do
--- a/honeybadgermpc/protocols/avid.py
+++ b/honeybadgermpc/protocols/avid.py
@@ -1,0 +1,224 @@
+# coding=utf-8
+import logging
+import math
+import asyncio
+from honeybadgermpc.exceptions import HoneyBadgerMPCError
+from honeybadgermpc.protocols.reliablebroadcast \
+    import encode, decode, merkle_tree, get_merkle_branch, merkle_verify
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.ERROR)
+# Uncomment this when you want logs from this file.
+# logger.setLevel(logging.NOTSET)
+
+
+class AVIDMessageType:
+    VAL = "VAL"
+    ECHO = "ECHO"
+    READY = "READY"
+    RETRIEVE = "RETRIEVE"
+    RESPONSE = "RESPONSE"
+
+
+class AVID:
+    # Class for Asynchronous Verifiable Information Dispersal
+    def __init__(self, n, t, leader, recv, send, input_size):
+        """ Initialize the class with parameters:
+        :param int n:  at least 3
+        :param int f: fault tolerance, ``N >= 3f + 1``
+        :param int leader: ``0 <= leader < N``
+        :param recv: :func:`recv()` blocks until a message is
+            recvd; message is of the form::
+            (i, (tag, ...)) = recv()
+            where ``tag`` is one of ``{"VAL", "ECHO", "READY"}``
+        :param send: sends (without blocking) a message to a designated
+            recipient ``send(i, (tag, ...))``
+        :param input_size: the size of the input messages to be retrieved
+        """
+        assert n >= 3 * t + 1
+        assert t >= 0
+        assert 0 <= leader < n
+
+        self._data = [None] * n
+        self.n = n
+        self.t = t
+        self.leader = leader
+        self.recv = recv
+        self.send = send
+        # size of input_list for disperse
+        self.input_size = input_size
+        # the response for retrieve
+        self.retrieval_queue = asyncio.Queue()
+        # the future value OK: retrive will be sent until ok is set
+        self.ok_future = asyncio.Future()
+
+    def broadcast(self, o):
+        for i in range(self.n):
+            self.send(i, o)
+
+    async def retrieve(self, sid, index):
+        """ Retrieve individual item
+        :param int index:  the index of retrieval
+        """
+        # we are able to retrieve things from AVID indivdually
+        assert 0 <= index < self.input_size
+
+        # send retrieve after ok is set
+        await self.ok_future
+
+        # send retrieve to all the parties
+        self.broadcast((sid, AVIDMessageType.RETRIEVE, index))
+
+        result = [None] * self.input_size
+        response_set = set()
+
+        # response threshold same as k
+        response_threshold = self.t + 1
+
+        while True:  # recv loop for retrieve
+            sender, msg = await self.retrieval_queue.get()
+            if msg[1] == AVIDMessageType.RESPONSE:
+                (_, _, roothash, data) = msg
+
+                if sender in response_set:
+                    logger.warning("Redundant RESPONSE from %s", sender)
+                    continue
+
+                result[sender] = data
+                response_set.add(sender)
+
+                if len(response_set) >= response_threshold:
+                    # decode the msg on enough response
+                    decoded_output = ""
+                    try:
+                        decoded_output = decode(response_threshold, self.n, result)
+                        # Rebuild the merkle tree to guarantee decoding is correct
+                        _stripes = encode(response_threshold, self.n, decoded_output)
+                        _mt = merkle_tree(_stripes)
+                        _roothash = _mt[1]
+                        if _roothash != roothash:
+                            raise HoneyBadgerMPCError("Failed to verify merkle tree")
+                    except Exception as e:
+                        logger.error("Failed to decode message: %s", e)
+
+                    logger.info("RETRIEVE complete.")
+
+                    return decoded_output
+
+    async def disperse(self, sid, pid, input_list):
+        """ Main information dispersal handling
+        :param int sid: e.x. tag to be used
+        :param int pid: current member id
+        :param int input_list: the list of message
+            for each party from 1 to n from the dealer
+        """
+        # setup some parameters
+        k = self.t + 1  # Wait to reconstruct.
+        echo_threshold = math.ceil((self.n + self.t + 1) / 2)  # number of ECHOs needed
+        ready_threshold = self.t + 1  # number of READYs needed for READY
+        output_threshold = 2 * self.t + 1  # number of READYs needed for future OK
+
+        # leader handling
+        if pid == self.leader:
+            # The leader erasure encodes the input, sending one strip to each participant
+            assert len(input_list) == self.input_size
+
+            # construct the stripe list which contains the stripes for party 1 to party n
+            stripes_list = [None] * self.input_size
+            mt_list = [None] * self.input_size
+            roothash_list = [None] * self.input_size
+            for i, m in enumerate(input_list):
+                stripes_list[i] = encode(k, self.n, m)
+                mt_list[i] = merkle_tree(stripes_list[i])
+                roothash_list[i] = mt_list[i][1]
+
+            # transpose for sending
+            stripes_list_per_party = [list(i) for i in zip(*stripes_list)]
+            for i in range(self.n):
+                branch_list = [None] * self.input_size
+                for j in range(self.input_size):
+                    branch_list[j] = get_merkle_branch(i, mt_list[j])
+                # send each person the column of stripes
+                self.send(i, (sid, AVIDMessageType.VAL, roothash_list,
+                              branch_list, stripes_list_per_party[i]))
+
+        # counters and variables for msg
+        echo_set = set()
+        ready_set = set()
+        ready_sent = False
+        from_leader = None
+        # internal storage
+        my_stripes = [None] * self.input_size
+        my_roothash_list = [None] * self.input_size
+
+        while True:  # main recv loop
+            sender, msg = await self.recv()
+            if msg[1] == AVIDMessageType.VAL and from_leader is None:
+                # Validation
+                (_, _, roothash_list, branch_list, stripes_for_each) = msg
+                if sender != self.leader:
+                    logger.warning(
+                        "[%d] VAL message from other than leader: %d", pid, sender)
+                    continue
+
+                # merkle tree verification
+                validation_fail_flag = False
+                for i in range(len(stripes_for_each)):
+                    # verify each entry in the stripes
+                    if not merkle_verify(self.n, stripes_for_each[i],
+                                         roothash_list[i], branch_list[i], pid):
+                        logger.error("[%d]Failed to validate VAL message", pid)
+                        validation_fail_flag = True
+                        break
+                if validation_fail_flag:
+                    continue
+
+                # Update and store the stripes & roothashes for retrieve
+                from_leader = pid
+                my_stripes = stripes_for_each
+                my_roothash_list = roothash_list
+
+                # echo others
+                self.broadcast((sid, AVIDMessageType.ECHO))
+
+            elif msg[1] == AVIDMessageType.ECHO:
+                # Validation
+                if sender in echo_set:
+                    logger.warning("[%d] Redundant ECHO", pid)
+                    continue
+
+                # Update
+                echo_set.add(sender)
+
+                if len(echo_set) >= echo_threshold and not ready_sent:
+                    ready_sent = True
+                    self.broadcast((sid, AVIDMessageType.READY))
+
+            elif msg[1] == AVIDMessageType.READY:
+                # Validation
+                if sender in ready_set:
+                    logger.warning("[%d] Redundant READY", pid)
+                    continue
+
+                # Update
+                ready_set.add(sender)
+
+                # Amplify ready messages
+                if len(ready_set) >= ready_threshold and not ready_sent:
+                    ready_sent = True
+                    self.broadcast((sid, AVIDMessageType.READY))
+
+                if len(ready_set) >= output_threshold and len(echo_set) >= k:
+                    # update ok future indicating ready for retrieve
+                    if not self.ok_future.done():
+                        self.ok_future.set_result(True)
+
+            elif msg[1] == AVIDMessageType.RETRIEVE:
+                _, _, index = msg
+                # send the response sender requested
+                self.send(sender, (sid, AVIDMessageType.RESPONSE,
+                                   my_roothash_list[index], my_stripes[index]))
+
+            elif msg[1] == AVIDMessageType.RESPONSE:
+                # put in the queue for retrieve
+                self.retrieval_queue.put_nowait((sender, msg))


honeybadgermpc/hbavss.py
```diff
@@ -7,6 +7,7 @@
 from honeybadgermpc.symmetric_crypto import SymmetricCrypto
 from honeybadgermpc.exceptions import HoneyBadgerMPCError
 from honeybadgermpc.protocols.reliablebroadcast import reliablebroadcast
+from honeybadgermpc.protocols.AVID import AVID
```
smkuls: Not sure if I have seen upper case anywhere in file names, lets rename the file to lower case.

honeybadgermpc/hbavss.py
```diff
@@ -267,15 +267,17 @@ def _get_dealer_msg(self, values):
         # for each party Pi and each k ∈ [t+1]
         #   1. w[i][k] <- CreateWitnesss(Ck,auxk,i)
         #   2. z[i][k] <- EncPKi(φ(i,k), w[i][k])
-        z = [None] * self.n * secret_size
+        ret = [None] * self.n
```
smkuls: Please rename this from `ret` to `dealer_msg` or `msg`.

honeybadgermpc/protocols/AVID.py
```diff
@@ -0,0 +1,140 @@
+# coding=utf-8
```
smkuls: Please add test cases for this file, there are no test cases.

honeybadgermpc/protocols/AVID.py
```diff
@@ -0,0 +1,140 @@
...
+            (i, (tag, ...)) = receive()
+        where ``tag`` is one of ``{"VAL", "ECHO", "READY"}``
+        :param send: sends (without blocking) a message to a designed
+        recipient ``send(i, (tag, ...))``
+        """
+        assert n >= 3 * f + 1
+        assert f >= 0
+        assert 0 <= leader < n
+
+        self.__data = [None] * n
```
smkuls: nit: I think the convention is single `_`

honeybadgermpc/protocols/AVID.py
```diff
@@ -0,0 +1,140 @@
...
+    def __init__(self, n, f, leader, receive, send):
+        """ Initialize the class with parameters:
+        :param int n:  at least 3
+        :param int f: fault tolerance, ``N >= 3f + 1``
+        :param int leader: ``0 <= leader < N``
+        :param receive: :func:`receive()` blocks until a message is
+        received; message is of the form::
+            (i, (tag, ...)) = receive()
+        where ``tag`` is one of ``{"VAL", "ECHO", "READY"}``
+        :param send: sends (without blocking) a message to a designed
```
smkuls: typo `designated`

honeybadgermpc/protocols/AVID.py
```diff
@@ -0,0 +1,140 @@
...
+        output_threshold = 2 * self.f + 1  # number of READYs needed for storing
+
+        def broadcast(o):
+            for i in range(self.n):
+                self.send(i, o)
+
+        # leader handling
+        if pid == self.leader:
+            # The leader erasure encodes the input, sending one strip to each participant
+            m_list = input_list  # block until an input list is received
```
smkuls: how is this a blocking call? you are just reassigning a variable, you should update the comment.

honeybadgermpc/protocols/AVID.py
```diff
@@ -0,0 +1,140 @@
...
+            assert len(m_list) == self.n
+
+            # construct the stripe list which contains the stripes for party 1 to party n
+            stripes_list = [None] * self.n
+            for i in range(len(m_list)):
```
smkuls: nit: this loop can be shortened:
```
for i, m in enumerate(m_list):
    striples_list[i] = encode(k, self.n, m)
```

honeybadgermpc/protocols/AVID.py
```diff
@@ -0,0 +1,140 @@
...
+                m = m_list[i]
+                stripes_list[i] = encode(k, self.n, m)
+
+            for i in range(self.n):
+                # send each person the column of stripes
+                stripes_for_each = [stripes[i] for stripes in stripes_list]
```
smkuls: nit: you can avoid this list comprehension by maintaining a `stripes_list_per_party` in the previous loop

honeybadgermpc/protocols/AVID.py
```diff
@@ -0,0 +1,140 @@
...
+                self.send(i, (sid, 'VAL', stripes_for_each))
+
+        result = [None] * self.n  # the result current party got from all parties
+        echo_set = set()
+        ready_set = set()
+        ready_sent = False
+        from_leader = None
+
+        def decode_output():
```
smkuls: you call this method only once in the end, and looks like you can easily get rid of this since you only do one thing inside this function

honeybadgermpc/protocols/AVID.py
```diff
@@ -0,0 +1,140 @@
...
+        where ``tag`` is one of ``{"VAL", "ECHO", "READY"}``
+        :param send: sends (without blocking) a message to a designed
+        recipient ``send(i, (tag, ...))``
+        """
+        assert n >= 3 * f + 1
+        assert f >= 0
+        assert 0 <= leader < n
+
+        self.__data = [None] * n
+        self.receive = receive
```
smkuls: nit: we use `send` and `recv` everywhere in the code, lets stick to that?

honeybadgermpc/protocols/AVID.py
```diff
@@ -0,0 +1,140 @@
...
+                (_, _, stripes_for_each) = msg
+                if sender != self.leader:
+                    logger.info(f"[{pid}] VAL message from other than leader: {sender}")
+                    continue
+
+                # Update
+                from_leader = pid
+                # only send out the part each party need
+                # this is efficient by not sending the whole m
+                for i in range(self.n):
```
smkuls: why not use the broadcast that you defined earlier?

honeybadgermpc/protocols/AVID.py
```diff
@@ -0,0 +1,140 @@
...
+            return m
+
+        while True:  # main receive loop
+            sender, msg = await self.receive()
+            if msg[1] == 'VAL' and from_leader is None:
+                # Validation
+                # TODO merkle tree validation
+                (_, _, stripes_for_each) = msg
+                if sender != self.leader:
+                    logger.info(f"[{pid}] VAL message from other than leader: {sender}")
```
smkuls: log these as warnings and not info messages

smkuls: Same comment below

honeybadgermpc/protocols/AVID.py
```diff
@@ -0,0 +1,140 @@
...
+                # Amplify ready messages
+                if len(ready_set) >= ready_threshold and not ready_sent:
+                    ready_sent = True
+                    broadcast((sid, 'READY'))
+
+                if len(ready_set) >= output_threshold and len(echo_set) >= k:
+                    # store the data inside the body
+                    self.__data[pid] = decode_output()
+                    broadcast((sid, 'STORE'))
+                    return
```
smkuls: Add a log in all places where you break out of the loop, this helps in debugging because you know that you are not stuck in the loop when you look at the log statement.

smkuls (changes requested):
@amiller can verify the implementation. I have given some feedback.

honeybadgermpc/hbavss.py
nqyy: resolved

honeybadgermpc/hbavss.py
nqyy: resolved

honeybadgermpc/protocols/AVID.py
nqyy: It is fully tested by batch HBAVSS. There is no need for special test case just for this. There is also no test for reliable broadcast.

honeybadgermpc/protocols/AVID.py
nqyy: resolved

honeybadgermpc/protocols/AVID.py
nqyy: resolved. There should be the same typo in ``reliablebroadcast.py``

honeybadgermpc/protocols/AVID.py
nqyy: resolved

honeybadgermpc/protocols/AVID.py
nqyy: resolved

honeybadgermpc/protocols/AVID.py
nqyy: resolved

honeybadgermpc/protocols/AVID.py
nqyy: It will be updated with other stuff.

honeybadgermpc/protocols/AVID.py
nqyy: resolved. But reliablebroadcast is still receive

honeybadgermpc/protocols/AVID.py
nqyy: Updated.

honeybadgermpc/protocols/AVID.py
nqyy: resolved. Same case in reliable broadcast.

honeybadgermpc/protocols/AVID.py
nqyy: resolved

honeybadgermpc/protocols/AVID.py
smkuls: Reliable broadcast was copied from HoneyBadgerBFT and is not written by us. It will be refactored later on.

honeybadgermpc/protocols/avid.py
```diff
@@ -0,0 +1,212 @@
...
+    def __init__(self, n, f, leader, recv, send, input_size):
+        """ Initialize the class with parameters:
+        :param int n:  at least 3
+        :param int f: fault tolerance, ``N >= 3f + 1``
+        :param int leader: ``0 <= leader < N``
+        :param recv: :func:`recv()` blocks until a message is
+        recvd; message is of the form::
+            (i, (tag, ...)) = recv()
+        where ``tag`` is one of ``{"VAL", "ECHO", "READY"}``
+        :param send: sends (without blocking) a message to a designated
```
amiller, line 34: Missing input size doc

honeybadgermpc/protocols/avid.py
```diff
@@ -0,0 +1,212 @@
...
+
+                # Amplify ready messages
+                if len(ready_set) >= ready_threshold and not ready_sent:
+                    ready_sent = True
+                    broadcast((sid, 'READY'))
+
+                if len(ready_set) >= output_threshold and len(echo_set) >= k:
+                    # update ok queue indicating ready for retrieve
+                    self.ok_queue.put_nowait(True)
+                    broadcast((sid, 'OK'))
```
amiller: This part isn't right. This "OK" is supposed to be a local output to the caller, e.g. AVSS. There is no "OK" message sent on the network in AVID. It signifies that the AVID is complete. It's necessary to wait for this OK locally before outputting a share for AVSS.

honeybadgermpc/hbavss.py
```diff
@@ -192,6 +193,29 @@ def _get_dealer_msg(self, value):
 
 
 class HbAvssBatch(HbAvss):
+    def __init__(self, public_keys, private_key, g, h, n, t, my_id, send, recv):
+        super(HbAvssBatch, self).__init__(public_keys,
+                                          private_key, g, h, n, t, my_id, send, recv)
+        self.q = asyncio.Queue()
```
smkuls: Please rename this to `avid_msg_queue`.

honeybadgermpc/hbavss.py
```diff
@@ -192,6 +193,29 @@ def _get_dealer_msg(self, value):
...
+        self.tasks = []
+
+    async def _recv_loop(self, q):
+        while True:
+            avid, tag, dispersal_msg_list = await q.get()
+            task = asyncio.create_task(
```
smkuls: nit: `self.tasks.append(asyncio.create_task....`

honeybadgermpc/hbavss.py
```diff
@@ -299,24 +324,19 @@ def _get_dealer_msg(self, values):
         logger.debug("[%d] Starting Batch AVSS. Id: %s, Dealer Id: %d, Client Mode: %s",
                      self.my_id, avss_id, dealer_id, client_mode)
 
-        broadcast_msg = None if self.my_id != dealer_id else self._get_dealer_msg(values)
+        dispersal_msg_list = None if self.my_id != dealer_id else self._get_dealer_msg(
```
smkuls: nit: Make this into an `if` `else` statement instead of inline `if` since this has become too long.

honeybadgermpc/protocols/avid.py
```diff
@@ -0,0 +1,212 @@
...
+                (_, _, roothash, data) = msg
+
+                if sender in response_set:
+                    logger.warning("[{pid}] Redundant RESPONSE")
+                    continue
+
+                result[sender] = data
+                response_set.add(sender)
+
+                if len(response_set) >= response_threshold:
```
amiller, line 90: This logic is incomplete. If decode fails you need to return a default value (e.g., empty string), not just infinite loop

honeybadgermpc/hbavss.py
```diff
@@ -299,24 +324,19 @@ def _get_dealer_msg(self, values):
...
-            tag,
-            self.my_id,
-            n,
-            self.t,
-            dealer_id,
-            broadcast_msg,
-            recv,
-            send
-        )
+        avid = AVID(n, self.t, dealer_id, recv, send, n)
```
smkuls, line 336: Is the `input_size_list` always going to be `n`?

honeybadgermpc/protocols/avid.py
nqyy: OK, does it mean I can just take the broadcast out? The ok queue will handle that.

honeybadgermpc/protocols/avid.py
```diff
@@ -0,0 +1,212 @@
...
+        :param int input_list: the list of message
+            for each party from 1 to n from the dealer
+        """
+        # setup some parameters
+        k = self.n - 2 * self.f  # Wait to reconstruct.
+        echo_threshold = math.ceil((self.n + self.f + 1) / 2)  # number of ECHOs needed
+        ready_threshold = self.f + 1  # number of READYs needed for READY
+        output_threshold = 2 * self.f + 1  # number of READYs needed for storing
+
+        def broadcast(o):
```
smkuls: You can declare this in the constructor and then reuse this even in the `retrieve` method.

honeybadgermpc/protocols/avid.py
```diff
@@ -0,0 +1,212 @@
...
+        """
+        # we are able to retrieve things from AVID indivdually
+        assert 0 <= index < self.input_size
+
+        # send retrieve until ok state
+        assert await self.ok_queue.get()
+
+        # send retrieve to all the parties
+        for i in range(self.n):
+            self.send(i, (sid, "RETRIEVE", index))
```
smkuls: Define this strings as constants in a class instead of hard coding them (like an `AvidMessageType` class). Imagine, if you had a typo somewhere then you will end up not processing the message.

honeybadgermpc/protocols/avid.py
```diff
@@ -0,0 +1,212 @@
...
+        # response threshold same as k
+        response_threshold = self.n - 2 * self.f
+
+        def decode_output(roothash):
+            m = decode(response_threshold, self.n, result)
+            # Rebuild the merkle tree to guarantee decoding is correct
+            _stripes = encode(response_threshold, self.n, m)
+            _mt = merkle_tree(_stripes)
+            _roothash = _mt[1]
+            assert _roothash == roothash
```
smkuls: Instead of an assert, raise an exception here and log that verification failed.

honeybadgermpc/protocols/avid.py
nqyy: I see. I can add a ``try`` ``except`` here.

amiller (changes requested):
I spotted a few logic bugs and commented on them.... ill check again later.

honeybadgermpc/protocols/avid.py
```diff
@@ -0,0 +1,212 @@
...
+            return m
+
+        while True:  # recv loop for retrieve
+            sender, msg = await self.retrieval_queue.get()
+            if msg[1] == 'RESPONSE':
+                (_, _, roothash, data) = msg
+
+                if sender in response_set:
+                    logger.warning("[{pid}] Redundant RESPONSE")
```
smkuls: do not use `logger.warning("[{pid}] Redundant RESPONSE")` instead use `logger.warning("[%s] Redundant RESPONSE", pid)`. The latter will make sure that we will not evaluate the format string when we don't want to log the message.

honeybadgermpc/protocols/avid.py
```diff
@@ -0,0 +1,212 @@
...
+                    continue
+
+                result[sender] = data
```
smkuls, line 87: What do you do with this `result`?

honeybadgermpc/protocols/avid.py
```diff
@@ -0,0 +1,212 @@
...
+            self.send(i, (sid, "RETRIEVE", index))
+
+        result = [None] * self.input_size
+        response_set = set()
+
+        # response threshold same as k
+        response_threshold = self.n - 2 * self.f
+
+        def decode_output(roothash):
+            m = decode(response_threshold, self.n, result)
```
smkuls: rename this to `decoded_output`

honeybadgermpc/protocols/avid.py
```diff
@@ -0,0 +1,212 @@
...
+
+                if sender in response_set:
+                    logger.warning("[{pid}] Redundant RESPONSE")
+                    continue
+
+                result[sender] = data
+                response_set.add(sender)
+
+                if len(response_set) >= response_threshold:
+                    m = decode_output(roothash)
```
smkuls: Add a log here that you are returning from the `loop`.

honeybadgermpc/protocols/avid.py
```diff
@@ -0,0 +1,216 @@
...
+        self.n = n
+        self.f = f
+        self.leader = leader
+        self.recv = recv
+        self.send = send
+        # size of input_list for disperse
+        self.input_size = input_size
+        # the response for retrieve
+        self.retrieval_queue = asyncio.Queue()
+        # A blocking queue: retrive will be sent until the ok state is reached
```
smkuls: Don't call this a blocking Queue, just say Queue, asyncio Queues are non-blocking.

nqyy:
Updated @amiller

honeybadgermpc/protocols/avid.py
```diff
@@ -0,0 +1,216 @@
...
+            recipient ``send(i, (tag, ...))``
+        :param input_size: the size of the input messages to be retrieved
+        """
+        assert n >= 3 * f + 1
+        assert f >= 0
+        assert 0 <= leader < n
+
+        self._data = [None] * n
+        self.n = n
+        self.f = f
```
smkuls: lets rename `f` to `t` -> the paper might say `f` but we use `t` as the threshold in most of the places in the codebase.

honeybadgermpc/protocols/avid.py
```diff
@@ -0,0 +1,216 @@
...
+                    continue
+
+                result[sender] = data
+                response_set.add(sender)
+
+                if len(response_set) >= response_threshold:
+                    try:
+                        m = decode_output(roothash)
+                    except Exception as e:
+                        logger.info(f"Failed to decode message: {e}")
```
smkuls: `logger.error`

honeybadgermpc/protocols/avid.py
```diff
@@ -0,0 +1,216 @@
...
+        my_roothash_list = [None] * self.input_size
+
+        while True:  # main recv loop
+            sender, msg = await self.recv()
+            if msg[1] == 'VAL' and from_leader is None:
+                # Validation
+                (_, _, roothash_list, branch_list, stripes_for_each) = msg
+                if sender != self.leader:
+                    logger.warning(
+                        f"[{pid}] VAL message from other than leader: {sender}")
```
smkuls: same here, use `%s` and not f"{pid}", please fix in all places in the code.

honeybadgermpc/protocols/avid.py
```diff
@@ -0,0 +1,216 @@
...
+
+                # merkle tree verification
+                validation_fail_flag = False
+                for i in range(len(stripes_for_each)):
+                    # verify each entry in the stripes
+                    try:
+                        assert merkle_verify(self.n, stripes_for_each[i],
+                                             roothash_list[i], branch_list[i], pid)
+                    except Exception as e:
+                        logger.info(f"[{pid}]Failed to validate VAL message: {e}")
```
smkuls: `logger.error`?

honeybadgermpc/protocols/avid.py
```diff
@@ -0,0 +1,216 @@
...
+                    logger.warning(
+                        f"[{pid}] VAL message from other than leader: {sender}")
+                    continue
+
+                # merkle tree verification
+                validation_fail_flag = False
+                for i in range(len(stripes_for_each)):
+                    # verify each entry in the stripes
+                    try:
+                        assert merkle_verify(self.n, stripes_for_each[i],
```
smkuls: This is weird, if `merkle_verify` returns a bool then why don't you do an `if` instead of calling `assert`? Checking using asserts is bad when you want to really throw an exception because we disable the asserts when running the code using the `-O` flag.

smkuls: This whole, `try` `except` can be replaced with `if` and `raise`

honeybadgermpc/protocols/avid.py
```diff
@@ -0,0 +1,216 @@
...
+                ready_set.add(sender)
+
+                # Amplify ready messages
+                if len(ready_set) >= ready_threshold and not ready_sent:
+                    ready_sent = True
+                    broadcast((sid, 'READY'))
+
+                if len(ready_set) >= output_threshold and len(echo_set) >= k:
+                    # update ok queue indicating ready for retrieve
+                    self.ok_queue.put_nowait(True)
```
smkuls: Can you get multiple messages in the `ok` queue corresponding to different things? Is it possible that you add an `ok` message in the queue `n` times and then call retrieve `k` `k < n` times and then lets say someone else comes and `retrieve` sees an `ok` in the queue and returns the value even though it does not probably correspond to this retrieve.


What I am trying to get at is that can multiple OK messages cause any problem for retrieve?

smkuls (changes requested):

smkuls:
@chitianhao  This is no longer a basic implementation but a complete implementation. Please update the PR title. Rebase all your commits to a single commit once you are done (also rebase with the `dev` branch). Make sure the commit message is descriptive.

honeybadgermpc/hbavss.py
nqyy: No, but i am using n now.

honeybadgermpc/protocols/avid.py
nqyy: It will be decoded in the ``decode_output``

honeybadgermpc/protocols/avid.py
amiller: Right - this is validation of input from untrusted parties. It should not be an `assert`.

nqyy: So do we want to raise an error and stop this process or we don't add it to the set and keep getting new values?

honeybadgermpc/protocols/avid.py
amiller: Yes you can take out the broadcast.
I think I see how `ok_queue` works.... however, I would suggest instead, making `self.ok_future` as a `Future` object. You can then call `self.ok_future.set_result(True)` instead of `self.ok_queue.put_nowait(True)`. It should basically work the same, but now you can have multiple awaiters on `self.ok_future`

nqyy: Yeah, that's better.

honeybadgermpc/protocols/avid.py
amiller: Print the error, keep accepting new values (from the other, non-faulty, nodes).

Aside: This is reminding me we should have a unified "mark party as byzantine" system (aka, failure detector) since basically if a party is faulty in one place, we should probably ignore them in other contexts too

honeybadgermpc/protocols/avid.py
nqyy: I think I'll update it to future

honeybadgermpc/hbavss.py
nqyy: Resolved.

honeybadgermpc/hbavss.py
nqyy: Resolved.

honeybadgermpc/hbavss.py
nqyy: Resolved.

honeybadgermpc/protocols/avid.py
nqyy: Resolved.

honeybadgermpc/protocols/avid.py
nqyy: Resolved.

honeybadgermpc/protocols/avid.py
nqyy: Resolved.

honeybadgermpc/protocols/avid.py
nqyy: Resolved.

honeybadgermpc/protocols/avid.py
chitianhao: Resolved.

honeybadgermpc/protocols/avid.py
nqyy: Resolved.

honeybadgermpc/protocols/avid.py
nqyy: Resolved.

honeybadgermpc/protocols/avid.py
nqyy: Resolved.

honeybadgermpc/protocols/avid.py
nqyy: Resolved.

honeybadgermpc/protocols/avid.py
nqyy: Resolved.

honeybadgermpc/protocols/avid.py
nqyy: Resolved.

honeybadgermpc/protocols/avid.py
nqyy: Resolved.

honeybadgermpc/protocols/avid.py
nqyy: Resolved.

chitianhao:
I think all comments are resolved except the ``Future`` one. Don't know why that doesn't work.

honeybadgermpc/protocols/avid.py
```diff
@@ -0,0 +1,225 @@
...
+        assert await self.ok_queue.get()
+
+        # send retrieve to all the parties
+        self.broadcast((sid, AVIDMessageType.RETRIEVE, index))
+
+        result = [None] * self.input_size
+        response_set = set()
+
+        # response threshold same as k
+        response_threshold = self.n - 2 * self.t
```
chitianhao: I am not sure about the response threshold we want to set here.

honeybadgermpc/protocols/avid.py
smkuls: This is very confusing code. You call a method above and pass only half the params and are calling the method only once. Can you get rid of the method completely and just move the code of `decode_output` inline?

nqyy: This is confusing to me too. I followed the standard of reliablebroadcast, I should make it better

amiller: I'm ok moving `decode_output` into a function for clarity and to keep the code in the recv loop small. But if so, it should not be a function that relies on global state ideally

honeybadgermpc/protocols/avid.py
```diff
@@ -0,0 +1,225 @@
...
+
+        def decode_output(roothash):
+            decoded_output = decode(response_threshold, self.n, result)
+            # Rebuild the merkle tree to guarantee decoding is correct
+            _stripes = encode(response_threshold, self.n, decoded_output)
+            _mt = merkle_tree(_stripes)
+            _roothash = _mt[1]
+            try:
+                assert _roothash == roothash
```
smkuls: Again, no `assert` when you want to validate something. Asserts are only meant for testing and are disabled when running the code. you should do 
```
if _roothash != roothash:
    raise HoneyBadgerMPCError("Failed to verify merkle tree: %s", e)
```

honeybadgermpc/protocols/avid.py
```diff
@@ -0,0 +1,225 @@
...
+                    continue
+
+                result[sender] = data
+                response_set.add(sender)
+
+                if len(response_set) >= response_threshold:
+                    m = ""
+                    try:
+                        m = decode_output(roothash)
+                    except Exception as e:
```
smkuls, line 101: You should catch only `HoneyBadgerMPCError` or the error that you throw. Don't catch all exceptions. See if reliablebroadcast.py defines a separate exception when merkle tree verification fails.

nqyy: reliablebroadcast doesn't do anything. I think it's ok to keep it this way.

nqyy:
Everything is updated accordingly. I think it should be good @amiller @smkuls .

honeybadgermpc/protocols/avid.py
smkuls: No, you should not eat up all exceptions. This is not good practice. You should only catch `HoneyBadgerMPCError`.

nqyy: But decode doesn't throw ``HoneyBadgerMPCError ``. We won't know what decode will throw

amiller: What is the rule for HoneyBadgerMPCError? We should have something we throw that is caused by clearly incorrect messages sent from another peer, and that should be the only exception we catch here. An

amiller (approved):

smkuls (approved):

PR #228, opened 2019-04-04
Add share and share array inversion

Drake-Eidukas:
This PR adds in the functionality to invert a Share using the `~` operator. I'm not sure if this is the best way to implement this (as it's somewhat not exactly the definition of inverting that `__invert__` normally targets), but this will make the code for my jubjub application easier.

### DIFF
--- a/honeybadgermpc/mpc.py
+++ b/honeybadgermpc/mpc.py
@@ -239,6 +239,23 @@ def share_in_context(context):
 
         def __str__(self): return '{%d}' % (self.v)
 
+        def __div__(self, other):
+            if MixinOpName.InvertShare not in context.config:
+                raise NotImplementedError
+            if MixinOpName.MultiplyShare not in context.config:
+                raise NotImplementedError
+
+            async def divide(curr, other):
+                other_inverted = await(
+                    context.config[MixinOpName.InvertShare](context, other))
+
+                multiplier = context.config[MixinOpName.MultiplyShare]
+                return await(multiplier(context, curr, other_inverted))
+
+            return divide(self, other)
+
+        __truediv__ = __floordiv__ = __div__
+
     def _binop_share(fut, other, op):
         assert type(other) in [ShareFuture, GFElementFuture, Share, GFElement]
         res = ShareFuture()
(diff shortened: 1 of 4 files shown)


Drake-Eidukas:
Not sure if `test_mpc.py` is a good place to test the actual inversion operator, but the test is going to be somewhat redundant compared to the ones in `test_mixins.py`.

amiller:
Should this use divide instead of invert for this? Not sure if there are guidelines that ~ should be something other than multiplicative inverse. Otherwise looks fine

amiller (approved):
Looks good

Drake-Eidukas:
We should decide if we want to merge this as is so that we can get some existing PRs merged before we achieve git-hell, or if we should first fix #234

smkuls:
@Drake-Eidukas can you rebase this once with `dev`. Lets merge this first and then take up #234.

honeybadgermpc/mpc.py
```diff
@@ -239,6 +239,23 @@ def __mul__(self, other):
 
         def __str__(self): return '{%d}' % (self.v)
 
+        def __div__(self, other):
+            if MixinOpName.InvertShare not in context.config:
```
smkuls, line 243: I am wondering if we should write a MixinValidator, now that we have started creating dependent mixins this might end up with a lot of if, else checks. What do you think @Drake-Eidukas ?

tests/test_mixins.py
```diff
@@ -182,3 +164,52 @@
     program_runner.add(_prog)
     results = await program_runner.join()
     assert len(results) == n
+
+
+@mark.asyncio
+async def test_invert_share(test_preprocessing):
```
smkuls, line 170: Do you want to consider adding a negative test case where you don't pass the mixin and check for `NotImplementedError`?

tests/test_mixins.py
```diff
@@ -182,3 +164,52 @@
...
+    n, t = 3, 1
+    test_preprocessing.generate("triples", n, t)
+    test_preprocessing.generate("rands", n, t)
+
+    async def _prog(context):
+        share = test_preprocessing.elements.get_rand(context)
+        # assert 1/x * x == 1
+        assert await(await(await(Inverter.invert_share(context, share))
```
smkuls, line 178: We need to fix this. This is crazy. This needs to align with the data flow design. Lets address this after this PR.

honeybadgermpc/mpc.py
Drake-Eidukas: I think a validator might not be too impossible-- especially if we are considering redoing some of the way we do mixins (i.e. moving stuff into progs). Could be interesting to create some way to have the mixins define the operations they need. I'll think about this some more when I start tackling the refactors to mixins

tests/test_mixins.py
Drake-Eidukas: Agreed-- it was a massive PITA to figure out how to get this to compile and all... err, run without crashing

tests/test_mixins.py
Drake-Eidukas: I'll add this-- good catch

Drake-Eidukas:
@smkuls Should be good to go, just rebased this and added that test. I also moved the tests for divide into test_mixins to fit where multiplication is. Let me know if you want to have that moved back.

smkuls (approved):
LGTM!

Issue #226, opened 2019-04-03
Implement smarter equality checks for `GF`


Drake-Eidukas:
Ran into this while debugging-- essentially, say you create two `GFElements` from two different `GF` objects, each of which are the exact same in every conceivable way except that the actual `GF` objects have a different internal `id` (simply different objects). You currently cannot perform math on these elements, as the operation fails when it checks for equality of the fields they were created on.

I propose we instead check equality of fields via checking their `modulus` field, as well as defining the `__hash__` method to explicitly be that of the `object` class.

@smkuls thoughts on this?

Drake-Eidukas:
Essentially, in the `GFElement` class, when we try to do math operations, we check to see if the elements are from the same field via:
```
if self.field is not other.field:
    raise FieldsNotIdentical
```

This checks if the fields have the same id (i.e. same object), not whether or not the fields that the elements are from are the same (which would require comparing moduli rather than ids).

Drake-Eidukas:
Alternatively, as pointed out by @smkuls, we should just implement `GF` using a more robust singleton pattern, for example, https://python-3-patterns-idioms-test.readthedocs.io/en/latest/Singleton.html

smkuls:
Nice catch, the intended functionality was supposed to be a `Singleton` and this implementation doesn't meet that criteria. The fix would be to make this a singleton class using the link pointed by @Drake-Eidukas and expose only the `get` method and prevent the constructor from getting called directly.

Drake-Eidukas:
Closing issue as we merged in #229

PR #229, opened 2019-04-04
Refactors GF to truly use the singleton pattern

Drake-Eidukas:
Currently, this is the state of `GF`:

```
g1 = GF(1)
g2 = GF(1)
assert g1 is g2 # fails assertion, they're different objects

g3 = GF.get(1)
g4 = GF.get(1)
assert g3 is g4 # passes assertion, intended functionality
```

This PR changes the way that we create `GF` objects, such that both assertions will pass and you can use either method of construction `GF` objects.

For more details, see #226

### DIFF
--- a/honeybadgermpc/field.py
+++ b/honeybadgermpc/field.py
@@ -24,7 +24,7 @@ arithmetic one would expect: addition, multiplication, etc.
 
 Defining a field:
 
->>> Zp = GF.get(19)
+>>> Zp = GF(19)
 
 Defining field elements:
 
@@ -62,7 +62,7 @@ prime modulus (see :func:`GF` for more information):
 Field elements from different fields cannot be mixed, you will get a
 type error if you try:
 
->>> Zq = GF.get(17)
+>>> Zq = GF(17)
 >>> z = Zq(2)
 >>> x + z
 Traceback (most recent call last):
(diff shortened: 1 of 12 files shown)


Drake-Eidukas:
Also, I couldn't get VSCode to save the file without changing line 414... but I'm pretty sure that this is the same functionality with a different format.

honeybadgermpc/field.py
```diff
@@ -109,15 +118,7 @@ def __reduce__(self):
 
     @staticmethod
     def get(modulus):
```
smkuls: * This is redundant now. We should get rid of this. 
* I also want you to run `honeybadgermpc/ipc.py` using the `launch.sh` script. I remember we faced some issue when pickling and unpickling. As of now that's the only way to exercise that code (there are no automated tests for that).

amiller (approved):
Nice fix

honeybadgermpc/field.py
```diff
@@ -98,7 +98,16 @@ def __int__(self):
 class GF(object):
     _field_cache = {}
 
+    def __new__(cls, modulus):
+        if modulus not in GF._field_cache:
+            GF._field_cache[modulus] = super(GF, cls).__new__(cls)
```
smkuls: Can we use the `setdefault` method of `dict`?

honeybadgermpc/field.py
```diff
@@ -410,7 +411,7 @@ def __init__(self, value):
         __invert__ = sqrt = lambda self: FakeFieldElement(return_value)
 
         # Bit extraction. We pretend that the number is *very* big.
-        bit = lambda self, index: 1     # noqa  XXX for the time being
+        def bit(self, index): return 1     # noqa  XXX for the time being
```
smkuls, line 412: Any reason why you changed this?

smkuls (changes requested):

honeybadgermpc/field.py
smkuls: Just saw your comment on the PR, closing this.

honeybadgermpc/field.py
Drake-Eidukas: We should consider adding running that script in travis, but it seems to run just fine

Drake-Eidukas:
Should be good to go now-- implemented samarth's suggestions

smkuls (approved):
LGTM!

honeybadgermpc/field.py
smkuls: Yea, we were discussing running the script using a python wrapper and then making sure that things passed.

PR #232, opened 2019-04-05
Change __nonzero__ to __bool__ in field.py

hanyunx:
We need `bool` instead of `nonzero` in python3.

### DIFF
--- a/honeybadgermpc/field.py
+++ b/honeybadgermpc/field.py
@@ -344,7 +344,7 @@ class GFElement(FieldElement):
         """Hash value."""
         return hash((self.field, self.value))
 
-    def __nonzero__(self):
+    def __bool__(self):
         """Truth value testing.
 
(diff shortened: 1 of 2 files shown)


amiller:
Thank you @hanyunxu! Could you also add a test of using a field in an if statement in  `test_field.py`

hanyunx:
There is a ` AttributeError: type object 'GF' has no attribute 'get'` error in my test function `test_bool()` which I don't know why.

smkuls:
> There is a ` AttributeError: type object 'GF' has no attribute 'get'` error in my test function `test_bool()` which I don't know why.

There's no `get` function anymore. You directly call the constructor. `GF(2)` and not `GF.get(2)`. This was changed by Drake recently.

smkuls (approved):
LGTM!

PR #233, opened 2019-04-05
Implement Jubjub MPC application

Drake-Eidukas:
This adds in a a sample application utilizing an implementation
of a Jubjub curve to perform arithmatic operations on secret shared
points on the curve.

The PR is against the `drake-implement-share-inversion` branch, as it relies on that functionality. This should be merged in against `dev` once #228 is merged in. Additionally, this also requires the functionality in #229 be merged into `dev` before this will work correctly.

### DIFF
--- a/progs/jubjub.py
+++ b/progs/jubjub.py
@@ -1,0 +1,228 @@
+from honeybadgermpc.elliptic_curve import Jubjub, Point
+from honeybadgermpc.mpc import Mpc
+
+
+class SharedPoint(object):
+    """
+    Represents a point with optimized operatons over Edward's curves.
+    This is the 'shared' version of this class, which does deal with shares
+    Math operations derived from
+    https://en.wikipedia.org/wiki/Twisted_Edwards_curve#Addition_on_twisted_Edwards_curves # noqa: E501
+    """
+
+    def __init__(self, context: Mpc, xs, ys, curve: Jubjub = Jubjub()):
+        if not isinstance(curve, Jubjub):
+            raise Exception(
+                f"Could not create Point-- given \
+                curve not of type Jubjub({type(curve)})")
+
+        self.context = context
+        self.curve = curve
+        self.xs = xs
+        self.ys = ys
+
+    async def __on_curve(self) -> bool:
+        """
+        Checks whether or not the given shares for x and y correspond to a
+        point that sits on the current curve
+
+        WARNING: This method currently leaks information about the shared point--
+                 We need to use share equality testing instead
+        """
+        x_sq = await(self.xs * self.xs)
+        y_sq = await(self.ys * self.ys)
+        a_ = self.context.Share(self.curve.a)
+        d_ = self.context.Share(self.curve.d)
+
+        # ax^2 + y^2
+        lhs = await(a_ * x_sq) + y_sq
+
+        # 1 + dx^2y^2
+        rhs = self.context.Share(1) + await(d_ * await(x_sq * y_sq))
+
+        # TODO: use share equality to prevent the leaking of data
+        return await lhs.open() == await rhs.open()
+
+    async def __init(self):
+        """asynchronous part of initialization via create or from_point
+        """
+        if not await(self.__on_curve()):
+            raise Exception(
+                f"Could not initialize Point {self}-- \
+                does not sit on given curve {self.curve}")
+
+    @staticmethod
+    async def create(context: Mpc, xs, ys, curve=Jubjub()):
+        """ Given a context, secret shared coordinates and a curve,
+            creates the given point
+        """
+        point = SharedPoint(context, xs, ys, curve)
+        await point.__init()
+
+        return point
+
+    @staticmethod
+    async def from_point(context: Mpc, p: Point) -> 'SharedPoint':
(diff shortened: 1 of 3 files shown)


progs/jubjub.py
```diff
@@ -0,0 +1,315 @@
...
+        y_sq = await(p.ys * p.ys)
+
+        # ax^2 + y^2
+        lhs = await(context.Share(self.a) * x_sq) + y_sq
+
+        # 1 + dx^2y^2
+        rhs = context.Share(1) + await(context.Share(self.d) * await(x_sq * y_sq))
+
+        return await lhs.open() == await rhs.open()
+
```
amiller: We should omit this method, it leaks information about the underlying value

amiller: Alternately we could compare lhs, rhs with equality program

progs/jubjub.py
```diff
@@ -0,0 +1,315 @@
...
+        # y3 = ((y1*y2) + (x1*x2)) / (1 - d*x1*x2*y1*y2)
+        y3 = await((y_prod + x_prod) / (one - d_prod))
+
+        return await(SharedPoint.create(self.context, x3, y3, self.curve))
+
+    async def sub(self, other: 'SharedPoint') -> 'SharedPoint':
+        return self.add(await(other.mul(-1)))
+
+    async def mul(self, n: int) -> 'SharedPoint':
```
amiller, line 117: This is cool. This is multiplication of a secret shared point with a public scalar.

Note:
What @hanyunxu needs to add to this is a different method for multiplication of a public point by a secret shared scalar

Drake-Eidukas:
Should be ready for review now that everything this depends on has been merged in

progs/jubjub.py
```diff
@@ -0,0 +1,315 @@
...
+        x_sq = await(p.xs * p.xs)
+        y_sq = await(p.ys * p.ys)
+
+        # ax^2 + y^2
+        lhs = await(context.Share(self.a) * x_sq) + y_sq
+
+        # 1 + dx^2y^2
+        rhs = context.Share(1) + await(context.Share(self.d) * await(x_sq * y_sq))
+
+        return await lhs.open() == await rhs.open()
```
rahulgovind: Not sure of the application here but wouldn't this reveal too much information?

progs/jubjub.py
```diff
@@ -0,0 +1,315 @@
...
+            return Ideal(self.curve)
+
+        Q = self
+        R = self if n & 1 == 1 else Ideal(self.curve)
+
+        i = 2
+        while i <= n:
+            Q = await(Q.double())
+
+            if n & i == i:
```
rahulgovind, line 134: I think this logic can be simplified. It's currently split between `R = self if n & 1 == 1 else Ideal(self.curve)` which only considers the least significant bit of `n` and the logic in this loop which handles the rest.

Drake-Eidukas: You're probably right-- I'll try to do this: this is just directly copied from the existing jubjub code.

amiller (changes requested):
meant to submit these comments days ago....

progs/jubjub.py
```diff
@@ -0,0 +1,315 @@
+from honeybadgermpc.field import GF, GFElement
+from honeybadgermpc.elliptic_curve import Subgroup
+from honeybadgermpc.mpc import Mpc
+
+Field = GF(Subgroup.BLS12_381)
```
smkuls: Umm, I am not usually a fan of global variables. It is very easy for someone to change this from another file which could potentially break a lot of things if more people start to reuse it. I would suggest thinking of another way to use this. Move this within the Jubjub class as a static member?

smkuls: I have noticed that if you expose the variable like this then more people start to reuse it in places which have no connection with Jubjub just because they want to use the field.

Drake-Eidukas:
So, this is more or less done, but I've been running up against issues testing my implementation of montgomery multiplication-- I keep getting StopIteration errors, which I assume are due to either timeout or running out of preprocessing...

smkuls:
> So, this is more or less done, but I've been running up against issues testing my implementation of montgomery multiplication-- I keep getting StopIteration errors, which I assume are due to either timeout or running out of preprocessing...

You can increase the number of preprocessed elements.

Drake-Eidukas:
> You can increase the number of preprocessed elements.

Ooh that worked beautifully, thanks @smkuls

progs/jubjub.py
Drake-Eidukas: I agree, It's now a static variable.

amiller:
Some of this is not specific to mpc and could go in `elliptic_curve.py`

Drake-Eidukas:
So, there's an interesting circular dependency issue of `Jubjub` having a shared point operation on it (`contains_shared_point`), so I'm tempted to put that single method in `jubjub.py`, but I think it makes the most sense as a class method on `Jubjub`, so I'm leaving it in `elliptic_curve.py` for now. Thoughts @amiller?

Drake-Eidukas:
@amiller should be good to go now

amiller (approved):

Issue #89, opened 2018-10-21
[travis] Coverage.py warning: No data was collected. (no-data-collected)


sbellem:
Seems to be a recent problem. See https://travis-ci.org/initc3/HoneyBadgerMPC/jobs/440765027#L599 for instance.

A quick possible workaround may be to add `__init__.py` file under the `tests` dir.

#### Related resources:
https://docs.pytest.org/en/latest/goodpractices.html
http://thomas-cokelaer.info/blog/2017/01/pytest-cov-collects-no-data-on-travis/
https://github.com/pytest-dev/pytest-cov/issues/176
https://github.com/pytest-dev/pytest-cov/issues/98

amiller:
Is fixed now? @Drake-Eidukas can you close if confirm this was fixed by the pyroc commit above?

noudald:
@amiller, I referenced #89 in a completely unrelated repository. Because I encountered the same issue as here. GitHub likes to mention these references. To be clear, the pyroc fix commit was not committed to HoneybadgerMPC!

I think the issue is with Travis-CI. However I can confirm that the `__init__.py` workaround works for the pyroc project, so it will possible work for your repository as well.

I hope this helps.

Drake-Eidukas:
So, our recent fix (largely adding ‘__init__.py’) did work to give us coverage on Travis, however, that specific log message is still present. Given that we have coverage working, I'm happy to close this.

PR #235, opened 2019-04-05
Improve Dockerfile style and Remove pytest-sugar from travis

Drake-Eidukas:
just testing effects of caching pip, as installing python dependencies takes a loooong time

### DIFF
--- a/.ci/travis-script.sh
+++ b/.ci/travis-script.sh
@@ -2,10 +2,18 @@
 
 set -ev
 
-BASE_CMD="docker-compose -f .travis.compose.yml run --rm test-hbmpc"
+BASE_CMD="docker-compose -f .travis.compose.yml run test-hbmpc"
 
 if [ "${BUILD}" == "tests" ]; then
-    $BASE_CMD pytest -v --cov=honeybadgermpc --cov-report=term-missing --cov-report=xml
+    $BASE_CMD pytest -v \
+        --cov \
+        --cov-report=term-missing \
+        --cov-report=xml \
+        -Wignore::DeprecationWarning \
+        -nauto
+
+    IMAGE_NAME=$(docker ps -alq --format "{{.Names}}")
+    docker cp $IMAGE_NAME:/usr/src/HoneyBadgerMPC/coverage.xml .
 elif [ "${BUILD}" == "flake8" ]; then
     flake8
 elif [ "${BUILD}" == "docs" ]; then
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,32 @@
 FROM python:3.7.1-stretch
 
-ENV PYTHONUNBUFFERED=1
+# Allows for log messages to be immediately dumped to the 
+# stream instead of being buffered.
+ENV PYTHONUNBUFFERED    1 
 
-RUN apt-get update && apt-get install -y vim tmux
+# Adds cargo to path for rust dependencies
+ENV PATH                /root/.cargo/bin:$PATH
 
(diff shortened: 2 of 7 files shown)


Drake-Eidukas:
I've cleaned up the branch, now this should just result in a cleaner dockerfile (for maintainability, and also follows best practices), as well as (marginally) faster / better CI (it no longer will be polluted by pytest-sugar).

Dockerfile
```diff
@@ -1,14 +1,22 @@
 FROM python:3.7.1-stretch
 
-ENV PYTHONUNBUFFERED=1
-
-RUN apt-get update && apt-get install -y vim tmux
+ENV PYTHONUNBUFFERED    1
```
smkuls, line 5: Since you started working on this, can you add a comment so as to why we need each variable?

setup.py
```diff
@@ -29,7 +29,6 @@
     'pytest-mock',
     'pytest-cov',
     'pytest-env',
-    'pytest-sugar',
     'pytest-xdist',
```
smkuls, line 32: Can you get rid of the `-nauto` in `pytest.ini` as well? You should edit the travis files to make the tests run using `-nauto` and it shouldn't be a default option.

Like we discussed lets not run the tests in parallel by default.  I found both sugar and xdist very annoying to work with. With sugar when I tried to use the `-s` switch it wouldn't spit out the logs immediately at times.

setup.py
```diff
@@ -39,6 +38,7 @@
 DEV_REQUIRES = [
     'ipdb',
     'ipython',
+    'pytest-sugar',
```
smkuls: Ummmm, I don't quite agree with this. This section is for all the packages need for development only. You should make a TEST_OPTIONAL list and then add it there.

smkuls (dismissed):

Dockerfile
smkuls: so the two paths are needed by `charm`. I think `PYTHONBUFFERED` has got to do something with the buffering of stream when spitting data on to the console.

smkuls:
Can you also create an issue to improve the build times of Travis and add the current time it takes? You can close the issue once this gets merged but it would be good to have it as an issue to look back at what was optimized and how it was optimized.

setup.py
Drake-Eidukas: We should probably also document what each of these sections is for as well for people in the future adding dependencies

smkuls: I thought the naming and categorization was pretty obvious given we have a separate tests and dev section. But you are right, some comments would be better :)

Drake-Eidukas:
So, it seems like this has gotten travis to start displaying coverage again (resolving #89), but it required me to suppress `DeprecationWarning` errors to get it to show up in the logs (they were drowning them out)

~More work has to be done to get `codecov` to work, namely copying out the coverage report into travis from docker.~

Drake-Eidukas:
That said, this might be a bit out of scope for this PR, so I can split out the stuff to get coverage to work too if you want

.ci/travis-script.sh
```diff
@@ -2,13 +2,21 @@
 
 set -ev
 
-BASE_CMD="docker-compose -f .travis.compose.yml run --rm test-hbmpc"
```
Drake-Eidukas, line 5: Not sure if we need to actually rm the docker container, but if we don't `--rm` it right away, then we can copy over the coverage report for use by `codecov`

amiller (approved):
I looked through this and reviewed the conversation, looks great to me

PR #236, opened 2019-04-08
Generate random bit in preprocessing

hanyunx:
add `get_bit( )` in `preprocessing.py` and its test.

### DIFF
--- a/tests/test_preprocessing.py
+++ b/tests/test_preprocessing.py
@@ -53,6 +53,24 @@ async def test_get_rand(test_preprocessing):
 
 
 @mark.asyncio
+async def test_get_bit(test_preprocessing):
+    n, t = 4, 1
+    num_bits = 20
+    test_preprocessing.generate("bits", n, t)
+
+    async def _prog(ctx):
+        shares = [test_preprocessing.elements.get_bit(ctx) for _ in range(num_bits)]
+        x = ctx.ShareArray(shares)
+        x_ = await x.open()
+        for i in x_:
+            assert i == 0 or i == 1
+
+    program_runner = TaskProgramRunner(n, t)
+    program_runner.add(_prog)
+    await program_runner.join()
+
+
+@mark.asyncio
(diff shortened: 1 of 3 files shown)


tests/test_preprocessing.py
```diff
@@ -52,6 +52,22 @@
...
+@mark.asyncio
+async def test_get_bit(test_preprocessing):
+    n, t = 4, 1
+    num_bits = 10
+    test_preprocessing.generate("bits", n, t)
+
+    async def _prog(ctx):
+        for _ in range(num_bits):
+            x_sh = test_preprocessing.elements.get_bit(ctx)
+            assert await x_sh.open() == 0 or 1
```
smkuls: Use ShareArray open instead of share open?

```
x = context.ShareArray([i.v for i in test_preprocessing.elements.get_bit(ctx)])
```

smkuls (changes requested):
Rest looks good

Drake-Eidukas:
LGTM,

tests/test_preprocessing.py
Drake-Eidukas: In addition, I think that `a == 0 or 1` isn't quite doing what you want it to do--
```
>>> for a in range(4):
...     print(f"{a} == 0 or 1 => {a == 0 or 1}")
... 
0 == 0 or 1 => True
1 == 0 or 1 => 1
2 == 0 or 1 => 1
3 == 0 or 1 => 1
```
This is effectively: 
```
assert (await x_sh.open() == 0) or 1
```
This will always be `True`, and if `x_sh.open() == 4`, for example, this will still pass the test.

Drake-Eidukas (commented):
Other than what samarth mentioned, LGTM

tests/test_preprocessing.py
smkuls: Nice catch!

smkuls (approved):
LGTM!

PR #237, opened 2019-04-08
Add equality operation for hbMPC

hanyunx:
The equality function is now returning an **opened** boolean result.


tests/fixtures.py
```diff
@@ -82,7 +82,7 @@ def __init__(self):
         self.cache = {}
         self.elements = PreProcessedElements()
 
-    def generate(self, kind, n, t, arg1=None, arg2=None, k=1000):
+    def generate(self, kind, n, t, arg1=None, arg2=None, k=2000):
```
Drake-Eidukas, line 85: I think it would be better to override this in your test cases-- I found a pattern like [this](https://github.com/initc3/HoneyBadgerMPC/pull/233/files#diff-5994374a4e4f9253d67655fb09c077ecR20) to work pretty well.

amiller:
Don't check in the `.DS_Store` files, also these should be added to `.gitignore` anyway

tests/test_equality.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+@mark.asyncio
+async def test_equality(test_preprocessing, galois_field):
+    from honeybadgermpc.mpc import TaskProgramRunner
+    from honeybadgermpc.mixins import BeaverTriple, MixinOpName
+    from honeybadgermpc.equality import equality
+
+    n, t = 3, 1
+    test_preprocessing.generate("zeros", n, t)
```
Drake-Eidukas, line 11: You could achieve the previous comment pretty cleanly like: 
```
[test_preprocessing.generate(x, n, t, k=2000) for x in ['zeros', 'rands', 'bits', 'triples']]
```

Also,
```
to_generate =  ['zeros', 'rands', 'bits', 'triples']
map(lambda x: test_preprocessing.generate(x, n, t, k=2000), to_generate)
```
does this a bit more verbosely, but is is clearer about how it's not actually trying to build up a list.

honeybadgermpc/equality.py
```diff
@@ -0,0 +1,107 @@
+from honeybadgermpc.mpc import PreProcessedElements, Subgroup, TaskProgramRunner
+import logging
+import asyncio
+from honeybadgermpc.field import GF
+from honeybadgermpc.mixins import BeaverTriple, MixinOpName
+SECURITY_PARAMETER = 32
```
amiller, line 6: This is a parameter that can passed as an argument to the equality function. It would also make sense to define such parameters in a config file, since we will often set `STATISTICAL_SECURITY_PARAMETER` separately from the computational security parameters

honeybadgermpc/equality.py
```diff
@@ -0,0 +1,107 @@
...
+        if legendre == 1:
+            xj = (1 / field(2)) * (bj + context.Share(1))
+        elif legendre == -1:
+            xj = (-1) * (1 / field(2)) * (bj - context.Share(1))
+        else:
+            gen_test_bit()
+
+        return xj
+
+    x = [await gen_test_bit() for _ in range(k)]
```
amiller: This is clearly sequential. This overall operation can be optimized (at a later step) to run all the trials in parallel

amiller, line 64: We now have random bits in preprocessing. We don't need to run gen_test_bit, we can just do `preprocessing.get_minus_one_one`

honeybadgermpc/equality.py
```diff
@@ -0,0 +1,107 @@
...
+
+    # Take the product (this is here the same as the "and") of all
+    # the x'es
+    while len(x) > 1:
```
amiller, line 68: This is `SECURITY_PARAMETER` sequential products, but by doing a tree we can make it only `log(SECURITY_PARAMETER)` rounds of batch products

honeybadgermpc/equality.py
```diff
@@ -0,0 +1,107 @@
...
+        x.append(await (x.pop(0) * x.pop(0)))
+
+    res = await x[0].open()
+    return bool(res)
+
+
+async def test_equality(context):
```
amiller, line 75: we don't need this now that we have `test_equality.py`

amiller (changes requested):
Requested changes:
- get rid of unnecessary files
- squash to one commit
- remove unnecessary test

amiller:
@hanyunxu  I'm waiting for you to respond to the above requested changes before we merge ^^

honeybadgermpc/equality.py
Drake-Eidukas: Assuming you mean `preprocessing.get_one_minus_one_rand`, it actually breaks this-- the results from `gen_test_bit` contain the values `{1}, {52435875175126190479447740508185965837690552500527637822603658699938581184511}`, whereas `preprocessing.get_one_minus_one_rand` generates `{1}, {52435875175126190479447740508185965837690552500527637822603658699938581184512}`.

Drake-Eidukas: ie `get_test_bit` generates values that are `context.field(1)` and `context.field(-2)`

Drake-Eidukas:
@amiller  I discussed with @hanyunxu and she's super bogged down with thesis work-- [here's a PR](https://github.com/initc3/HoneyBadgerMPC/pull/247) that incorporates just about everything you asked for (aside from the stuff with `get_one_minus_one_rand`.

PR #238, opened 2019-04-08
Precompute

tyurek:
Code to do precomputation for elliptic curve point scalar multiplication (currently only for G1). Precomputation is tweakable through "precomputation levels" which determine how many elements to precompute. Specifically, (2^{level} - 1) * ceil(255/{level}) elements are precomputed, which should result in a speedup by a factor of roughly {level}. Note that the number of precompute elements is exponential in {level}, and so there's probably a point where this relation breaks down and memory accesses start to become more significant as the precomputed data can no longer be stored in cache. 

Precomputation is integrated into G1, G2, and GT

### DIFF
--- a/pairing/src/lib.rs
+++ b/pairing/src/lib.rs
@@ -27,15 +27,47 @@ extern crate rand;
 pub mod tests;
 
 pub mod bls12_381;
-use bls12_381::{G1, G2, Fr, Fq, Fq2, Fq6, Fq12, FqRepr};
+use bls12_381::{G1, G2, Fr, Fq, Fq2, Fq6, Fq12, FqRepr, FrRepr};
 mod wnaf;
 pub use self::wnaf::Wnaf;
 
 use ff::{Field,  PrimeField, PrimeFieldDecodingError, PrimeFieldRepr, ScalarEngine, SqrtField};
 use std::error::Error;
 use std::fmt;
+use std::io::{self, Write};
 use rand::{Rand, Rng, SeedableRng, XorShiftRng};
 
+fn hex_to_bin (hexstr: &String) -> String
+{
+    let mut out = String::from("");
+    let mut bin = "";
+    //Ignore the 0x at the beginning
+    for c in hexstr[2..].chars()
+    {
+        match c
+        {
+            '0' => bin = "0000",
+            '1' => bin = "0001",
+            '2' => bin = "0010",
+            '3' => bin = "0011",
+            '4' => bin = "0100",
+            '5' => bin = "0101",
+            '6' => bin = "0110",
+            '7' => bin = "0111",
+            '8' => bin = "1000",
+            '9' => bin = "1001",
+            'A'|'a' => bin = "1010",
+            'B'|'b' => bin = "1011",
+            'C'|'c' => bin = "1100",
+            'D'|'d' => bin = "1101",
+            'E'|'e' => bin = "1110",
+            'F'|'f' => bin = "1111",
+            _ => bin = ""
+        }
+        out.push_str(bin);
+    }
+    out
+}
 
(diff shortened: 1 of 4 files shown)


amiller:
This is not building in ci, let us know if you want help diagnosing

amiller (changes requested):
This looks fine. However, I notice there are no tests. In keeping with the code quality of the rest of the branch, can you add test cases to say `tests/test_betterpairing.py` that exercise the new precompute code?

pairing/src/lib.rs
```diff
@@ -223,20 +305,86 @@ impl PyG1 {
...
+
     pub fn projective(&self) -> PyResult<String> {
         Ok(format!("({}, {}, {})",self.g1.x, self.g1.y, self.g1.z))
     }
+
     pub fn __str__(&self) -> PyResult<String> {
         Ok(format!("({}, {})",self.g1.into_affine().x, self.g1.into_affine().y))
     }
 
+    fn preprocess(&mut self, level: usize) -> PyResult<()> {
```
Drake-Eidukas, line 325: It's kinda nitpicky, but I think stylistically you should add some whitespace and reorder some lines to help readability. For example on this method--
```
fn preprocess(&mut self, level: usize) -> PyResult<()> {
    self.pplevel = level;
    
    //Everything requires a different kind of int (and only works with that kind)
    let mut base: u64 = 2;
    
    //calling pow on a u64 only accepts a u32 parameter for reasons undocumented
    base = base.pow(level as u32);
    
    let ppsize = (base as usize - 1) * (255 + level - 1) / (level);
    self.pp = Vec::with_capacity(ppsize);
    self.pp.push(self.g1.clone());

    for i in 1..base-1
    {
        //Yes, I really need to expicitly cast the indexing variable...
        let mut next = self.pp[i as usize -1].clone();
        next.add_assign(&self.g1);
        self.pp.push(next);
    }

    //FrRepr::from only takes a u64
    let factor = Fr::from_repr(FrRepr::from(base)).unwrap();
    
    //(x + y - 1) / y is a way to round up the integer division x/y
    for i in base-1..(base - 1) * (255 + level as u64 - 1)/(level as u64) {
        let mut next = self.pp[i as usize - (base-1) as usize].clone();
        //Wait, so add_assign takes a borrowed object but mul_assign doesn't?!?!?!?
        next.mul_assign(factor);
        self.pp.push(next);
    }
    
    //It's not really Ok. This is terrible.
    Ok(())
}
```

I think formalizing some of the comments would be a good idea too

pairing/src/lib.rs
```diff
@@ -141,20 +175,30 @@ impl PyG1 {
...
+            pp: Vec::new(),
+            pplevel : 0
         })
     }
 
     fn rand(&mut self, s1: u32, s2: u32, s3: u32, s4: u32) -> PyResult<()>{
         let mut rng = XorShiftRng::from_seed([s1,s2,s3,s4]);
         let g = G1::rand(&mut rng);
         self.g1 = g;
+        if self.pplevel != 0{
```
Drake-Eidukas: I think it would be cleaner if you added a space after the conditional check in the `if` statements--
```suggestion
if self.pplevel != 0 {
...
}
```

Drake-Eidukas (commented):
few stylistic suggestions

amiller:
I'm happy to accept this now, but we also need to add more tests. The rust library itself includes a variety of tests, and I would like to have a similar level of coverage. It would also be good to double check that at least every operation is covered by the tests. There are quite a lot of conditional branches based on different types of inputs, lmul vs rmul, etc., and I don't think the coverage is too full yet.

amiller (approved):

Drake-Eidukas:
@tyurek  Can you squash these commits and rebase?

PR #240, opened 2019-04-09
Add wait_for_all to simulate non robust mode

smkuls:
* Add `wait_for_all` to simulate non robust mode wherein we wait for all
values during reconstruction.
* Demonstrate using test cases how we can pass faulty values and things
continue to work during reconstruction.

### DIFF
--- a/honeybadgermpc/batch_reconstruction.py
+++ b/honeybadgermpc/batch_reconstruction.py
@@ -91,11 +91,13 @@ def list_transpose(lst):
 
 
 async def incremental_decode(receivers, encoder, decoder, robust_decoder, batch_size, t,
-                             n):
+                             n, f, wait_for_all=False):
     inc_decoder = IncrementalDecoder(encoder, decoder, robust_decoder,
                                      degree=t, batch_size=batch_size,
-                                     max_errors=t)
+                                     max_errors=f)
 
+    if wait_for_all:
+        inc_decoder._min_points_required = lambda: n
     async for idx, d in fetch_one(receivers):
         inc_decoder.add(idx, d)
         if inc_decoder.done():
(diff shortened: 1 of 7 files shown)


honeybadgermpc/mpc.py
```diff
@@ -85,26 +85,33 @@ def __init__(self, sid, n, t, myid, pid, send, recv, prog, config, **prog_args):
...
+        degree = sharearray.t
 
         def _send(j, o):
             (tag, share) = o
             self.send(j, (tag, shareid, share))
         _recv = self._sharearray_buffers[shareid].get
-
+        # Maximum number of errors to tolerate. This needs to be
+        # `t` to ensure that we are in the robust setting.
+        c = self.t
```
smkuls, line 109: This would make the reconstruction wait for at least **degree+t+1** points. This corresponds to the maximum number of errors that can be tolerated.

If I understand you correctly, the `f` that we discussed is just a way to send faulty values and is completely independent of how you want to do batch reconstruction since in order to be robust you'd want to wait for at least **degree+t+1** points.

honeybadgermpc/batch_reconstruction.py
```diff
@@ -104,13 +106,14 @@ def list_transpose(lst):
...
-                            use_fft=False, debug=False):
+async def batch_reconstruct(secret_shares, p, t, n, f, myid, send, recv, config=None,
+                            use_fft=False, debug=False, wait_for_all=False):
     """
     args:
       shared_secrets: an array of points representing shared secrets S1 - SB
       p: field modulus
       t: degree t polynomial
+      f: number of faults to tolerate
       n: total number of nodes n >= 3t+1
```
amiller, line 117: this comment should now be.... `n >= t + 2f+1` for robust decoding, `n >= t + f + 1` for malicious secure decoding

rahulgovind: What is malicious secure decoding?

amiller: it guarantees it will not output a *wrong* answer if `f` nodes are malicious, but it may also fail to produce any answer

Issue #25, opened 2018-08-24
Symmetric Cryptography primitives in honeybadgermpc


amiller:
We should have some programs ready and standing by that implement

~- [ ] SHA256~
~- [ ] AES~

Then we can use these for symmetric encryption.

One approach is to build an MPC program that interprets a boolean circuit. Then we can use existing boolean circuits from other projects, such as EMP toolkit.

This is a prerequisite for the "supply chain" demonstration. Samarth began working on this in a different repo but it is still unfinished.

amiller:
AES is also implemented in Viff and needs to be ported over. https://github.com/amiller/viff/blob/master/viff/aes.py

Relies on #27 comparisons and other operations like bit decomposition.

They also include the GF(2^256) field but we will only be using the arithmetic field for now

amiller:
Deprioritizing this for now, for the supply chain application we may be able to just use equality testing and jubjub #23 instead

amiller:
This is reopened and we're going to implement 
- [x] MiMC encryption instead https://eprint.iacr.org/2016/492

PR #241, opened 2019-04-09
MiMC function implementation in MPC

hanyunx:
Implementation of F_MiMIC(x, k )
https://eprint.iacr.org/2016/542.pdf Section 5

### DIFF
--- a/tests/progs/test_mimc.py
+++ b/tests/progs/test_mimc.py
@@ -1,0 +1,32 @@
+from pytest import mark
+from honeybadgermpc.field import GF
+from honeybadgermpc.mpc import Subgroup
+from honeybadgermpc.progs.mimc import mimc_mpc, mimc_plain
+from honeybadgermpc.progs.mixins.share_arithmetic import BeaverMultiply
+
+MIXINS = [BeaverMultiply()]
+PREPROCESSING = ['rands', 'triples', 'zeros', 'cubes']
+n, t = 3, 1
+k = 10000
+
+
+@mark.asyncio
+async def test_mimc(test_preprocessing, test_runner):
+
+    async def _prog(context):
+        x = test_preprocessing.elements.get_zero(context)
+        field = GF(Subgroup.BLS12_381)
+        key = field(15)
+
+        # Compute F_MiMC_mpc
+        mm = await mimc_mpc(context, x, key)
+
+        # open x, then compute F_MiMC_plain
+        x_open = await x.open()
+        mp = mimc_plain(x_open, key)
+
+        # Compare the MPC evaluation to the plain one
+        mm_ = await mm.open()
+        assert mm_ == mp
+
+    await test_runner(_prog, n, t, PREPROCESSING, k, MIXINS)
--- a/honeybadgermpc/progs/mimc.py
+++ b/honeybadgermpc/progs/mimc.py
@@ -1,0 +1,38 @@
+from math import log, ceil
+from honeybadgermpc.mpc import PreProcessedElements, Subgroup
+
+# ROUND: iteration time of MiMC encryption function,
+# In BLS12_381, r = ceil(log(p, 3)) = 161
+ROUND = ceil(log(Subgroup.BLS12_381, 3))
+
+
+async def mimc_mpc(context, x, k):
+    """
+    MiMC block cipher encryption encrypts message x with secret key k,
+    where x and k are elements of F_p
+    See: https://eprint.iacr.org/2016/542.pdf
+    """
+    pp_elements = PreProcessedElements()
+
+    # def cubing_share(): [x] -> [x^3]
+    async def cubing_share(x):
+        r1, r2, r3 = pp_elements.get_cube(context)
+        y = await (x - r1).open()
+        # [x^3] = 3y[r^2] + 3y^2[r] + y^3 + [r^3]
+        x3 = 3*y*r2 + 3*(y**2)*r1 + y**3 + r3
+        return x3
+
+    # iterating the round function ROUND times
+    inp = x
+    for ctr in range(ROUND):
+        inp = await cubing_share(inp + (k + ctr))
+
+    return inp + k
+
+
+def mimc_plain(x, k):
+    inp = x
+    for ctr in range(ROUND):
+        inp = (inp + (k + ctr)) ** 3
+
+    return inp + k


honeybadgermpc/mimc.py
```diff
@@ -0,0 +1,64 @@
...
+
+    # def cube(): [x] -> [x^3]
+    async def cube(x):
+        # first compute a tuple ([r], [r^2], [r^3])
+        r = pp_elements.get_rand(context)
+        r2 = await (r * r)
+        r3 = await (r * r2)
+
+        y = await (x-r).open()
+        x3 = 3*y*r2 + 3*(y**2)*r + y**3 + r3
```
Drake-Eidukas: I think documenting this with a comment like `# [x^3] = 3y[r^2] + 3y^2[r] + y^3 + [r^3]` would be a good idea for future maintenance.

honeybadgermpc/mimc.py
```diff
@@ -0,0 +1,64 @@
...
+from honeybadgermpc.mpc import PreProcessedElements, TaskProgramRunner, Subgroup
+from honeybadgermpc.field import GF
+from honeybadgermpc.mixins import BeaverTriple, MixinOpName
+
+# ROUND: iteration time of MiMC encryption function, r = ceil(log(p, 3)) = 161
+ROUND = ceil(log(Subgroup.BLS12_381))
+
+
+async def mimc(context, x, k):
+    # message x \in Fp, secret key k \in Fp
```
Drake-Eidukas: I think using a docstring comment here would be better, as well as including the link to the algorithm.
```suggestion
"""
MiMC block cipher encryption encrypts message x with secret key k,
where x and k are elements of F_p
See: https://eprint.iacr.org/2016/542.pdf
"""
```

Drake-Eidukas (commented):
Just saw you tagged this as [NOT FOR MERGE] but already typed these comments out... so I'll submit them anyways

honeybadgermpc/mimc.py
```diff
@@ -0,0 +1,64 @@
...
+
+    pp_elements = PreProcessedElements()
+
+    # def cube(): [x] -> [x^3]
+    async def cube(x):
+        # first compute a tuple ([r], [r^2], [r^3])
+        r = pp_elements.get_rand(context)
+        r2 = await (r * r)
+        r3 = await (r * r2)
```
amiller: We should ideally be getting the entire [r] [r^2] [r^3] from preprocessing, similar to powermix

honeybadgermpc/mimc.py
```diff
@@ -0,0 +1,64 @@
...
+        return x3
+
+    inp = x
+    for ctr in range(ROUND):
+        inp = await cube(inp + (k + ctr))
+
+    return inp + k
+
+
+async def test_mimc(context):
```
amiller: We should have a non-MPC implementation of MiMC as well as our encrypt/decrypt mode to use for testing... similar to what we have for Jubjub

honeybadgermpc/mimc.py
hanyunx: Made it as a separate PR.

honeybadgermpc/mimc.py
hanyunx: Resolved.

honeybadgermpc/mimc.py
hanyunx: Resolved.

honeybadgermpc/mimc.py
hanyunx: Moved `mimc.py` to `progs/`

tests/progs/test_mimc.py
```diff
@@ -0,0 +1,29 @@
...
+    test_preprocessing.generate("triples", n, t)
+    test_preprocessing.generate("cubes", n, t)
+
+    async def _prog(context):
+        x = test_preprocessing.elements.get_zero(context)
+        field = GF(Subgroup.BLS12_381)
+        key = field(15)
+
+        mm = await mimc(context, x, key)
+        print("F_MiMC(x, k) = ", mm)
```
amiller: a print statement doesn't count as a test.....
can you add a "plain" version of mimc (not MPC) and compare the MPC evaluation to the plain one?

hanyunxu: I'll complete the test file after the implementation if encrypt & decrypt...

hanyunx: Added.

Drake-Eidukas:
(tackles #25)

honeybadgermpc/preprocessing.py
```diff
@@ -78,6 +80,18 @@ def generate_triples(self, k, n, t):
             polys.append(self.poly.random(t, c))
         self._write_polys(PreProcessingConstants.TRIPLES_FILE_NAME_PREFIX, n, t, polys)
 
+    def generate_cubes(self, k, n, t):
```
Drake-Eidukas: Strip this out of the PR / rebase with `dev`-- this has already been merged

progs/mimc.py
```diff
@@ -0,0 +1,38 @@
+from math import log, ceil
+from honeybadgermpc.mpc import PreProcessedElements, Subgroup
+
+# ROUND: iteration time of MiMC encryption function,
+# In BLS12_381, r = ceil(log(p, 3)) = 161
+ROUND = ceil(log(Subgroup.BLS12_381))
+
+
+async def mimc_mpc(context, x, k):
```
Drake-Eidukas: I think naming these methods `mimc_encrypt` and `mimc_encrypt_plain` would be better.

I think the way I would do it would be as follows--
- `mimc_encrypt` and `mimc_decrypt` functions, which are high level functions that handle both plain and shared mimc implementations.
    - These would then call on either `__mimc_encrypt_shared` or `__mimc_encrypt_plain` and `__mimc_decrypt_shared` or `__mimc_decrypt_plain` respectively based on whether or not the methods are invoked with secret-shared values or plain values.

progs/mimc.py
```diff
@@ -0,0 +1,38 @@
...
+    """
+    MiMC block cipher encryption encrypts message x with secret key k,
+    where x and k are elements of F_p
+    See: https://eprint.iacr.org/2016/542.pdf
+    """
+    pp_elements = PreProcessedElements()
+
+    # def cubing_share(): [x] -> [x^3]
+    async def cubing_share(x):
```
Drake-Eidukas: I think this, alongside other common operations (such as squaring), could belong more in an MPC utility file. This will likely tie into however we end up refactoring mixins and whatnot.

Drake-Eidukas (commented):
Things to do: 
- [x] rebase with dev
- [ ] strip out content relating to `get_cube`'s implementation
- [ ] Consider structuring your MiMC encryption and decryption as suggested-- I think it's cleaner, but then we'll have to deal with the (minimal) overhead of type-checking.

progs/mimc.py
hanyunx: I think `mimc` is a function used in both encryption and decryption process. I'll have other encryption/decryption function which will call `mimc_mpc`, `mimc_plain` here.

amiller:
@Drake-Eidukas can we pull this independently of #254 ? 
This is the only one you need for garbled circuit stuff. #254 (just renamed) is really about public key encryption.

amiller (approved):
I think this is ready to go

Issue #234, opened 2019-04-05
Update Share multiplication to follow VIFF's data flow design


Drake-Eidukas:
Previous work has implemented most of the share operations according to #44, however,  `__mul__` for shares doesn't follow the design. When designing share inversion, I had to write code like `sig = await (await(x*r)).open()`, which is because x*r returns a coroutine when it should return a `GFElementFuture` instead. The code should instead look like `await context.Share(x*r).open()` This also requires that `context.Share` takes a `GFElementFuture` and call open once it is resolved.

@smkuls can give more info if required-- he raised this to my attention.

Drake-Eidukas:
Main issue is that the dataflow should be as follows:
```
await FieldFuture -> Field
await ShareFuture -> Share
Share + Share -> Share
Share * Share -> ShareFuture
ShareFuture + ShareFuture -> ShareFuture
ShareFuture * ShareFuture -> ShareFuture
Share.open() -> FieldFuture
ShareFuture.open() -> FieldFuture
```
Currently, `Share * Share -> coroutine`, and we need to implement divide in this manner as well.

PR #246, opened 2019-04-14
Refactor MPC to follow VIFF's data flow model

Drake-Eidukas:
This PR refactors the way that we work with arithmetic on `Share`s-- by adding more arithmetic operations to `Share` and formalizing the type checking, this makes writing MPC applications with `Share` and `ShareFuture` way easier. This also generally improves codestyle everywhere I touched. 

Potential TODO: more tests

This tackles #234

### DIFF
--- a/honeybadgermpc/mixins.py
+++ b/honeybadgermpc/mixins.py
@@ -53,9 +53,9 @@ class Inverter(MixinBase):
         assert isinstance(x, context.Share)
 
         r = MixinBase.pp_elements.get_rand(context)
-        sig = await(await(x * r)).open()
+        sig = await (x*r).open()
 
-        return await(r * context.Share(1/sig))
+        return r * (1/sig)
 
     @staticmethod
     async def invert_share_array(context, x):
@@ -79,7 +79,8 @@ class DoubleSharing(MixinBase):
 
         r_t, r_2t = MixinBase.pp_elements.get_double_share(context)
         diff = await (x_2t - r_2t).open()
-        return context.Share(await (r_t + diff).open())
+
+        return r_t + diff
 
     @staticmethod
     async def reduce_degree_share_array(context, x_2t):
--- a/honeybadgermpc/mpc.py
+++ b/honeybadgermpc/mpc.py
@@ -67,12 +67,21 @@ class Mpc(object):
         self.Share, self.ShareArray = share_in_context(self)
 
     async def open_share(self, share):
+        """ Given secret-shared value share, open the value by
+        broadcasting our local share, and then receive the likewise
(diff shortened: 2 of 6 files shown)


honeybadgermpc/mixins.py
```diff
@@ -53,9 +53,9 @@ class Inverter(MixinBase):
         assert isinstance(x, context.Share)
 
         r = MixinBase.pp_elements.get_rand(context)
-        sig = await(await(x * r)).open()
+        sig = await (x*r).open()
 
-        return await(r * context.Share(1/sig))
+        return r * (1/sig)
```
amiller, line 58: good simplification

honeybadgermpc/mpc.py
```diff
@@ -161,129 +188,254 @@ def _send(j, o):
...
     class Share(object):
+        """Represents a local share of a secret-shared GFElement
+        """
+
         def __init__(self, v, t=None):
             # v is the local value of the share
             if type(v) is int:
                 v = context.field(v)
-            assert type(v) is GFElement
+            assert isinstance(v, (GFElement, GFElementFuture))
```
amiller, line 243: Interesting. So is the idea that a `Share` object can actually be backed by a `GFElementFuture` rather than an already-resolved field element? Does this mean there's no need for a `ShareFuture` object? It's not obvious to me what are all the consequences of this, but interesting idea

amiller (changes requested):
So I'm happy to accept this change on the basis that all the tests pass.
But what I would really like to see is less coupling between the "dataflow programming" model and the `mpc.py` engine itself. What I would like is for the MPC interface in `mpc.py` to be very simple, just exposing `open_share` and `open_share_array` and that's it. The dataflow programming model could then be provided separately. The significance of this is that we could potentially support multiple variations of the dataflow programming model, provide backward compatibility if we want to change it, etc. 
Could you check if it's possible simply to take all the `Share` object and `ShareArray` objects and move them out of mpc.py, so it's possible to run a program that doesn't rely on them at all?

honeybadgermpc/mpc.py
Drake-Eidukas: So, the main idea is that `ShareFuture` objects allow you to stack up `Share` arithmetic operations on each other on potentially unresolved values. I'm not entirely sure when you'd want to explicitly use a share with a `GFElementFuture`, but @smkuls  mentioned that it would be a good method to add in during the refactor.

Drake-Eidukas:
I think it would be possible to do what you mentioned @amiller, but I'm thinking that it might be more appropriate to put that in the refactor that also moves stuff out of `mixins` so that this PR can stay more focused in scope.

amiller (approved):

Issue #27, opened 2018-08-24
Comparisons and other operations from Viff


amiller:
Viff comes with support for several useful operations, such as comparing greater than / less than / equality for integers, splitting integers into parts, etc. see this for example
https://github.com/amiller/viff/blob/master/apps/compare.py

~- [ ] Bit decomposition~
- [x] Equality testing
- [x] Comparison

amiller:
Breaking this into tasks:
- [ ] Bit decomposition
- [ ] Equality testing
- [ ] Comparison

hanyunx:
For Equality Testing, a potential way is [[NO07]](https://www.iacr.org/archive/pkc2007/44500343/44500343.pdf). It provides the following solutions:
- Simplified BD with complexity of `(2+23) R + (93 L + 47L logL) M`.
- Without BD
    - Deterministic Equality Test Protocol with complexity of `(2+6) R + 81L M` ([ongoing](https://github.com/hanyunxu/viff/blob/equality/viff/equality_dtm.py))
    - Probabilistic Equality Test Protocol  with complexity of `(2+2) R + 12k M`, k - secure parameter
        - p = 3 mod 4  -> `(-1 / p) = -1`,  `bj \in { -1, 1 }`  _which is implemented in viff_
        - p = 5 mod 8 -> `(2 / p) = -1`, `bj' = -(bj - 3) / 2  \in { 2, 1 }`
        - To extend to arbitrary primes -> `(y/p) = -1`
            - [ ] p = 1 mod 8 -> let `y = 5`, s.t. `(y / p) = -1`, `bj' = -2 * (bj - 3/2)  \in { 5, 1 }`

amiller:
https://github.com/initc3/HoneyBadgerMPC/blob/4dc26133030aebe31d70b239d502fac6a8c8afa5/honeybadgermpc/passive.py#L393 work in progress

PR #247, opened 2019-04-14
Implement equality operation for shares

Drake-Eidukas:
I discussed with Hanyun, and she's super bogged down with thesis work, so I implemented some of the refactors suggested from @amiller  in #237

### DIFF
--- a/honeybadgermpc/equality.py
+++ b/honeybadgermpc/equality.py
@@ -1,0 +1,68 @@
+from honeybadgermpc.mpc import PreProcessedElements
+import asyncio
+
+
+async def equality(context, p_share, q_share, security_parameter=32):
+    assert isinstance(p_share, context.Share)
+    assert isinstance(q_share, context.Share)
+
+    pp_elements = PreProcessedElements()
+    diff_a = p_share - q_share
+
+    def legendre_mod_p(a):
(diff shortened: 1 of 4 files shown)


honeybadgermpc/equality.py
```diff
@@ -0,0 +1,64 @@
...
+
+    x = await asyncio.gather(*[gen_test_bit() for _ in range(security_parameter)])
+
+    # Take the product (this is here the same as the "and") of all
+    # the x'es
+    while len(x) > 1:
+        x_first, x_left = context.ShareArray(x[::2]), context.ShareArray(x[1::2])
+        x = (await (x_first * x_left))._shares
+
+    return await x[0].open() != 0
```
amiller: just `return x[0]` here. Equality should return a secret share of `[a == b]`, it's up to the caller if they wish to publicly reconstruct it

amiller (changes requested):

honeybadgermpc/equality.py
Drake-Eidukas: Sounds good.

Drake-Eidukas:
@amiller  should be good to go now

amiller (approved):

Drake-Eidukas:
This resolves one of the three points in #27

PR #248, opened 2019-04-14
Fix a bug in AVID that cause issues with multiple retrieval

nqyy:
I found this bug when doing implicate handling. We need to keep track of the index we are currently retrieving. If not, the message from the retrieval queue may not be the message with the exact index that we want to retrieve. The whole retrieval queue will be a mess.


amiller (approved):
Makes sense. In case multiple retrieves are done concurrently, let the responses carry a tag of which index they pertain to

Drake-Eidukas:
Force rebased

PR #249, opened 2019-04-15
Add mult_share function for <share> * <plain_point>

hanyunx:
The input is a bitwise shared value,` [a1, a2, a3, ... ak]` where `ai` is the shared value of 0 or 1, and a plain point. 
Each of the bit share is opened in the iterator to check if it's equal to 1, which looks not safe. But given that the equality application hasn't been merged, let's take it as a tentative solution.

### DIFF
--- a/honeybadgermpc/progs/jubjub.py
+++ b/honeybadgermpc/progs/jubjub.py
@@ -1,6 +1,5 @@
-from honeybadgermpc.elliptic_curve import Jubjub, Point
+from honeybadgermpc.elliptic_curve import Jubjub, Point, Ideal
 from honeybadgermpc.mpc import Mpc
-import asyncio
 
 
 class SharedPoint(object):
@@ -19,13 +18,10 @@ class SharedPoint(object):
         self.xs = xs
         self.ys = ys
 
-    async def __on_curve(self) -> bool:
+    async def _on_curve(self) -> bool:
         """
         Checks whether or not the given shares for x and y correspond to a
         point that sits on the current curve
-
-        WARNING: This method currently leaks information about the shared point--
-                 We need to use share equality testing instead
         """
         x_sq = self.xs * self.xs
         y_sq = self.ys * self.ys
@@ -36,17 +32,7 @@ class SharedPoint(object):
         # 1 + dx^2y^2
         rhs = self.context.field(1) + self.curve.d * x_sq * y_sq
 
-        # TODO: use share equality to prevent the leaking of data
-        lhs, rhs = await asyncio.gather(lhs.open(), rhs.open())
-        return lhs == rhs
-
-    async def __init(self):
-        """asynchronous part of initialization via create or from_point
-        """
-        if not await self.__on_curve():
-            raise ValueError(
-                f"Could not initialize Point {self}-- \
-                does not sit on given curve {self.curve}")
+        return await (await lhs == await rhs).open()
 
     @staticmethod
     async def create(context: Mpc, xs, ys, curve=Jubjub()):
@@ -54,18 +40,22 @@ class SharedPoint(object):
             creates the given point
         """
         point = SharedPoint(context, xs, ys, curve)
-        await point.__init()
+        if not await point._on_curve():
+            raise ValueError(
+                f"Could not initialize Point {point}-- \
+                does not sit on given curve {point.curve}")
 
         return point
 
     @staticmethod
-    async def from_point(context: Mpc, p: Point) -> 'SharedPoint':
+    def from_point(context: Mpc, p: Point) -> 'SharedPoint':
         """ Given a local point and a context, created a shared point
         """
         if not isinstance(p, Point):
             raise Exception(f"Could not create shared point-- p ({p}) is not a Point!")
 
-        return await(SharedPoint.create(context, context.Share(p.x), context.Share(p.y)))
+        return SharedPoint(context, context.Share(p.x),
+                           context.Share(p.y), curve=p.curve)
 
     def __str__(self) -> str:
         return f"({self.xs}, {self.ys})"
@@ -73,11 +63,11 @@ class SharedPoint(object):
     def __repr__(self) -> str:
         return str(self)
 
-    async def neg(self):
-        return await SharedPoint.create(self.context,
-                                        self.context.field(-1) * self.xs,
-                                        self.ys,
-                                        self.curve)
+    def neg(self):
+        return SharedPoint(self.context,
+                           -1 * self.xs,
+                           self.ys,
+                           self.curve)
 
     async def add(self, other: 'SharedPoint') -> 'SharedPoint':
         if isinstance(other, SharedIdeal):
@@ -104,10 +94,10 @@ class SharedPoint(object):
         # y3 = ((y1*y2) + (x1*x2)) / (1 - d*x1*x2*y1*y2)
         y3 = (y_prod + x_prod) / (one - d_prod)
 
-        return await SharedPoint.create(self.context, x3, y3, self.curve)
+        return SharedPoint(self.context, x3, y3, self.curve)
 
     async def sub(self, other: 'SharedPoint') -> 'SharedPoint':
-        return await self.add(await other.neg())
+        return await self.add(other.neg())
 
     async def mul(self, n: int) -> 'SharedPoint':
         # Using the Double-and-Add algorithm
@@ -116,13 +106,12 @@ class SharedPoint(object):
             raise Exception("Can't scale a SharedPoint by something which isn't an int!")
 
         if n < 0:
-            negated = await self.neg()
-            return await negated.mul(-n)
+            return await self.neg().mul(-n)
         elif n == 0:
             return SharedIdeal(self.curve)
 
         current = self
-        product = await SharedPoint.from_point(self.context, Point(0, 1, self.curve))
+        product = SharedPoint.from_point(self.context, Point(0, 1, self.curve))
 
         i = 1
         while i <= n:
@@ -147,7 +136,7 @@ class SharedPoint(object):
             return SharedIdeal(self.curve)
 
         current = self
-        product = await SharedPoint.from_point(self.context, Point(0, 1, self.curve))
+        product = SharedPoint.from_point(self.context, Point(0, 1, self.curve))
 
         i = 1 << n.bit_length()
         while i > 0:
@@ -173,10 +162,7 @@ class SharedPoint(object):
         x = (2 * x_ * y_) / x_denom
         y = (y_sq - ax_sq) / (self.context.field(2) - x_denom)
 
-        return await SharedPoint.create(self.context,
-                                        x,
-                                        y,
-                                        self.curve)
+        return SharedPoint(self.context, x, y, self.curve)
 
 
 class SharedIdeal(SharedPoint):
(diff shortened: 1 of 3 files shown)


tests/progs/test_jubjub.py
```diff
@@ -191,3 +191,20 @@ def test_basic_point_functionality():
             await p1.montgomery_mul(5))
 
     await run_test_prog(_prog, test_preprocessing)
+
+
+@mark.asyncio
+async def test_share_mul(test_preprocessing):
```
Drake-Eidukas: Try adding more preprocessing elements to get this test to pass-- 
`run_test_prog(_prog, test_preprocessing, k=5000)`

Drake-Eidukas: (that's the solution I've used for this error in the past)

hanyunx: Will try, thanks!

progs/jubjub.py
```diff
@@ -226,3 +226,26 @@ def __str__(self):
 
     async def double(self):
         return self
+
+
+async def share_mul(context, p: Point, x: list) -> 'SharedPoint':
```
Drake-Eidukas: I feel like this might be more appropriately placed in the `Point` class, so that we could do `multiplied = p.share_mul([x])` and it will work more natively with how the existing operations are structured.

Drake-Eidukas: Also, the type signature would be better as `async def share_mul(context: Mpc, p: Point, x: list) -> SharedPoint`-- I only have the quotes in the other examples because you can't use a type signature of a class within the class.

hanyunx: I was considering placing it in Point class, but the problem is the function needs the `SharedPoint` class which imports `Point` class itself.

Drake-Eidukas: Oh, good point.

tests/progs/test_jubjub.py
```diff
@@ -216,3 +210,22 @@ def test_basic_point_functionality():
...
```
Drake-Eidukas: Can you add another operation to this unit test? I'd like to see that `TEST_POINTS[0] * <some number> == TEST_POINTS[0]` (TEST_POINTS[0] represents zero)

progs/jubjub.py
Drake-Eidukas: BTW, you could actually put this in the `Point` class if you imported `SharedPoint` within the method-- 
```
async def share_mul(self, context, x: list) -> 'Point':
        """
        The multiplication of the share of a field element and a point
        e.g. [X] <- [x]G, where G is a point on the given elliptic curve
        x is the bitwise shared value,
        starting from the most significant bit.
        """
        from progs.jubjub import SharedPoint

        product = Ideal(self.curve)
        addend = self
        x_ = await gather(*[e.open() for e in reversed(x)])

        for i_ in x_:
            # Need equality application to compare if i == 1 here
            # Open it to compare for now
            if i_ == 1:
                product += addend
            addend = addend.double()

        return await SharedPoint.from_point(context, product)
```

That said, I'd be curious what @amiller thinks, as it would be putting mpc stuff outside of `progs`. Another thought is that we could add something so that you can easily call this on a `Point`--
```
async def share_mul(self, context, x: list) -> 'Point': 
    from progs.jubjub import share_mul
    return await share_mul(context, self, x)
```

Drake-Eidukas: This can probably wait though, as this would make sense to throw in with a refactor of `mixins`

Drake-Eidukas (changes requested):
Mostly looks good to me-- just some discussion on where to put `share_mul`. Can you do a few small things though--
- [x] amend the commit message to be something more like: `Adds share_mul to jubjub to multiply a Share with a public Point`
- [x] add a citation to the algorithm's docstring
- [x] Use the `equality` method-- it's merged into dev now
- [x] Add the mentioned test case

progs/jubjub.py
```diff
@@ -219,3 +220,29 @@ def __str__(self):
...
+
+async def share_mul(context: Mpc, p: Point, x: list) -> SharedPoint:
+    """
+    The multiplication of the share of a field element and a point
+    e.g. [x]G -> [X], where G is a point on the given elliptic curve
+    x is the bitwise shared value,
+    starting from the most significant bit.
+
+    Source: Section `Scalar multiplication` of
+    https://andrea.corbellini.name/2015/05/17/elliptic-curve-cryptography-a-gentle-introduction/
```
Drake-Eidukas: You'll probably want to add in `# noqa: E501` at the end of this line to pass the linter

progs/jubjub.py
```diff
@@ -219,3 +220,29 @@ def __str__(self):
...
+    """
+    if isinstance(p, Ideal):
+        return SharedIdeal(Jubjub())
```
Drake-Eidukas:
```suggestion
if isinstance(p, Ideal):
  return SharedIdeal(p.curve)

assert isinstance(p, Point):   
```

progs/jubjub.py
```diff
@@ -219,3 +220,29 @@ def __str__(self):
...
+
+    product = Ideal(p.curve)
+    addend = p
+    x.reverse()
+    one_share = context.Share(1)
+
+    for i_ in x:
+        if await (await equality(context, i_, one_share)).open():
```
Drake-Eidukas: We should make `equality` be able to take in an `int` or `GFElement` and have it create a share to compare with. This is fine for now.

tests/progs/test_jubjub.py
```diff
@@ -216,3 +212,28 @@ def test_basic_point_functionality():
...
+
+    p = TEST_POINTS[1]
+    multiplier = 7
+    multiplier_bin = bin(multiplier)[2:]
+
+    async def _prog(context):
+        print("type(SharedIdeal(TEST_CURVE)): ", type(SharedIdeal(TEST_CURVE)))
```
Drake-Eidukas: take this out

Also, you should check out the program [pdb](https://realpython.com/python-debugging-pdb/) for quickly getting this info-- you can invoke it by adding `--pdb` to however you run the tests (e.g. `docker-compose run --rm honeybadgermpc pytest -v --pdb`). You could put a breakpoint somewhere in the code by adding `import pdb; pdb.set_trace()` and then you can evaluate simple expressions like that to quickly test stuff out.

Drake-Eidukas (commented):
Looks good, just have a few comments and this should be about good to go!

tests/progs/test_jubjub.py
```diff
@@ -216,3 +212,28 @@ def test_basic_point_functionality():
...
+        p_ = await SharedPoint.from_point(context, p)
+        multiplier_ = [context.Share(int(i)) for i in multiplier_bin]
+
+        p1_, p2_ = await gather(
+            share_mul(context, p, multiplier_),
+            p_.mul(multiplier))
+        assert await shared_point_equals(p1_, p2_)
+
+        q1_ = await share_mul(context, Ideal(TEST_CURVE), multiplier_)
+        q2_ = SharedIdeal(TEST_CURVE)
```
Drake-Eidukas: oh wait, q2_ should just be an `Ideal`-- `SharedIdeal` is meant for use with `SharedPoint`, which is not used with what you're implementing. I should document this.

progs/jubjub.py
Drake-Eidukas: We can make this more concurrent--
```suggestion
equalities = await asyncio.gather(*[equality(context, i_, one_share) for i_ in x])
opened_equalities = await asyncio.gather(*[eq.open() for eq in equalities])

for eq in opened_equalities:
  if eq:
    product += addend
  
  addend = addend.double()

return await SharedPoint.from_point(context, product)
```

Drake-Eidukas (changes requested):
We should probably do this to make this more concurrent

tests/progs/test_jubjub.py
hanyunx: Nice skill!

Drake-Eidukas (approved):
Looks good to me, requesting review from @amiller

progs/jubjub.py
```diff
@@ -219,3 +220,31 @@ def __str__(self):
...
+    if isinstance(p, Ideal):
+        return SharedIdeal(p.curve)
+
+    product = Ideal(p.curve)
+    addend = p
+    x.reverse()
+    one_share = context.Share(1)
+
+    equalities = await asyncio.gather(*[equality(context, i_, one_share) for i_ in x])
+    opened_equalities = await asyncio.gather(*[eq.open() for eq in equalities])
```
amiller: this is not correct, it is revealing each bit of the secret x, by comparing each bit of x to 1 and then opening the equality share. Think of another way to achieve this without using an `if` statement!

amiller (changes requested):
needs a different algorithm to keep the key secret

tests/progs/test_jubjub.py
```diff
@@ -216,3 +212,26 @@ def test_basic_point_functionality():
             await p1.montgomery_mul(5))
 
     await run_test_prog(_prog, test_preprocessing)
+
+
+@mark.asyncio
+async def test_share_mul(test_preprocessing):
+
+    p = TEST_POINTS[1]
+    multiplier = 7
```
amiller: this is not a good test... if the share to multiply by is represented by a sequence of secret shared bits, which can potentially take the full range of Fp, approx ~255 bits, then you shouldn't test it just by passing in 3 bits rather than 255

progs/jubjub.py
```diff
@@ -219,3 +219,41 @@ def __str__(self):
...
+          = [b_i] * (P2i.y - 1) + 1
+    So we get the SharedPoint of each term.
+    """
+    if isinstance(p, Ideal):
+        return SharedIdeal(p.curve)
+
+    terms = []
+    for i in range(len(bs)):
+        p2i = (2**i) * p
+        term = await SharedPoint.create(context,
```
amiller: Why do we have this `.create` method instead of just the ordinary constructor for `SharedPoint`?  Can we just use the constructor?

amiller (changes requested):
We should make a couple of changes to avoid unnecessary computation. The `__on_curve` method currently written is wrong, and in any case it usually should not be called. For now, can you modify the code so that `__add__`, `double`, and `share_mul` all only use the constructor, not `create`? This will bypass the unnecessary `__on_curve` check.

amiller (changes requested):
We should make a couple of changes to avoid unnecessary computation. The `__on_curve` method currently written is wrong, and in any case it usually should not be called. For now, can you modify the code so that `__add__`, `double`, and `share_mul` all only use the constructor, not `create`? This will bypass the unnecessary `__on_curve` check.

progs/jubjub.py
Drake-Eidukas: Currently, it's this way because the existing code I took jubjub from had the `on_curve` check for created points, and you can't use `await` from `__init__` for computing `on_curve`. I think it's reasonable to remove this check for code that should not produce points off the curve.

amiller (approved):

amiller:
Thanks! I think this looks great

honeybadgermpc/progs/jubjub.py
```diff
@@ -74,10 +74,10 @@ def __repr__(self) -> str:
         return str(self)
 
     async def neg(self):
```
Drake-Eidukas: Since nothing about this is asynchronous anymore, let's remove `async` from this method definition

Drake-Eidukas (approved):
LGTM, as soon as you rebase and the checks pass let's get this merged in

amiller:
I'm stuck and don't know why the tests here are not finishing quickly. 
In the .ci log, it gets stuck at 
```[gw1] [ 98%] PASSED tests/progs/mixins/test_utils.py::test_static_type_check_overrun_unnamed 
```
When I run the entire test suite locally, I actually see it get stuck somewhere at `test_sharedpoint_sub`.
But when I just run `pytest -v tests/progs/test_jubjub.py` all of these new tests pass. So I'm not sure where the trouble is

amiller:
Ok, with setting `n=4` how it originally was, this PR passes. To me this indicates a bug somewhere that is masked by setting n=4. But I don't think it's a problem with this PR. So what we should do is merge this to unblock hanyun but investigatee with a separate issue

PR #251, opened 2019-04-17
HBAVSS: implement implicate handling and share recovery

nqyy:
The implementation of implicate handling part and share recovery part is complete. They are tested separately in the code. The whole structure of the code is changed and it doesn't return a value but set a future value for output instead, as you can see in the code change of the test.

A new method of interpolation on G1 is added to betterpairing.

New tests are also added.

### DIFF
--- a/honeybadgermpc/betterpairing.py
+++ b/honeybadgermpc/betterpairing.py
@@ -668,6 +668,9 @@ class ZR:
                 'Invalid division param. Expected ZR or int. Got '
                 + str(type(other)))
 
+    def __rtruediv__(self, other):
+        return ZR(other).__truediv__(self)
+
     def __pow__(self, other):
         if type(other) is int:
             exponend = ZR(other % (bls12_381_r-1))
@@ -717,3 +720,30 @@ class ZR:
     @staticmethod
     def one():
         return ZR(PyFr("1"))
+
+
+def lagrange_at_x(s, j, x):
+    s = sorted(s)
+    assert j in s
+    l1 = [x - jj for jj in s if jj != j]
+    l2 = [j - jj for jj in s if jj != j]
+    (num, den) = (ZR(1), ZR(1))
+    for item in l1:
+        num *= item
+    for item in l2:
+        den *= item
+    return num / den
+
+
+def interpolate_g1_at_x(coords, x, order=-1):
+    if order == -1:
+        order = len(coords)
+    xs = []
+    sortedcoords = sorted(coords, key=lambda x: x[0])
+    for coord in sortedcoords:
+        xs.append(coord[0])
+    s = set(xs[0:order])
+    out = G1.one()
+    for i in range(order):
+        out = out * (sortedcoords[i][1] ** (lagrange_at_x(s, xs[i], x)))
+    return out
--- a/honeybadgermpc/hbavss.py
+++ b/honeybadgermpc/hbavss.py
@@ -1,12 +1,11 @@
 import logging
 import asyncio
 from pickle import dumps, loads
-from honeybadgermpc.betterpairing import ZR
+from honeybadgermpc.betterpairing import ZR, interpolate_g1_at_x
 from honeybadgermpc.polynomial import polynomials_over
 from honeybadgermpc.poly_commit_const import PolyCommitConst
 from honeybadgermpc.poly_commit_lin import PolyCommitLin
 from honeybadgermpc.symmetric_crypto import SymmetricCrypto
-from honeybadgermpc.exceptions import HoneyBadgerMPCError
 from honeybadgermpc.protocols.reliablebroadcast import reliablebroadcast
 from honeybadgermpc.protocols.avid import AVID
 from honeybadgermpc.utils import wrap_send
@@ -23,8 +22,11 @@ logger.setLevel(logging.ERROR)
 
 class HbAVSSMessageType:
     OK = "OK"
-    # IMPLICATE = "IMPLICATE"
+    IMPLICATE = "IMPLICATE"
     READY = "READY"
+    RECOVERY = "RECOVERY"
+    RECOVERY1 = "RECOVERY1"
+    RECOVERY2 = "RECOVERY2"
 
 
 class HbAvssLight():
@@ -55,6 +57,24 @@ class HbAvssLight():
     def __exit__(self, type, value, traceback):
         self.subscribe_recv_task.cancel()
 
+    def _handle_implication(self, commitment, ephemeral_public_key, j, j_sk, j_z):
+        """
+        Handle the implication of AVSS.
+        Return True if the implication is valid, False otherwise.
+        """
+        # discard if PKj ! = g^SKj
+        if self.public_keys[j] != pow(self.g, j_sk):
+            return False
+        # decrypt and verify
+        j_shared_key = pow(ephemeral_public_key, j_sk)
+        try:
+            j_share, j_aux = SymmetricCrypto.decrypt(
+                str(j_shared_key).encode(), j_z)
+        except Exception:  # TODO: specific exception
+            return True
+        return not self.poly_commit.verify_eval(
+                commitment, j+1, j_share, j_aux)
+
     async def _process_avss_msg(self, avss_id, dealer_id, avss_msg):
         tag = f"{dealer_id}-{avss_id}-AVSS"
         send, recv = self.get_send(tag), self.subscribe_recv(tag)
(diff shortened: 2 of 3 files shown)


honeybadgermpc/hbavss.py
```diff
@@ -323,28 +270,156 @@ def _get_dealer_msg(self, values):
...
+                    for j in range(self.n):
+                        phi_i_j = polynomials_over(ZR).interpolate_at(
+                            zip(range(0, secret_size), shares), x_recomb=j)
+                        w_i_j = polynomials_over(ZR).interpolate_at(
+                            zip(range(0, secret_size), witnesses), x_recomb=j)
+                        send(j, (HbAVSSMessageType.RECOVERY1, phi_i_j, w_i_j))
+            # R1
+            if avss_msg[0] == HbAVSSMessageType.RECOVERY1:
+                _, phi_k_i, w_k_i = avss_msg
+                # Will be uncommented once the interpolation of commitments is done
```
nqyy: @tyurek said he'll implement the interpolation on the commitments. So, I commented this part out for now.

tyurek:
I'm thinking we wait until we have implicate fully working before trying to merge this

amiller:
@tyurek said he will prefer to review this after the new polycommit is integrated

nqyy:
Will update this once https://github.com/initc3/HoneyBadgerMPC/pull/257 is merged.

nqyy:
Fully updated.

amiller:
Tom and I discussed and realized that something this is missing from this for our use case is using a discrete log proofs about the encryption key, rather than the key itself. We know what the next step is for that, but I think it can go as a separate PR.

honeybadgermpc/betterpairing.py
```diff
@@ -717,3 +717,45 @@ def zero():
     @staticmethod
     def one():
         return ZR(PyFr("1"))
+
```
smkuls, line 723: I am not okay with this change. We have a whole module dedicated to polynomial operations - `polynomial.py`. We should try to make use of that instead of duplicating functionality just because we are operating on a different type.

honeybadgermpc/hbavss.py
```diff
@@ -229,57 +232,144 @@ def __exit__(self, typ, value, traceback):
         for task in self.tasks:
             task.cancel()
 
-    async def _process_avss_msg(self, avss_id, dealer_id, dispersal_msg):
+    async def _process_avss_msg(self, avss_id, dealer_id, rbc_msg, avid):
```
smkuls, line 308: Consider splitting this method, this is an extremely long method.

honeybadgermpc/hbavss.py
```diff
@@ -229,57 +232,144 @@ def __exit__(self, typ, value, traceback):
...
-        # if 2t+1 OK or t+1 READY -> send all ready
-        while len(ok_set) < 2 * self.t + 1 and len(ready_set) < self.t + 1:
-            sender, avss_msg = await recv()  # First value is the `sid`
-            if avss_msg == HbAVSSMessageType.OK and sender not in ok_set:
-                ok_set.add(sender)
-            elif avss_msg == HbAVSSMessageType.READY and sender not in ready_set:
-                ready_set.add(sender)
-
-        multicast(HbAVSSMessageType.READY)
+        r1_set = set()
```
smkuls, line 352: what's `r1_set`? I am not aware of the protocol but can you pick a more descriptive name?

honeybadgermpc/hbavss.py
```diff
@@ -229,57 +232,144 @@ def __exit__(self, typ, value, traceback):
...
+        r2_set = set()
+        ready_sent = False
```
smkuls: nit:
`read_sent = r1_sent = r2_sent = False`

honeybadgermpc/hbavss.py
```diff
@@ -229,57 +232,144 @@ def __exit__(self, typ, value, traceback):
...
-        return shares
+            # IMPLICATE
+            if avss_msg[0] == HbAVSSMessageType.IMPLICATE:
+                # Handling Implication
+                j = sender
+                j_pk = avss_msg[1]
+                j_k = avss_msg[2]
+                # discard if PKj ! = g^SKj
+                if self.public_keys[j] != pow(self.g, j_pk):
+                    logger.debug("[%d] invalid implicate from sender %s",
```
smkuls: `logger.warn`

honeybadgermpc/hbavss.py
```diff
@@ -229,57 +232,144 @@ def __exit__(self, typ, value, traceback):
...
+                                 self.my_id, sender)
+                    continue
+                # decrypt and verify
+                implicate_msg = await avid.retrieve(tag, j)
+                j_z = loads(implicate_msg)
+                j_shared_key = pow(ephemeral_public_key, j_pk)
+                j_share, j_aux, j_witnesses = SymmetricCrypto.decrypt(
+                    str(j_shared_key).encode("utf-8"), j_z[j_k])
```
smkuls: calling `encode()` is faster than calling `encode("utf-8")`

honeybadgermpc/hbavss.py
```diff
@@ -229,57 +232,144 @@ def __exit__(self, typ, value, traceback):
...
+                aux_coords = [None] * secret_size
+                w_coords = [None] * secret_size
+                for i in range(secret_size):
+                    c_coords[i] = (i, commitments[i])
+                    phi_coords[i] = (i, shares[i])
+                    aux_coords[i] = (i, auxes[i])
+                    w_coords[i] = (i, witnesses[i])
+                # interpolate commitments
+                interpolated_c = [None] * self.n
+                for i in range(self.n):
```
smkuls, line 388: This should not be using lagrange - we should switch to Vandermonde/FFT. I think we got rid of lagrange from most places in our code.

honeybadgermpc/hbavss.py
```diff
@@ -229,57 +232,144 @@ def __exit__(self, typ, value, traceback):
...
+                    interpolated_c[i] = interpolate_g1_at_x(c_coords, i)
+                # if my shares are valid
+                if not r1_sent and self.shares_future.done():
+                    r1_sent = True
+                    for j in range(self.n):
+                        phi_i_j = interpolate_at_x(phi_coords, j)
```
smkuls: same here, this looks like you are interpolating at all `n` points - consider using Vandermonde/FFT

tests/test_hbavss.py
```diff
@@ -54,18 +54,24 @@ def get_avss_params(n, t):
...
             hbavss = HbAvssBatch(pks, sks[i], crs, n, t, i, sends[i], recvs[i])
+            hbavss_list.append(hbavss)
             stack.enter_context(hbavss)
             if i == dealer_id:
                 avss_tasks[i] = asyncio.create_task(hbavss.avss(0, values=values))
             else:
                 avss_tasks[i] = asyncio.create_task(hbavss.avss(0, dealer_id=dealer_id))
-        shares = await asyncio.gather(*avss_tasks)
+        for i in range(n):
+            shares[i] = await hbavss_list[i].shares_future
```
smkuls: use `asyncio.gather` this serializes the code.

tests/test_hbavss.py
```diff
@@ -54,18 +54,24 @@ def get_avss_params(n, t):
     crs = gen_pc_const_crs(t, g=g, h=h)
 
     values = [ZR.random()] * (t+1)
-    avss_tasks = [None]*n
+    avss_tasks = [None] * n
     dealer_id = randint(0, n-1)
 
+    shares = [None] * n
     with ExitStack() as stack:
+        hbavss_list = []
```
smkuls: nit:
`hbavss_list = [None]*n`

honeybadgermpc/hbavss.py
```diff
@@ -334,28 +424,30 @@ def _get_dealer_msg(self, values):
...
+            # dispersal_msg_list: the list of payload z
+            broadcast_msg, dispersal_msg_list = self._get_dealer_msg(values)
 
         # In the client_mode, the dealer is the last node
         n = self.n if not client_mode else self.n+1
 
         tag = f"{dealer_id}-{avss_id}-B-AVID"
         send, recv = self.get_send(tag), self.subscribe_recv(tag)
 
+        rbc_msg = await reliablebroadcast(
```
smkuls, line 512: add a log statement before and after, this helps in debugging what got stuck where

honeybadgermpc/hbavss.py
```diff
@@ -308,13 +398,13 @@ def _get_dealer_msg(self, values):
                 z[k] = SymmetricCrypto.encrypt(
                     str(shared_key).encode("utf-8"),
                     (phi[k](i+1), aux_poly[k](i+1), witness))
-            dealer_msg_list[i] = dumps((commitments, ephemeral_public_key, z))
+            dispersal_msg_list[i] = dumps(z)
 
-        return dealer_msg_list
+        return dumps((commitments, ephemeral_public_key)), dispersal_msg_list
```
smkuls, line 474: Can we avoid a double `dumps` here? Can you not pickle them just here and get rid of the previous pickle?

honeybadgermpc/betterpairing.py
smkuls: That module actually caches interpolations and here you don't get that benefit. I would strongly suggest using that code since that contains similar operations in one file making the code more maintainable in the future.

smkuls (changes requested):

honeybadgermpc/betterpairing.py
nqyy: @tyurek

honeybadgermpc/hbavss.py
nqyy: That message is recovery phase 1 and it is called exactly as R1 in the paper

smkuls:
@chitianhao I don't see any tests as well. You should cover the entire code - this includes the implicate and recovery phase. We never got to adding proper tests for your last PR so I'd want to have them checked in along with this.

honeybadgermpc/hbavss.py
nqyy: Yeah, but we also encode it this way in the light version.

smkuls: You can change it both places. :)

tests/test_hbavss.py
nqyy: The batch avss doesn't return anything now. It output the value in a Future variable.

honeybadgermpc/hbavss.py
nqyy: No, they are separate and one is for RBC and the other one (need to be a list) for AVID.

nqyy:
Yeah, I think we should add test for that. @tyurek is trying to find a way to test this part properly.

honeybadgermpc/hbavss.py
nqyy: Yep. I'll see what I can do.

honeybadgermpc/hbavss.py
nqyy: @tyurek

tests/test_hbavss.py
smkuls: You can collect all futures and then do a gather on them:
```
shares = await asyncio.gather(*[hbavss_list[i].shares_future for i in range(n)])
```

nqyy: Yeah, this is a good way.

honeybadgermpc/hbavss.py
tyurek: I don't have code handy to do FFT interpolation for commitments in G1. If someone wants to tackle it, feel free, but I don't think it's terribly important as this is just fault handling anyway. It would be new work with little benefit

honeybadgermpc/hbavss.py
tyurek: As this is ZR, FFT should be used, yeah

tyurek (changes requested):
Implicate handling for hbavss-lite would be a good thing to include. Needs tests as well

honeybadgermpc/hbavss.py
tyurek: interpolation in a finite field should use the polynomial class and FFT and all that though

nqyy:
Thanks to @tyurek for adding the test. And I also fixed an issue with it,

tyurek (approved):

honeybadgermpc/hbavss.py
```diff
@@ -62,44 +84,78 @@ def multicast(msg):
...
-            if avss_msg == HbAVSSMessageType.OK and sender not in ok_set:
+        recovery_shares = []
+        sent_recovery = False
+        output = False
+        while True:
+            if len(ok_set) == 2 * self.t + 1 and share_valid and not output:
+                share_int = int(share)
+                self.output_queue.put_nowait((dealer_id, avss_id, share_int))
+                output = True
+            if len(recovery_shares) == self.t + 1 and not output:
```
Drake-Eidukas: Is there ever a case where both of these `if` statements are supposed to be true? If not, we should make this second `if` and `elif` to make this clearer.

honeybadgermpc/hbavss.py
```diff
@@ -229,57 +285,166 @@ def __exit__(self, typ, value, traceback):
...
+        # decrypt and verify
+        implicate_msg = await avid.retrieve(tag, j)
+        j_z = loads(implicate_msg)
+        j_shared_key = pow(ephemeral_public_key, j_pk)
+        try:
+            j_share, j_aux, j_witnesses = SymmetricCrypto.decrypt(
+                str(j_shared_key).encode(), j_z[j_k])
+        except Exception:
+            return True
+        if self.poly_commit.verify_eval(
```
Drake-Eidukas:
```suggestion
return not self.poly_commit.verify_eval(
    commitments[j_k], j+1, j_share, j_aux, j_witnesses)
```

honeybadgermpc/hbavss.py
```diff
@@ -54,6 +56,26 @@ def __enter__(self):
...
+        if self.public_keys[j] != pow(self.g, j_sk):
+            return False
+        # decrypt and verify
+        j_shared_key = pow(ephemeral_public_key, j_sk)
+        try:
+            j_share, j_aux = SymmetricCrypto.decrypt(
+                str(j_shared_key).encode(), j_z)
+        except Exception:  # TODO: specific exception
+            return True
+        if self.poly_commit.verify_eval(
```
Drake-Eidukas: Same as other comment

tests/test_hbavss.py
```diff
@@ -54,18 +194,268 @@ def get_avss_params(n, t):
...
+async def test_hbavss_batch_t_share_faults(test_router):
+    def callback(future):
+        if future.done():
+            ex = future.exception()
+            if ex is not None:
+                print('\nException:', ex)
+                raise ex
+
+    class BadDealer(HbAvssBatch):
+        def _get_dealer_msg(self, values):
```
Drake-Eidukas, line 318: This looks copy-pasted from the other tests-- if so, could you move it into `tests/fixtures`? It'll be better for future maintenance if this is just defined once.

Drake-Eidukas (changes requested):
Mostly style nits. Could you also squash the commits and make a unified commit message?

tests/test_hbavss.py
tyurek: it's slightly different each time. The tests have the dealer introduce faults in differing ways

tyurek: repeatedly overriding the same class function was just the most convenient way to do that

Drake-Eidukas: Fair enough-- they looked _mostly_ the same. It might be a good idea to leave a comment describing how they introduce faults to help anyone that has to work on this in the future then.

tyurek: unless you meant the callback?

honeybadgermpc/betterpairing.py
```diff
@@ -717,3 +720,30 @@ def zero():
...
+        den *= item
+    return num / den
+
+
+def interpolate_g1_at_x(coords, x, order=-1):
+    if order == -1:
+        order = len(coords)
+    xs = []
+    sortedcoords = sorted(coords, key=lambda x: x[0])
+    for coord in sortedcoords:
```
Drake-Eidukas, line 743: This will be faster if you do--
```
sorted_coords = sorted(coords, key=lambda x: x[0])
xs = [coord[0] for coord in sorted_coords]
```

tests/test_hbavss.py
```diff
@@ -29,18 +31,158 @@ def get_avss_params(n, t):
...
+    g, h, pks, sks = get_avss_params(n, t)
+    sends, recvs, _ = test_router(n)
+    crs = [g, h]
+
+    value = ZR.random()
+    avss_tasks = [None]*n
+    hbavss_list = [None]*n
+    dealer_id = randint(0, n-1)
+
+    with ExitStack() as stack:
```
Drake-Eidukas, line 99: This part with ExitStack() is definitely duplicated across tests.

Drake-Eidukas (approved):
I would like it if you can pull out the stuff using `with ExitStack() as stack` into a method that you call rather than have it duplicated, but other than that, this looks good to me. I'll wait on a final review from @smkuls

tests/test_hbavss.py
smkuls: Lets take this later. This PR is too blown up. Lets not take up any more changes in this one. This is blocking other people as well.

honeybadgermpc/betterpairing.py
smkuls: @tyurek are these methods still needed?

smkuls: I thought these were moved these to the polynomial class?

smkuls (commented):
As of this moment there's way too much code for this to be reviewed properly.

From next time, lets split it in multiple PRs. As a feedback, you can always reply to a review suggestion that you want to take something later and immediately follow up with another PR. 

I am approving this.

smkuls (approved):

nqyy:
Rebased. Let's ship it.

PR #254, opened 2019-04-18
Implementation of JubJub+MiMC Public Key encryption

hanyunx:
Implementation of `mimc_encrypt` and `mimc_decrypt`.
The running time of the test is very long ...


progs/symmetrc_mimc.py
```diff
@@ -0,0 +1,69 @@
...
+CURVE = Jubjub()
+KEY_LENGTH = 32
+# GP: The generator of Jubjub curve, hardcode here
+GP = Point(5,
+           6846412461894745224441235558443359243034138132682534265960483512729196124138,
+           CURVE)
+
+
+async def key_generation(context):
+    """
```
Drake-Eidukas: Great documentation here

progs/symmetrc_mimc.py
```diff
@@ -0,0 +1,69 @@
...
+    The MPC system creates random bitwise shared value [x]_B
+    as the private key (priv_key),
+    then calcultes X = ([x]G).open as the public key (pub_key)
+    """
+    # Generate the private key
+    priv_key = [pp_elements.get_bit(context) for _ in range(KEY_LENGTH)]
+    # priv_key_p = field(0)
```
Drake-Eidukas: Make sure to clean up commented code, or provide a reason why it's commented out

progs/symmetrc_mimc.py
```diff
@@ -0,0 +1,69 @@
...
+from honeybadgermpc.elliptic_curve import Point, Jubjub
+from progs.jubjub import share_mul
+from progs.mimc import mimc_mpc, mimc_plain
+
+pp_elements = PreProcessedElements()
+field = GF(Subgroup.BLS12_381)
+CURVE = Jubjub()
+KEY_LENGTH = 32
+# GP: The generator of Jubjub curve, hardcode here
+GP = Point(5,
```
Drake-Eidukas: This probably belongs in the Jubjub class

progs/symmetrc_mimc.py
```diff
@@ -0,0 +1,69 @@
...
+from honeybadgermpc.field import GF
+from honeybadgermpc.mpc import PreProcessedElements, Subgroup
+from honeybadgermpc.elliptic_curve import Point, Jubjub
+from progs.jubjub import share_mul
+from progs.mimc import mimc_mpc, mimc_plain
+
+pp_elements = PreProcessedElements()
+field = GF(Subgroup.BLS12_381)
+CURVE = Jubjub()
+KEY_LENGTH = 32
```
Drake-Eidukas: This should be a method parameter on encryption and decryption. I think having this be a default named argument is sufficient for now, but otherwise, see prof. miller's previous comment about this on your last PR

Drake-Eidukas (changes requested):
So, other than the things I mentioned in comments-- you'll want to do a few things here:
- [x] Squash the commits for _this PR_ into 1 commit, and have the commits from other branches (e.g. the `share_mul` function) come before that one commit. This will help us review just the changes that should come from this PR (the alternative is having your branches be on this fork and having the base of the PR point to the branch that needs to be merged first)
- [ ] Add benchmarking tests for MiMC (something I need to do)

progs/symmetrc_mimc.py
amiller: No sarcasm in code review please :o

Drake-Eidukas: It was a compliment not sarcasm-- apologies if it came off otherwise

amiller:
I think this should be refactored a bit. The file `symmetric_mimc.py` should not include any dependency on jubjub. 
The `key_generation` function defined here is appropriate for public key encryption.

progs/symmetrc_mimc.py
hanyunx: There was neat documentation here but was overwritten by my messed rebase afterwards -- apologies for the misunderstanding I brought lol

honeybadgermpc/progs/mimc_jubjub_pkc.py
```diff
@@ -0,0 +1,77 @@
...
+from honeybadgermpc.elliptic_curve import Point, Jubjub
+from honeybadgermpc.progs.jubjub import share_mul
+from honeybadgermpc.progs.mimc import mimc_mpc, mimc_plain
+
+pp_elements = PreProcessedElements()
+field = GF(Subgroup.BLS12_381)
+CURVE = Jubjub()
+KEY_LENGTH = 32
+# GP: The generator of Jubjub curve, hardcode here
+GP = Point(5,
```
Drake-Eidukas, line 10: This should go in the jubjub class as a constant

honeybadgermpc/progs/mimc_jubjub_pkc.py
```diff
@@ -0,0 +1,77 @@
...
+from honeybadgermpc.field import GF
+from honeybadgermpc.mpc import PreProcessedElements, Subgroup
+from honeybadgermpc.elliptic_curve import Point, Jubjub
+from honeybadgermpc.progs.jubjub import share_mul
+from honeybadgermpc.progs.mimc import mimc_mpc, mimc_plain
+
+pp_elements = PreProcessedElements()
+field = GF(Subgroup.BLS12_381)
+CURVE = Jubjub()
+KEY_LENGTH = 32
```
Drake-Eidukas: this should be a default parameter of the `key_generation` method

Drake-Eidukas: I think `CURVE` as well could fit the bill for moving it into a default argument.

honeybadgermpc/progs/mimc_jubjub_pkc.py
```diff
@@ -0,0 +1,77 @@
+import asyncio
+from random import randint
+from honeybadgermpc.field import GF
+from honeybadgermpc.mpc import PreProcessedElements, Subgroup
+from honeybadgermpc.elliptic_curve import Point, Jubjub
+from honeybadgermpc.progs.jubjub import share_mul
+from honeybadgermpc.progs.mimc import mimc_mpc, mimc_plain
+
+pp_elements = PreProcessedElements()
+field = GF(Subgroup.BLS12_381)
```
Drake-Eidukas: use `context.field` instead

honeybadgermpc/progs/mimc_jubjub_pkc.py
```diff
@@ -0,0 +1,77 @@
+import asyncio
+from random import randint
+from honeybadgermpc.field import GF
+from honeybadgermpc.mpc import PreProcessedElements, Subgroup
+from honeybadgermpc.elliptic_curve import Point, Jubjub
+from honeybadgermpc.progs.jubjub import share_mul
+from honeybadgermpc.progs.mimc import mimc_mpc, mimc_plain
+
+pp_elements = PreProcessedElements()
```
Drake-Eidukas: make this a local variable of the `key_generation` method

honeybadgermpc/progs/mimc_jubjub_pkc.py
```diff
@@ -0,0 +1,77 @@
...
+    then sents (ciphertext, A) to the MPC system.
+
+    a - randomly generated, only kept by the dealer
+    a_ - auxiliary variable to be sent to MPC system for decryption
+    k - x coordinate of a * pub_key, secret key only kept by the dealer
+    ms - blocks of plaintext, that is, plaintext -> {m1, m2,...,ml}
+    i - The index of the plaintext array, take it as counter
+    ciphertext - a field element
+    """
+    if seed:
```
Drake-Eidukas: small nit--
```suggestion
a = randint(0, field.modulus) if seed is None else seed
```

honeybadgermpc/progs/mimc_jubjub_pkc.py
```diff
@@ -0,0 +1,77 @@
...
+    pub_key = Point(x, y, CURVE)
+    return priv_key, pub_key
+
+
+def mimc_encrypt(pub_key, ms, seed=None):
+    """
+    The dealer does the counter-mode encryption to plaintext,
+    then sents (ciphertext, A) to the MPC system.
+
+    a - randomly generated, only kept by the dealer
```
Drake-Eidukas: You have fantastic method documentation, but with one small nit-pick-- i think it would be better to document these things in-line in the code for readability. For example--
```
# Randomly generated variable used only by the dealer
a = randint(0, field.modulus) if seed is None else seed

# Auxiliary variable needed for decryption
a_ = a * GP

# secret key kept by dealer
k = (a * pub_key).x

# ciphertext sent to user
ciphertext = [mimc_plain(i, k) + m for (idx, m) in enumerate(ms)]

return ciphertext, a_
```

Drake-Eidukas: Then, in your actual docstring for the method, using the google style formatting is super helpful for readability:
```
def mimc_encrypt(pub_key, ms, seed=None):
""" Encrypts blocks of plaintext data using counter-mode encryption.

args: 
  pub_key (Point): public key etc.
  ms (list): list of plaintext data to encode
  seed(int): seed to use for random generation in encryption

output: 
  ciphertext (list): encoded blocks
  a_ (Point) auxilliary point sent to caller for decryption
"""
```

honeybadgermpc/progs/mimc_jubjub_pkc.py
```diff
@@ -0,0 +1,77 @@
...
+    """
+    The MPC system decrypts the ciphertext to get the shared value of plaintext.
+
+    ([x]A).x -> [k], where k = S.x and S = aX = axG := xA
+    cs - blocks of ciphertext, {c1, c2,...,cl}
+    """
+    (cs, a_) = ciphertext
+    k_share = (await share_mul(context, priv_key, a_)).xs
+    decrypted = []
+
```
Drake-Eidukas, line 75:
```suggestion 
mpcs = asyncio.gather(*[mimc_mpc(context, context.field(i), k_share) for i in range(len(cs))])
decrypted = [c - m for (c, m) in zip(cs, mpcs)]

return decrypted
```

This will proceed more efficiently.

honeybadgermpc/progs/mimc_symmetric.py
```diff
@@ -0,0 +1,40 @@
+from honeybadgermpc.field import GF
+from honeybadgermpc.mpc import PreProcessedElements, Subgroup
+from honeybadgermpc.progs.mimc import mimc_mpc, mimc_plain
+
+pp_elements = PreProcessedElements()
+field = GF(Subgroup.BLS12_381)
+
+
+# def get_key():
```
Drake-Eidukas: can you either remove this or comment why it's commented out?

honeybadgermpc/progs/mimc_symmetric.py
```diff
@@ -0,0 +1,40 @@
...
+    """
+    ms - blocks of plaintext, that is, plaintext -> {m1, m2,...,ml}
+         Each plaintext is a field element.
+    ciphertext <- F_MiMC(counter, key) + plaintext
+    """
+    ciphertext = []
+    for i in range(len(ms)):
+        ciphertext.append(mimc_plain(i, key) + ms[i])
+
+    return ciphertext
```
Drake-Eidukas: `return [mimc_plain(idk, key) + m for (idx, m) in enumerate(ms)]`

honeybadgermpc/progs/mimc_jubjub_pkc.py
Drake-Eidukas: Also, the second half of this method is the exact same as `mimc_decrypt` from `mimc_symmetric`-- it would be great if we could write the code once and reuse it across both methods.

honeybadgermpc/progs/mimc_jubjub_pkc.py
```diff
@@ -0,0 +1,77 @@
...
+    if seed:
+        a = seed
+    else:
+        a = randint(0, field.modulus)
+
+    a_ = a * GP
+    k = (a * pub_key).x
+    ciphertext = []
+
+    for i in range(len(ms)):
```
Drake-Eidukas: 1. `ciphertext = [mimc_plain(idx, k) + m for (idx, m) in enumerate(ms)]`
2. This is duplicated from `mimc_symmetric`

honeybadgermpc/progs/jubjub.py
```diff
@@ -232,15 +232,13 @@ def __str__(self):
...
-        terms.append(term)
+        x = p2i.x * bs[i]
+        y = (p2i.y - 1) * bs[i] + p.curve.Field(1)
+        terms.append(SharedPoint(context, x, y, p.curve))
         p2i = p2i.double()
 
     accum = terms[0]
-    for i in terms[1:]:
-        accum = await accum.add(i)
+    for term in terms[1:]:
```
Drake-Eidukas: This will proceed sequentially, but we can do it in `log(n)` rounds. we can use the same trick we used in `Equality`--
```
while len(terms) > 1:
    left_terms, right_terms = terms[::2], terms[1::2]
    terms = await asyncio.gather(*[l.add(r) for (l, r) in zip(left_terms, right_terms)])

return terms[0]
```

Drake-Eidukas (changes requested):
Mostly style and cleanup stuff, some performance improvements (hopefully). PR should be good to go once these changes are made.

honeybadgermpc/progs/mimc_jubjub_pkc.py
Drake-Eidukas: hmm, I see why you need this though since the one method doesn't take in a context... but taking in a field argument would probably be better in that method. This could cause problems down the road otherwise if we add in more fields.

honeybadgermpc/progs/mimc_jubjub_pkc.py
Drake-Eidukas: actually, that could potentially make this have way worse performance as it would have to read the preprocessing file each time you run the method -- keep this variable like this.

honeybadgermpc/progs/mimc_jubjub_pkc.py
hanyunx: I removed this line. Now in `mimc_encrypt` method I use `Jubjub.Field` instead, though the problem when more fields added still exists.

honeybadgermpc/progs/jubjub.py
hanyunx: This may overlook the last element if  the number of elements is odd, so padding ZERO to `terms` at the start of each loop is to be added.

honeybadgermpc/progs/mimc_jubjub_pkc.py
hanyunx: I think it's ok to leave the duplicate part here. I prefer to take them as two separate files without calling one from the other.

honeybadgermpc/progs/mimc_jubjub_pkc.py
hanyunx: Neat!

honeybadgermpc/progs/mimc_jubjub_pkc.py
hanyunx: left it as TODO

honeybadgermpc/progs/jubjub.py
Drake-Eidukas: Actually, there's no need to pad with 0-- if `left_terms` has more elements than `right_terms`, you just need to append `left_terms[-1]` to terms.

hanyunx: oh right.

Drake-Eidukas (approved):
Looks good-- just squash and we can merge this in.

PR #256, opened 2019-04-19
removes leftover ideal functionality stuff

amiller:
We can revisit using ideal functionality mockups for modularity at a later time :o


Drake-Eidukas (approved):
Great!

PR #257, opened 2019-04-19
HBAVSS: Use const polynomial commitment

nqyy:
Use the new PolyCommitConst.

The light class and the batch class are totally separate classes now because they are totally different and even more different with the new changes in share recovery.


Drake-Eidukas:
You'll have to rebase before this can be merged btw

amiller:
@tyurek needs to review this together with #251

nqyy:
I will update that one after this is merged I think.

tyurek (approved):
Tweaked some interfaces. Looks good to me assuming Travis is happy. Approved.

chitianhao:
I think I should squash

Issue #253, opened 2019-04-18
Refactor Mixin code


Drake-Eidukas:
Current design: 
- Create abstract base `MixinOperation` class that requires the overriding of a method to get the name of the operation, as well as  the `__call__()` operator for running the mixin operation.
  - Consider listing preprocessing requirements
- Refactor current mixins to use this `MixinOperation`
- Move mixin code largely to `progs/`
- Look into easier ways of loading mixins into our mpc application
  - One idea is to have a `TestProgramRunner` which handles this
  - Another is to have `MixinManager` class which delegates the work behind calling and verifying the existence of mixins

I'll be trying out some potential approaches to see what works best.

Other potential things to do--
- move equality to be a mixin
- implement point multiplying shared value as a mixin instead of a plain prog
- Consider making subclasses for unary and binary operations with convenient type checking, and another for more generic mixins
- Look into most convenient ways to integrate these mixins with our types that don't require mixin code encroaching other code

Issue #223, opened 2019-03-29
Add type hints to codebase


Drake-Eidukas:
Currently, more or less the entire codebase is written without [type hints](https://docs.python.org/3/library/typing.html), and this can cause some pain points especially with methods that require a specific type to correctly execute. It would be great if we could integrate these hints in our codebase.

- Users of our code who choose to also use type hints / `mypy` would have more ease of use
- We can use this while coding / debugging, as well as benefit from extra documentation.

amiller:
Type hints are now used fairly extensively with #259

PR #259, opened 2019-04-21
[Issue 253] Refactor Mixins and Implement Type-Checker

Drake-Eidukas:
Tackles #253 
As mentioned in the commits/headline, this PR really does 3 things

- Breaks out relevant `mixin` code into `progs/mixins`. This includes equality, and division, which previously weren't mixins. It's getting somewhat annoying to include all of the necessary mixins though, so it would be interesting to refactor `TaskProgramRunner` to more intelligently include mixins (i.e., since `MixinBase` objects have a `NAME` field, there's no need for the current way of loading in mixins being a dictionary). I Implemented a `test_program_runner` fixture which eases this process. 
- Refactors `Share`, `ShareArray` (barely), `ShareFuture`, and `GFElementFuture` to be abstract base classes defined outside of a method. These facilitate much easier introspection / typechecking, and makes `Mpc` create concrete implementations with calls to `type()` to create the `context` class property.
- Implements a type checker. Better docs are `# TODO`, but the Tl;DR is that code that today looks like
```
def func(self, context, a, b, c=context.Share(3)): 
  if not isintsance(context, Mpc):
     raise TypeError
  elif not isinstance(a, (type_a, type_b)): 
    raise TypeError
  elif not isinstance(a, type_c):
    raise TypeError
  elif not isinstance(c, context.Share):
    raise TypeError

  # rest of implementation...
```
will tomorrow look like:
```
# Using strings for types will evaluate the string within the global and local namespace
# of the method definition, so you can write expressions such as 'context.Share', and it will
# _just work_
@type_check(Mpc, (type_a, type_b), type_c, c='context.Share')
def func(self, context, a, b, c=context.Share(3)):
  # rest of implementation...
```
It works for `classmethod`, `staticmethod`, and functions as well as instance methods. It also supports type checking for some, but not all parameters, as well as some more complicated stuff.


Drake-Eidukas:
This PR contains 2 commits-- one for type checking, one for mixin stuff. I can split this into 2 PRs if requested, or even squash them, but I think keeping this as 2 commits is the best.

Drake-Eidukas:
An alternative approach to the type-checker proposed here could be to use standard type annotations (#223 ) in addition with working `mypy` into our workflow, but I personally think that the type hints clutter up method declarations way more. Furthermore, when enabled in debug mode, we can freely add as much type checking as we want, then simply make the decorator a no-op in production, which should remove any overhead of simple type checking (although some duck typing will still be needed in implementations)

amiller:
This looks awesome. While we're working at this level, could you take all the `Share`, `ShareArray`, `ShareFuture`, `GFElementFuture` objects out of `mpc.py`, and also include them somewhere within `progs/mixins/`? Possibly as `progs/mixins/dataflow.py`? The reason is that then the base context `mpc.py` can have a very narrow interface, just in terms of `open_share` and `open_share_array`, which just operate on lists of field elements (they don't care about share objects)

[edit] Alternately, I'm OK simply deferring that to a later step. It shouldn't make much difference for programs using this interface.

amiller (changes requested):
I'd love to take this all as it is. 
There are two things I'd like to batch into this, just because they'll be easier now than any other time:
- can we move `progs` into `honeybadgermpc/progs`? 
- can you move all the `Share` objects out of MPC py, so it is clear they are part of the programming convenience layer and not the core protocol?

amiller (approved):

Drake-Eidukas:
Requested changes have been made, all tests pass, merging this in!

PR #261, opened 2019-04-22
remove logging from router.py

amiller:
In my experience this is making tests (especially jubjub tests, which are slow) up to 33% faster

### DIFF
--- a/honeybadgermpc/router.py
+++ b/honeybadgermpc/router.py
@@ -1,5 +1,5 @@
 import asyncio
-import logging
+# import logging
 
 
 def simple_router(n):
@@ -12,7 +12,7 @@ def simple_router(n):
 
     def make_send(i):
         def _send(j, o):
-            logging.debug('SEND %8s [%2d -> %2d]' % (o, i, j))
+            # logging.debug('SEND %8s [%2d -> %2d]' % (o, i, j))
             # delay = random.random() * 1.0
             # asyncio.get_event_loop().call_later(delay, mbox[j].put_nowait,(i,o))
             mbox[j].put_nowait((i, o))
@@ -22,7 +22,7 @@ def simple_router(n):
     def make_recv(j):
         async def _recv():
             (i, o) = await mbox[j].get()
-            logging.debug('RECV %8s [%2d -> %2d]' % (o, i, j))
+            # logging.debug('RECV %8s [%2d -> %2d]' % (o, i, j))
             return (i, o)
 
         return _recv


honeybadgermpc/router.py
```diff
@@ -12,7 +12,7 @@ def simple_router(n):
 
     def make_send(i):
         def _send(j, o):
-            logging.debug('SEND %8s [%2d -> %2d]' % (o, i, j))
```
Drake-Eidukas, line 15: Would making this an `info` debugging call be able to get performance improvements since our tests run with `DEBUG` logging?

amiller:
~There's something suspicious with what I did. First of all, when I ran this test to check that it was adding overhead even when it shouldn't, I changed the log level to `pytest -vs --log-level=INFO`, but that's changing logging in the wrong direction. To exclude the debug message, I should have set it to `--log-level=WARNING`.~

~But a second problem... 
a) Even when I change this code to `logging.info`, I still see info messages printed out in the test
b) Even when I change `pytest -vs --log-level=WARNING`, I still see debug messages.
So, I think the real problem is that I'm misundersatnding something about how pytest logging is run. Will update later...~

I've been mixing up the logging order.... it goes `DEBUG`, `INFO`, `WARNING`, `ERROR`. So it does make sense to leave this as "debug" if anything at all.

By default, however, our pytest seems to be running in `DEBUG` which is quite noisy. Still it makes tests faster to remove this.

smkuls:
The default log level for tests is specified in `pytest.ini`. If you don't want debug messages and only want info messages you set the log level to INFO.

Drake-Eidukas:
# Speedup
I was curious, so I compiled some statistics from my machine.
I ran using `docker-compose run --rm honeybadgermpc pytest -v` with the following parallelisms

## Original code with DEBUG test logging level

- `-nauto --dist=loadfile`: 53.37s
- `-n0`: 106.51s

---

## With no logging in router with DEBUG test logging level

- `-nauto --dist=loadfile`: 59.24s
- `-n0`: 92.31s

---

## Original code with INFO test logging level

- `-nauto --dist=loadfile`: 50.92s
- `-n0`: 96.74s

---

## Original code with WARNING test logging level

- `-nauto --dist=loadfile`: 51.0s
- `-n0`: 95.47

---

## With no logging in router with WARNING test logging level

- `-nauto --dist=loadfile`: 50.65s
- `-n0`: 94.42s

Drake-Eidukas:
I'd be curious what the results of editing pytest.ini would be for you @amiller, as I don't get quite the drastic speedup you do by commenting this out. 

Regardless, we should probably set travis's logging to CRITICAL (if something fails in travis, we're going to re-run it locally, and we can set this to DEBUG if we need to debug something in travis).
Personally, I've actually been using this debug statement for debugging, it's actually quite useful when debugging code hangs for me

smkuls:
Looks like python's logger is not as smart as `log4j` which actually omits the entire statement when you generate a non debug build.

This seems to be our best bet of solving this:
https://stackoverflow.com/questions/17579775/checking-for-debug-and-some-other-condition-in-python

I wouldn't want to comment the logging out completely because there are times when you need to actually see what messages are being sent/received.

Drake-Eidukas:
Closing this, as this has been carried out in #249

PR #265, opened 2019-04-24
testing n=3

Drake-Eidukas:


smkuls:
What's happening here? Is something failing only on Travis?

amiller:
It reaches a point where it hangs. It does this even locally. But it reaches a different test before hanging locally than it does on travis. And the test it hangs on is not the one that is modified in this code. So I'm not sure what's going on!

smkuls:
We can simulate a Travis build on local without having to push changes and then wait for the Travis build, that could help in iterating faster.

Drake-Eidukas:
It actually wasn't hanging locally for me, I was doing this to see if that exception would get raised when it hung.

Drake-Eidukas:
I think if we speed up travis startup time, we can just run the tests without xdist and these problems will go away.

smkuls:
That's actually not a good solution since this suggests that there's some code which is not thread safe. We should be able to run our tests in parallel. 

The only code that I can think of which has a session scope is the use of preprocessed elements. In fact, it wasn't written keeping concurrent access in mind but I am not sure if that could be a cause of this particular issue.

PR #266, opened 2019-04-24
ZMQ Implementation + Random/Triple Refinement refactoring + Minor Cython changes

smkuls:
**Implement ZMQ based IPC code**

* Use ZMQ for sending/receiving messages across nodes.
* Update ProcessProgramRunner to allow program results to be returned as
Futures instead of waiting for all at the end.
* Update the config value of k to 32 this will ensure that all nodes get
enough time to read and process the inputs
* Change logging from DEBUG to INFO for default case
* Making AVSS Value Processor benchmark take on an average k values

**Refactoring random refinement**
* Move random refinement to progs
* Add tests for random refinement
* Add benchmarks for random refinement

**Add benchmark code and make minor cython changes** 
* Cython code unnecessarily iterates over the arrays multiple times

### DIFF
--- a/honeybadgermpc/progs/triple_refinement.py
+++ b/honeybadgermpc/progs/triple_refinement.py
@@ -3,10 +3,10 @@ import asyncio
 import itertools
 
 
-def rename_and_unpack_inputs(field, a_, b_, c_, d, m):
+def rename_and_unpack_inputs(a_, b_, c_, d, m):
     n = m // 2  # number of triples
 
-    def pad(arr, k): return arr + [field(0) for _ in range(d-len(arr))]
+    def pad(arr, k): return arr + [0]*(d-len(arr))
     a, b, c = pad(a_[:n], d), pad(b_[:n], d), pad(c_[:n], d)
     x, y, z = pad(a_[n:], d), pad(b_[n:], d), pad(c_[n:], d)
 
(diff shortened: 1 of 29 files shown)


honeybadgermpc/progs/triple_refinement.py
```diff
@@ -3,19 +3,19 @@
...
     x, y, z = pad(a_[n:], d), pad(b_[n:], d), pad(c_[n:], d)
 
     return a, b, c, x, y, z
 
 
 def get_extrapolated_values(poly, a, b, d, omega):
-    a_ = poly.interp_extrap(poly.interp_extrap(a, omega**2), omega)
-    b_ = poly.interp_extrap(poly.interp_extrap(b, omega**2), omega)
+    a_ = poly.interp_extrap_cpp(poly.interp_extrap_cpp(a, omega**2), omega)
+    b_ = poly.interp_extrap_cpp(poly.interp_extrap_cpp(b, omega**2), omega)
```
amiller, line 18: why isn't interp_extrap once enough?

smkuls: I am just moving the code that I had used for benchmarking, this does not include the single interp_extrap like we discussed. I want to take that in a separate PR.

smkuls: This is anyway triple refinement, single `interp_extrap` will work only with random refinement.

amiller (approved):
I scanned through all the changes. I looked at the ZMQ implementation a few days ago when first written and it made sense at the time! Way more reliable than what we had with sockets before. Everything else looks fine

PR #268, opened 2019-04-25
Encoding/Decoding Optimisations

rahulgovind:
Overview of commit

## Improved FFT
FFT has algorithmic improvements and improvements based on using NTL better. Here are some changes,
- Previously `_fft` was optimized to decrease the number of memory allocations. However, this led to a lot of ZZ_p objects being moved around which is fairly expensive. Instead, `_fft` makes more allocations now and moves around objects less often.
- Use procedural versions of NTL operations wherever possible. This also reduces the number of copies. For example, `A = B * C` -> `mul(A, B, C)`
- Some small optimisations are included to reduce computation based on how much of the FFT output we actually need. For example, reduce computation if we wanted only the first 5 values of the FFT output when using n=8.
- Moving towards a hybrid approach, FFT also uses Vandermonde matrices internally. The vandermonde matrices are precomputed when `fft` is called for the first time, and stored as a global variable. Access is protected with a mutex to enable multiple threads to use these precomputed values.

## Improved FNT Decode
- Q(.) in `fnt_decode_step2` is now only calculated upto `t+1` coefficients rather than up to `n` coefficients as suggested in the paper. This decreases the cost of the operation `P = Q * A` since Q is smaller.
- Use procedural versions of NTL operations wherever possible

## Better conversion between Python and NTL objects
Thanks to Chirantan's work on finding better techniques to convert between NTL objects and Python Integers, these operations have significantly improved.

## Parallelization!!
- `fft_batch_evaluate` and `fft_batch_interpolate` are parallelized using OpenMP. 
- Matrix multiplications used in `vandermonde_batch_evaluate` and `vandermonde_batch_interpolate` are parallelized by NTL automatically. This requires building NTL from source.
- `batch_fft_evaluate` introduced which also makes use of parallelization for FFT-based evaluation
- `SetNumThreads` introduces a common interface to set the number of threads for both OpenMP and NTL threads

## Building NTL from source
- NTL has significant improvements in mat_ZZ_p multiplications in recent versions which help improve vandermonde-based interpolation and evaluation
- This was also required for parallelization as noted above

## Automatic selection of encoder / decoder
Finally, this is what the main target of this PR was actually supposed to be. Based on experimental results, `OptimalEncoder` and `OptimalDecoder` set the number of threads and also automatically select if Vandermonde or FFT Encoder/Decoder should be used based on `n` and the batch size.

The number of threads is always selected to be `min(physical cores, batch_size)`. There are some odd cases where the optimal number of threads might be 1 more. However, this heuristic works for most cases.

### Decoder selection

Vandermonde-based interpolation was observed to always be better than FFT in the given range of `n` values {4, ..., 8196} at a large enough batch size. Therefore, for selecting Vandermonde, I only sought to find the exact batch size at which Vandermonde is better. The final selection was that if `batch_size > 0.5 * n * num_threads` then Vandermonde should be used, else FFT-based interpolation should be used. The `num_threads` parameter tries to fit observations for experiments which were run on both `t2.medium` and `c5.x4large` instances.

### Encoder selection

FFT-based evaluation is faster than evaluation using vandermonde matrices at almost all `n` where `n` is close to the nearest power of 2. Therefore, the selection just checks this condition by checking if `npow2 - self.n > npow2 // 4` where `npow2` is the nearest power of 2 greater than `n`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,8 +23,14 @@ RUN apt-get update && apt-get install -y \
     libgmp-dev \
     libmpc-dev \
     libmpfr-dev \
-    libntl-dev \
     libflint-dev
+# Download and build NTL from source
+# Shoup recommends not using O3
+RUN wget -qO- https://www.shoup.net/ntl/ntl-11.3.2.tar.gz | tar xzvf - \
+         &&  cd ntl-11.3.2/src \
+         && ./configure CXXFLAGS="-g -O2 -fPIC -march=native -pthread -std=c++11" \
+         && make -j4 \
+         && make install
 
 # Sets default directory for running the rest of the commands
 WORKDIR /usr/src/HoneyBadgerMPC
--- a/honeybadgermpc/ntl/ctypes.pxd
+++ b/honeybadgermpc/ntl/ctypes.pxd
@@ -1,23 +1,24 @@
 from libcpp.vector cimport vector
 
 cdef extern from "ntlwrapper.h":
-    cdef cppclass ZZ_c "ZZ":
+    cdef cppclass ZZ "ZZ":
+        ZZ() except +
         pass
 
     cdef cppclass ZZ_pContext_c "ZZ_pContext":
         void restore()
         pass
 
-    cdef cppclass ZZ_p_c "ZZ_p":
-        pass
+    cdef cppclass ZZ_p "ZZ_p":
+        ZZ_p() except +
 
     cdef cppclass ZZ_pX_c "ZZ_pX":
         void SetMaxLength(size_t)
         pass
 
     cdef cppclass vec_ZZ_p:
-        ZZ_p_c& operator[](size_t)
-        void SetLength(int n)
+        ZZ_p& operator[](size_t)
+        void SetLength(int n) nogil
         void kill()
         size_t length()
     cdef cppclass mat_ZZ_p:
@@ -25,17 +26,18 @@ cdef extern from "ntlwrapper.h":
         vec_ZZ_p& operator[](size_t)
         void kill()
 
-    void ZZ_p_init "ZZ_p::init"(ZZ_c x)
-    void SetNumThreads(int n)
-    void ZZ_conv_from_int "conv"(ZZ_c x, int i)
-    void ZZ_p_conv_from_int "conv"(ZZ_p_c x, int i)
-    void ZZ_conv_to_int "conv"(int i, ZZ_c x)
-    void ZZ_conv_from_long "conv"(ZZ_c x, long l)
-    void ZZ_conv_to_long "conv"(long l, ZZ_c x)
+
+    void ZZ_p_init "ZZ_p::init"(ZZ x) nogil
+    void SetNTLNumThreads_c "SetNumThreads"(int n)
     void mat_ZZ_p_mul "mul"(mat_ZZ_p x, mat_ZZ_p a, mat_ZZ_p b)
     void mat_ZZ_p_mul_vec "mul"(vec_ZZ_p x, mat_ZZ_p a, vec_ZZ_p b) nogil
-    void ZZ_pX_get_coeff "GetCoeff"(ZZ_p_c r, ZZ_pX_c x, int i)
-    void ZZ_pX_set_coeff "SetCoeff"(ZZ_pX_c x, int i, ZZ_p_c a)
-    void ZZ_pX_eval "eval" (ZZ_p_c b, ZZ_pX_c f, ZZ_p_c a)
-    void SqrRootMod "SqrRootMod"(ZZ_c x, ZZ_c a, ZZ_c n)
+    void ZZ_pX_get_coeff "GetCoeff"(ZZ_p r, ZZ_pX_c x, int i)
+    void ZZ_pX_set_coeff "SetCoeff"(ZZ_pX_c x, int i, ZZ_p a)
+    void ZZ_pX_eval "eval" (ZZ_p b, ZZ_pX_c f, ZZ_p a)
+    void SqrRootMod "SqrRootMod"(ZZ x, ZZ a, ZZ n)
     int AvailableThreads()
+    ZZ ZZFromBytes(const unsigned char*, long)
+    unsigned char* bytesFromZZ(ZZ x)
+    ZZ_p to_ZZ_p(ZZ)
+    ZZ to_ZZ "rep"(ZZ_p)
+    int ZZNumBytes "NumBytes"(ZZ)
--- a/honeybadgermpc/ntl/helpers.pxd
+++ b/honeybadgermpc/ntl/helpers.pxd
@@ -1,21 +1,24 @@
-from .ctypes cimport ZZ_c, mat_ZZ_p, vec_ZZ_p, ZZ_p_c, ZZ_pX_c
+from .ctypes cimport ZZ, mat_ZZ_p, vec_ZZ_p, ZZ_p, ZZ_pX_c
 from libcpp.vector cimport vector
 from libcpp cimport bool
 
 cdef extern from "rsdecode_impl.h":
-    cdef void interpolate_c "interpolate"(vector[ZZ_c] r, vector[ZZ_c] x,
-                                          vector[ZZ_c] y, ZZ_c modulus)
-    cdef bool vandermonde_inverse_c "vandermonde_inverse"(mat_ZZ_p r, vector[ZZ_c] x,
-                                                          ZZ_c modulus)
+    cdef void interpolate_c "interpolate"(vector[ZZ] r, vector[ZZ] x,
+                                          vector[ZZ] y, ZZ modulus)
+    cdef bool vandermonde_inverse_c "vandermonde_inverse"(mat_ZZ_p r, vector[ZZ] x,
+                                                          ZZ modulus)
     cdef void set_vm_matrix_c "set_vm_matrix"(mat_ZZ_p r, vec_ZZ_p x_list,
-                                              int d, ZZ_c modulus)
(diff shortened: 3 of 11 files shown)


Dockerfile
```diff
@@ -23,8 +23,14 @@ RUN apt-get update && apt-get install -y \
     libgmp-dev \
     libmpc-dev \
     libmpfr-dev \
-    libntl-dev \
     libflint-dev
+# Download and build NTL from source
+# Shoup recommends not using O3
+RUN wget https://www.shoup.net/ntl/ntl-11.3.2.tar.gz && \
```
smkuls: Oh no!

How much does this add to the build time? :|

honeybadgermpc/ntl/helpers.pxd
```diff
@@ -1,21 +1,24 @@
...
+    cdef void interpolate_c "interpolate"(vector[ZZ] r, vector[ZZ] x,
+                                          vector[ZZ] y, ZZ modulus)
+    cdef bool vandermonde_inverse_c "vandermonde_inverse"(mat_ZZ_p r, vector[ZZ] x,
+                                                          ZZ modulus)
     cdef void set_vm_matrix_c "set_vm_matrix"(mat_ZZ_p r, vec_ZZ_p x_list,
-                                              int d, ZZ_c modulus)
-    cdef void fft_c "fft"(vec_ZZ_p r, vec_ZZ_p coeffs, ZZ_p_c omega, int n)
+                                              int d)
+    cdef void fft_c "fft"(vec_ZZ_p r, vec_ZZ_p coeffs, ZZ_p omega, int n)
+    cdef void fft2_c "fft2"(vec_ZZ_p r, vec_ZZ_p coeffs, ZZ_p omega,
```
smkuls: name it something else so we know what's the difference?

honeybadgermpc/ntl/rsdecode_impl.h
```diff
@@ -4,10 +4,63 @@
 #include <NTL/vec_ZZ_p.h>
 #include <vector>
 #include <iostream>
+#include <map>
+#include <omp.h>
 
 using namespace NTL;
 using namespace std;
 
+#define FFT_NAIVE_THRESHOLD 16
```
smkuls: why 16? Add a comment for some explanation?

honeybadgermpc/ntl/ctypes.pxd
```diff
@@ -1,41 +1,48 @@
...
-    void SetNumThreads(int n)
-    void ZZ_conv_from_int "conv"(ZZ_c x, int i)
-    void ZZ_p_conv_from_int "conv"(ZZ_p_c x, int i)
-    void ZZ_conv_to_int "conv"(int i, ZZ_c x)
-    void ZZ_conv_from_long "conv"(ZZ_c x, long l)
-    void ZZ_conv_to_long "conv"(long l, ZZ_c x)
+
+    void ZZ_p_init "ZZ_p::init"(ZZ x) nogil
+    void SetNTLNumThreads_c "SetNumThreads"(int n)
+    void ZZonv_from_int "conv"(ZZ x, int i)
```
smkuls: ZZonv? Is there a reason to name these this way?

honeybadgermpc/ntl/helpers.pyx
```diff
@@ -158,16 +174,25 @@ cpdef vandermonde_batch_interpolate(x, data_list, modulus):
     for i in range(n_chunks):
         l = len(data_list[i])
         for j in range(l):
-            m[j][i] = py_obj_to_ZZ_p(data_list[i][j])
+            m[j][i] = intToZZp(data_list[i][j])
         for j in range(l, k):
-            m[j][i] = py_obj_to_ZZ_p(0)
+            m[j][i] = intToZZp(0)
     cdef mat_ZZ_p reconstructions
+    end_time = time.time()
```
smkuls: Are these still needed?

honeybadgermpc/ntl/ntlwrapper.h
```diff
@@ -6,4 +6,13 @@
 #include <NTL/mat_ZZ_p.h>
 #include <NTL/BasicThreadPool.h>
 using namespace NTL;
+
+unsigned char* bytesFromZZ(const ZZ& a) {
+   long n = NumBytes(a);
+   // An extra byte read
```
smkuls: Add a comment so as to why this is needed?

tests/test_reed_solomon.py
```diff
@@ -90,6 +91,24 @@ def test_fft_encode(fft_encoding_test_cases):
         assert actual == encoded
 
 
+def test_auto_encode_fft_disabled(encoding_test_cases):
+    # Just check if some encoder is being picked
```
smkuls, line 98: I think we should validate that the correct type of the encoder gets picked for the values. If there's a bug which creeps up and for some reason we reverse them then we will never know and the code will perform worse.

smkuls (changes requested):
Some more suggestions:

- You are switching to FFT which means it now expects powers of omega for the shares, I don't see any change in the preprocessing code which reflects that.
- Add a README so as to how I can modify OpenMP/Cython code so that anyone else can make changes easily.

smkuls:
I really like your PR description but it would be better if we moved some of these to a README within the code.

Dockerfile
rahulgovind: Not much. Probably a minute or so.

honeybadgermpc/ntl/rsdecode_impl.h
rahulgovind: Experimentally determined

honeybadgermpc/ntl/helpers.pyx
rahulgovind: Ah. Missed a few. They are not needed anymore. Removing them now.

honeybadgermpc/ntl/ctypes.pxd
rahulgovind: Find-and-replace went wrong in a few places. These functions are not required anymore. Removing them.

honeybadgermpc/ntl/ntlwrapper.h
rahulgovind: Sure.

tests/test_reed_solomon.py
rahulgovind: I can't think of a good way since the selection might be platform dependent. Can you suggest a way to test this?

honeybadgermpc/ntl/helpers.pxd
rahulgovind: Sure

Dockerfile
Drake-Eidukas: Can you format this `RUN` statement like the other one in the file?
```
RUN wget -q0- https://www.shoup.net/ntl/ntl-11.3.2.tar.gz | tar xzvf - \
         &&  cd ntl-11.3.2/src \
         && ./configure CXXFLAGS="-g -O2 -fPIC -march=native -pthread -std=c++11" \
         && make -j4
         && make install
```

Drake-Eidukas: This matches the style of the dockerfile [best practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/) guide, and I think it is cleaner to just pipe the output of `wget` into `tar`, and should help reduce the docker image size.

honeybadgermpc/reed_solomon.py
```diff
@@ -374,6 +375,69 @@ def get_results(self):
...
+        if self.n < 8:
+            return self.decoder_van.decode_one(z, data)
+        else:
+            return self.decoder_fft.decode_one(z, data)
+
+    def decode_batch(self, z, data):
+        SetNumThreads(min(len(data), self.cores))
+
+        # Parameters determined experimentally
+        if self.n < 8:
```
Drake-Eidukas: You should pull 8 out into a private static constant of the class. This will make it easier to test different parameters of this on different machines, as well as make the code easier to read.

tests/test_reed_solomon.py
Drake-Eidukas: [Mocking](https://stackoverflow.com/questions/16134281/python-mocking-a-function-from-an-imported-module) is the answer here--
```
from unittest.mock import patch
import psutil

@patch('psutil.cpu_count')
def test(mocked_cpu_count):
    print(f"return value when you don't set return value: {psutil.cpu_count()}")
    mocked_cpu_count.return_value = 1
    print(f"return value after mocking: {psutil.cpu_count()}")
    mocked_cpu_count.return_value = 9000
    print(f"return value after mocking once more: {psutil.cpu_count()}")
    
print (f'normal cpu_count: {psutil.cpu_count()}')
test()
print (f'normal cpu_count after test: {psutil.cpu_count()}')
```
This outputs
```
normal cpu_count: 8
return value when you don't set return value: <MagicMock name='cpu_count()' id='140643940895544'>
return value after mocking: 1
return value after mocking once more: 9000
normal cpu_count after test: 8
```

Drake-Eidukas (changes requested):
Some style suggestions, as well as an example of how to test what samarth requested

tests/test_reed_solomon.py
rahulgovind: That's pretty cool. I'll try rewriting the test with this.

Drake-Eidukas (approved):
Looks good to me-- just rebase this and it should be good to go.

amiller:
Looks awesome. Nicely documented and tested code. I have remaining questions about how to test and choose the metaparameters like batch_thresh_slope but let's roll with this for now. The readme refers to experiments but we still need to report on and document those.

amiller (approved):

Issue #267, opened 2019-04-25
`batch_reconstruction` Doesn't properly induce faults


Drake-Eidukas:
In the [code](https://github.com/initc3/HoneyBadgerMPC/blob/9fdef08b5b9e9d8e073387bdc0c78dc9054878dc/honeybadgermpc/batch_reconstruction.py#L128-L134) for this, when we induce faults, we update `secret_shares`, however, `secret_shares` aren't sent out: `round1_chunks` are broadcasted instead. According to @smkuls, we don't currently test induced faults like this, so I'm treating this as a low priority thing.

PR #269, opened 2019-04-25
Fix fault inducing in batch_reconstruction and refactors

Drake-Eidukas:
I started by fixing the issue from #267, but ended up refactoring some of the code I came across that I found to be hard to read to be cleaner / more documented.

### DIFF
--- a/honeybadgermpc/batch_reconstruction.py
+++ b/honeybadgermpc/batch_reconstruction.py
@@ -4,90 +4,29 @@ from .polynomial import EvalPoint
 import logging
 from collections import defaultdict
 from asyncio import Queue
-from math import ceil
 import time
 from .reed_solomon import Algorithm, EncoderFactory, DecoderFactory, RobustDecoderFactory
 from .reed_solomon import IncrementalDecoder
 import random
+from honeybadgermpc.utils import chunk_data, flatten_lists, transpose_lists
 
 
-async def fetch_one(aws):
-    aws_to_idx = {aws[i]: i for i in range(len(aws))}
-    pending = set(aws)
-    while len(pending) > 0:
-        done, pending = await asyncio.wait(pending,
-                                           return_when=asyncio.FIRST_COMPLETED)
-        for d in done:
-            yield (aws_to_idx[d], await d)
+async def fetch_one(awaitables):
+    """ Given a list of awaitables, run them concurrently and
+    return them in the order they complete
 
+    args:
+        awaitables: List of tasks to run concurrently
 
-def subscribe_recv(recv):
-    tag_table = defaultdict(Queue)
-    taken = set()  # Replace this with a bloom filter?
-
-    async def _recv_loop():
-        while True:
-            j, (tag, o) = await recv()
-            tag_table[tag].put_nowait((j, o))
-
-    def subscribe(tag):
-        # take everything from the queue
-        # further things sent directly
-        assert tag not in taken
-        taken.add(tag)
-        return tag_table[tag].get
-
-    _task = asyncio.create_task(_recv_loop())
-    return _task, subscribe
-
-
-def wrap_send(tag, send):
-    def _send(j, o):
-        send(j, (tag, o))
-
-    return _send
-
-
-def recv_each_party(recv, n):
-    queues = [Queue() for _ in range(n)]
-
-    async def _recv_loop():
-        while True:
-            j, o = await recv()
-            queues[j].put_nowait(o)
-
-    _task = asyncio.create_task(_recv_loop())
-    return _task, [q.get for q in queues]
-
-
-def to_chunks(data, chunk_size, default=0):
-    """Chunkize data into `chunk_size` length pieces
-
-    If len(data) is not a multiple of chunk_size, then the default value is appended
-    to the last chunk to make sure the chunk is of size `chunk_size`
+    output:
+        Yields tuples of the form (idx, result) in the order that the tasks finish
     """
-    res = []
-    n_chunks = ceil(len(data) / chunk_size)
-    for j in range(n_chunks):
-        start = chunk_size * j
-        stop = chunk_size * (j + 1)
-        chunk = data[start:stop]
-        if len(chunk) < chunk_size:
-            chunk += [default] * (chunk_size - len(chunk))
-        res.append(chunk)
-    return res
-
-
-def merge_lists(lists):
-    result = []
-    for l in lists:
-        result += l
-    return result
-
-
-def list_transpose(lst):
-    """Transpose of list of lists"""
-    return [[lst[j][i] for j in range(len(lst))] for i in range(len(lst[0]))]
+    mapping = {elem: idx for (idx, elem) in enumerate(awaitables)}
+    pending = set(awaitables)
+    while len(pending) > 0:
+        done, pending = await asyncio.wait(pending, return_when=asyncio.FIRST_COMPLETED)
+        for d in done:
+            yield(mapping[d], await d)
 
 
 async def incremental_decode(receivers, encoder, decoder, robust_decoder, batch_size, t,
@@ -101,7 +40,67 @@ async def incremental_decode(receivers, encoder, decoder, robust_decoder, batch_
         if inc_decoder.done():
             result, _ = inc_decoder.get_results()
             return result
+
     return None
+
+
+def subscribe_recv(recv):
+    """ Given the recv method for this batch reconstruction,
+    create a background loop to put the received events into
+    the appropriate queue for the tag
+
+    Returns _task and subscribe, where _task is to be run in
+    the background to forward events to the associated queue,
+    and subscribe, which is used to register a new tag/queue pair
+
+    TODO: move this out of this file
+    """
+    # Stores the queues for each subscribed tag
+    tag_table = defaultdict(Queue)
+    taken = set()  # Replace this with a bloom filter?
+
+    async def _recv_loop():
+        while True:
+            # Whenever we receive a share array, directly put it in the
+            # appropriate queue for that round
+            j, (tag, o) = await recv()
+            tag_table[tag].put_nowait((j, o))
+
+    def subscribe(tag):
+        # TODO: make this raise an exception
+        # Ensure that this tag has not been subscribed to already
+        assert tag not in taken
+        taken.add(tag)
+
+        # Return the getter of the queue for this tag
+        return tag_table[tag].get
+
+    _task = asyncio.create_task(_recv_loop())
+    return _task, subscribe
+
+
+def recv_each_party(recv, n):
+    """ Given a recv function and number of parties,
+    creates a set of queues for each party, and forwards
+    any recv event to the respective queue for each party
+
+    args:
+        recv: async function that eventually returns a received object
+        n: number of nodes
+
+    output:
+        tuple of a background task to forward elements to the correct queue,
+        and a list of recv functions that corresponds to each node.
+    """
+    queues = [Queue() for _ in range(n)]
+
+    async def _recv_loop():
+        while True:
+            j, o = await recv()
+            queues[j].put_nowait(o)
+
+    _task = asyncio.create_task(_recv_loop())
+    return _task, [q.get for q in queues]
 
 
 async def batch_reconstruct(secret_shares, p, t, n, myid, send, recv, config=None,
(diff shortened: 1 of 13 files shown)


honeybadgermpc/reed_solomon.py
```diff
@@ -141,36 +147,29 @@ def __init__(self, d, point):
     def robust_decode(self, z, encoded):
```
Drake-Eidukas, line 150: wasn't a fan of how there was a lot of shared logic, so hopefully this helps to clean up the control flow.

honeybadgermpc/reed_solomon.py
```diff
@@ -313,37 +311,35 @@ def _robust_update(self):
...
     # Public API
     def add(self, idx, data):
-        if self.done() or idx in self._available_points or \
-                idx in self._confirmed_errors:
+        if self.done() or idx in [*self._available_points, *self._confirmed_errors]:
             return
 
         if self._validate(data) is False:
-            logging.error("Logging failed for data from %d: %s", idx, str(data))
+            logging.error("Validation failed for data from %d: %s", idx, str(data))
```
Drake-Eidukas, line 348: Pretty sure this was a typo before...

honeybadgermpc/batch_reconstruction.py
```diff
@@ -118,51 +150,59 @@ def list_transpose(lst):
...
+                                         {"node_id": myid})
 
-    fp = GF(p)
-    secret_shares = [v.value for v in secret_shares]
-    round1_chunks = to_chunks(secret_shares, t + 1)
-    num_chunks = len(round1_chunks)
+    # (optional) Induce faults
 
+    secret_shares = [v.value for v in secret_shares]
     if config is not None and config.induce_faults:
```
Drake-Eidukas, line 131: Moving this chunk right here to execute before assigning `round1_chunks = to_chunks(secret_shares, t+1)` was really the main bugfix of this PR.

honeybadgermpc/batch_reconstruction.py
```diff
@@ -4,51 +4,127 @@
...
 import time
 from .reed_solomon import Algorithm, EncoderFactory, DecoderFactory, RobustDecoderFactory
 from .reed_solomon import IncrementalDecoder
 import random
 
 
-async def fetch_one(aws):
-    aws_to_idx = {aws[i]: i for i in range(len(aws))}
-    pending = set(aws)
+def wrap_send(tag, send):
```
smkuls: I am not sure if you want to take this now but we should move these to a separate module in itself. This and all the other helpers to split receive channels and wrap sends. I have tried to reuse this code in other places as well so a separate module would be better. Also, this has got nothing to do with batch reconstruction.

honeybadgermpc/batch_reconstruction.py
```diff
@@ -4,51 +4,127 @@
...
+    def _send(j, o):
+        send(j, (tag, o))
+
+    return _send
+
+
+def chunk_data(data, chunk_size, default=0):
+    """ Break data into chunks of size `chunk_size`
+    Last chunk is padded with the default value to have `chunk_size` length
+    e.g. chukn_data([1,2,3,4,5], 2) => [[1,2], [3,4], [5, 0]]
```
smkuls: typo: chunk

honeybadgermpc/batch_reconstruction.py
```diff
@@ -4,51 +4,127 @@
...
-    aws_to_idx = {aws[i]: i for i in range(len(aws))}
-    pending = set(aws)
+def wrap_send(tag, send):
+    def _send(j, o):
+        send(j, (tag, o))
+
+    return _send
+
+
+def chunk_data(data, chunk_size, default=0):
```
smkuls: Maybe we should move all the helpers to a `utils.py`, these have got nothing to do with batch reconstruction..

honeybadgermpc/batch_reconstruction.py
```diff
@@ -4,51 +4,127 @@
...
+            # Whenever we receive a share array, directly put it in the
+            # appropriate queue for that round
             j, (tag, o) = await recv()
             tag_table[tag].put_nowait((j, o))
 
     def subscribe(tag):
-        # take everything from the queue
-        # further things sent directly
+        # Ensure that this tag has not been subscribed to already
         assert tag not in taken
```
smkuls, line 72: NYC: This should not be an assert, this should raise an Exception - we disable asserts when running and this would break a lot of things if the messages went to the wrong queue.

honeybadgermpc/batch_reconstruction.py
```diff
@@ -118,51 +150,59 @@ def list_transpose(lst):
...
+
+    # Step 1: Compute the polynomial P1, then send the elements
     start_time = time.time()
+
     encoded = enc.encode(round1_chunks)
-    to_send = list_transpose(encoded)
-    for j in range(n):
-        send(j, ('R1', to_send[j]))
+    to_send = transpose_lists(encoded)
+    [send(idx, ('R1', elem)) for idx, elem in enumerate(to_send)]
```
smkuls: I am not sure what's the convention to have list comprehensions without having their values assigned to a variable, can you check? Won't a list comprehension create a list in the memory which will eventually need to be garbage collected?

honeybadgermpc/batch_reconstruction.py
```diff
@@ -192,16 +233,11 @@ def list_transpose(lst):
...
 
-    task_r1.cancel()
-    task_r2.cancel()
-    subscribe_task.cancel()
-    for q in data_r1:
-        q.cancel()
-    for q in data_r2:
-        q.cancel()
+    # Cancel all created tasks
+    [task.cancel() for task in [task_r1, task_r2, subscribe_task, *data_r1, *data_r2]]
```
smkuls: same here

honeybadgermpc/reed_solomon.py
```diff
@@ -313,37 +311,35 @@ def _robust_update(self):
...
+
         # We're done
         if self._num_decoded == self.batch_size:
             self._result = self._partial_result
 
     # Public API
     def add(self, idx, data):
-        if self.done() or idx in self._available_points or \
-                idx in self._confirmed_errors:
+        if self.done() or idx in [*self._available_points, *self._confirmed_errors]:
```
smkuls: similar comment here, will this create a new list which can take up additional memory?

honeybadgermpc/mpc.py
```diff
@@ -186,6 +169,23 @@ def _send(j, o):
...
+        result = asyncio.create_task(self.prog(self, **self.prog_args))
+        await asyncio.wait((bgtask, result), return_when=asyncio.FIRST_COMPLETED)
+
+        if result.done():
+            bgtask.cancel()
+            return result.result()
+        else:
+            logging.info(f'bgtask exception: {bgtask.exception()}')
+            raise bgtask.exception()
+            # FIXME: This code is unreachable and needs to be investigated
```
smkuls, line 185: NYC: I think it is safe to remove any code after `raise`

honeybadgermpc/reed_solomon.py
```diff
@@ -32,7 +35,7 @@ def encode_batch(self, data):
         raise NotImplementedError
 
 
-class Decoder(object):
+class Decoder(ABC):
```
smkuls, line 40: Does this not require any change in the Encoder/DecoderFactory?

honeybadgermpc/reed_solomon.py
```diff
@@ -313,37 +311,35 @@ def _robust_update(self):
...
+            self._partial_result.append(decoded)
+
+            # Errors detected considered to be confirmed errors
+            self._confirmed_errors |= set(errors)
+            self._available_points -= set(errors)
+
+            for e in errors:
+                error_idx = self._z.index(e)
+
+                del self._z[error_idx]
```
smkuls, line 332: NYC: Is this a list, how does delete from a list work? Does it create a new list? Can we not replace it with `None` and reuse the same list?

honeybadgermpc/reed_solomon.py
```diff
@@ -313,37 +311,35 @@ def _robust_update(self):
...
+                for i in range(len(self._available_data)):
+                    del self._available_data[i][error_idx]
```
smkuls, line 334: Same here

smkuls (changes requested):
Some comments are not on your code but since you started refactoring I thought I will give my comments.

honeybadgermpc/batch_reconstruction.py
```diff
@@ -4,51 +4,127 @@
...
+def chunk_data(data, chunk_size, default=0):
+    """ Break data into chunks of size `chunk_size`
+    Last chunk is padded with the default value to have `chunk_size` length
+    e.g. chukn_data([1,2,3,4,5], 2) => [[1,2], [3,4], [5, 0]]
+    """
+    # Main chunking
+    res = [data[start:(start + chunk_size)] for start in range(0, len(data), chunk_size)]
+
+    # Pad final chunk with default value
+    res[-1] += [default] * (chunk_size - len(res[-1]))
```
rahulgovind: This will raise an IndexError is `res` is an empty array. In the off chance that this is used else where it would be better to fix it here.

honeybadgermpc/batch_reconstruction.py
```diff
@@ -4,51 +4,127 @@
...
+
+    return res
+
+
+def flatten_lists(lists):
+    """ Given a 2d list, return a flattened 1d list
+    e.g. [[1,2,3],[4,5,6],[7,8,9]] => [1,2,3,4,5,6,7,8,9]
+    """
+    return [elem for inner in lists for elem in inner]
```
rahulgovind: This is not very readable since it's not clear what the precedence of the `for`-`in` operation is. This is also considerably slower than `merge_lists`. `list(itertools.chain(lists))` might be a better alternative.

honeybadgermpc/reed_solomon.py
```diff
@@ -141,36 +147,28 @@ def __init__(self, d, point):
...
+        if len(error_poly) > 1:
+            if self.use_fft:
+                err_eval = fft(error_poly, self.point.omega.value,
+                               self.modulus, self.point.order)[:self.point.n]
+            else:
+                x = [self.point(i).value for i in range(self.point.n)]
+                err_eval = vandermonde_batch_evaluate(x, [error_poly], self.modulus)[0]
+
+            errors = [i for i in range(self.point.n) if err_eval[i] == 0]
+
```
rahulgovind, line 172: At this point, I think this entire operation can be refactored using simply an `OptimalEncoder`

honeybadgermpc/reed_solomon.py
rahulgovind: Definitely much easier to read now

rahulgovind (changes requested):
Mostly checked for logic and efficiency. I think the refactor is great, and the changes and comments definitely make the code easier to follow. 

At this point, I really think most of the operations we do with anything more than a 1-D list should be replaced with `numpy.ndarray` or something equivalent. A fair amount of logic is involved in just handling the shape of the lists - transposing, checking that the lengths in a 2-D list are consistent, etc. Maybe this can be taken up in a future PR. Additionally, Cython can also handle numpy arrays better than lists. But I'm not sure how well it does when the `dtype` is not a primitive data type.

honeybadgermpc/reed_solomon.py
rahulgovind: Can be taken up after #268

honeybadgermpc/batch_reconstruction.py
Drake-Eidukas: I agree. I'm thinking this might be worked into an overhaul of share openings.

honeybadgermpc/batch_reconstruction.py
Drake-Eidukas: I looked it up, and you're right-- this isn't pythonic.

honeybadgermpc/reed_solomon.py
Drake-Eidukas: None actually-- I don't believe we create any plain `Decoder` objects anywhere.

honeybadgermpc/reed_solomon.py
Drake-Eidukas: [good explanation for dicts, i believe it's the same for lists](https://stackoverflow.com/a/36716057)
Essentially, this contains the overhead of removing the symbol from namespaces, but doesn't actually deallocate memory (unless this was the final remaining reference to the object, in which case I think it will get garbage collected). 
Setting this to None would likely have the same affect, and be faster.

honeybadgermpc/batch_reconstruction.py
Drake-Eidukas: good catch!

smkuls:
> Mostly checked for logic and efficiency. I think the refactor is great, and the changes and comments definitely make the code easier to follow.
> 
> At this point, I really think most of the operations we do with anything more than a 1-D list should be replaced with `numpy.ndarray` or something equivalent. A fair amount of logic is involved in just handling the shape of the lists - transposing, checking that the lengths in a 2-D list are consistent, etc. Maybe this can be taken up in a future PR. Additionally, Cython can also handle numpy arrays better than lists. But I'm not sure how well it does when the `dtype` is not a primitive data type.

I am not a big fan of adding new dependencies - each new dependencies in a long run makes things harder to maintain. If we do go the numpy way then we should make the change in the entire codebase and not in just a few places.

honeybadgermpc/reed_solomon.py
Drake-Eidukas: [turns out it's the fastest way to remove an element from a dict](https://stackoverflow.com/questions/11277432/how-to-remove-a-key-from-a-python-dictionary)

Drake-Eidukas:
I'm curious too if the speed improvements would overcome the serialization and deserialization overhead from numpy here given that we aren't using native datatypes

honeybadgermpc/batch_reconstruction.py
Drake-Eidukas: I don't think this is a massive priority right now, since the only tags used as of now are `"S"`, `"R1"`, and  `"R2"`. I agree on principle though.

honeybadgermpc/mpc.py
Drake-Eidukas: I'll leave this for a later refactor

honeybadgermpc/batch_reconstruction.py
smkuls: This code is actually reused in a lot of other places and not just batch reconstruction. We create one tag per run of AVSS, ACS for the offline phase and that is expected to run endlessly. Since it is a very small change, I'd suggest to take it.

honeybadgermpc/reed_solomon.py
Drake-Eidukas: This will take up `8*len(self._available_points) + 8*len(self._confirmed_errors) + 22` additional bytes of memory I believe. It'll then be garbage collected during the next collection loop, as the refcount of the created list should immediately hit 0 after the if statement. I think the overhead is worth the cleaner code, but I'll make the change.

[at least, the math works if I understood this correctly](https://jakevdp.github.io/blog/2014/05/09/why-python-is-slow/#3.-Python's-object-model-can-lead-to-inefficient-memory-access)

Drake-Eidukas: that link also is making me think that numpy could be a really good idea in our critical path if we can manage to do it cleanly.

Drake-Eidukas: [Did some more reading](https://rushter.com/blog/python-memory-managment/), this will likely allocate the array in an already allocated `arena` (fairly quick, as is deallocation), so an actual os allocation is unlikely with how aggressive python memory caching is. This is likely a safe pattern outside of the case where the array created is greater than 4kb (i.e. > 500 elements), unless there's existing memory pressure.

Drake-Eidukas: This pattern when using a tuple rather than a list is much safer, as it will deallocate the tuple immediately after the if statement, as tuples use the reference-count based GC vs. the cyclic cycle detecting algorithm used for lists which only runs periodically.

Drake-Eidukas:
I addressed most of the comments aside from some of the misc. comments (such as numpy, optimalencoder, stuff with `del`).

Drake-Eidukas:
I addressed most of the comments aside from some of the misc. comments (such as numpy, optimalencoder, stuff with `del`).

honeybadgermpc/utils.py
```diff
@@ -31,19 +31,47 @@ def check_types(func, types, kwtypes, args, kwargs):
...
-        # If there are strings passed in, evaluate them as if they were
-        # evaluated within the method
-        if isinstance(t, tuple):
-            t = tuple(eval(t_, func.__globals__, passed_dict)
-                      if isinstance(t_, str) else t_ for t_ in t)
-        elif isinstance(t, str):
-            t = eval(t, func.__globals__, passed_dict)
-
-        type_dict[k] = t
+        # Turn type into array of types
```
Drake-Eidukas, line 38: Now supports type constraints like: 
```
@static_type_check('callable(a)')
def func(a):
    a('hello')
```

rahulgovind (approved):

smkuls (approved):

PR #270, opened 2019-04-25
Implement HyperInvertibleMatrix based random gen.

smkuls:
* Implement the approach described in the HyperMPC paper to generate
double sharings using HyperInvertible Matrix.
* Make the random in field rely on a unique instance and not on the
global instance.

### DIFF
--- a/honeybadgermpc/hyper_invertible.py
+++ b/honeybadgermpc/hyper_invertible.py
@@ -1,0 +1,104 @@
+import asyncio
+import logging
+from honeybadgermpc.exceptions import HoneyBadgerMPCError
+from honeybadgermpc.polynomial import EvalPoint, polynomials_over
+from honeybadgermpc.reed_solomon import Algorithm, EncoderFactory, DecoderFactory
+from honeybadgermpc.batch_reconstruction import subscribe_recv, wrap_send
+
+
+class HyperInvMessageType(object):
+    SUCCESS = "S"
+    ABORT = "A"
+
+
+async def _recv_loop(n, recv, s=0):
+    results = [None]*n
+    for _ in range(n):
+        sender_id, value = await recv()
+        results[sender_id-s] = value
+    return results
+
+
+async def generate_double_shares(n, t, my_id, _send, _recv, field):
+    poly = polynomials_over(field)
+    eval_point, algorithm = EvalPoint(field, n, use_fft=True), Algorithm.FFT
+    big_t = n - (2 * t) - 1  # This is same as `T` in the HyperMPC paper.
+    encoder = EncoderFactory.get(eval_point, algorithm)
+
+    # Pick a random element.
+    my_random = field.random()
+    coeffs_t = tuple(map(int, poly.random(t, my_random).coeffs))
+    coeffs_2t = tuple(map(int, poly.random(2*t, my_random).coeffs))
+
+    # Generate t and 2t shares of the random element.
+    shares_to_send = (encoder.encode(coeffs_t), encoder.encode(coeffs_2t))
+
+    subscribe_recv_task, subscribe = subscribe_recv(_recv)
+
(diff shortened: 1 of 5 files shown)


honeybadgermpc/hyper_invertible.py
```diff
@@ -0,0 +1,104 @@
...
+        send(i, (shares_to_send[0][i], shares_to_send[1][i]))
+
+    # Wait until all shares are received.
+    received_shares = await share_recv_task
+    unrefined_t_shares, unrefined_2t_shares = zip(*received_shares)
+
+    # Apply the hyper-invertible matrix.
+    # Assume the unrefined shares to be coefficients of a polynomial
+    # and then evaluate that polynomial at powers of omega.
+    refined_shares = (encoder.encode(unrefined_t_shares),
```
amiller, line 56: Awesome, it's great how this can reuse the encoder for this purpose

amiller (approved):
Looks great. Can't wait to see how this works for improving offline phase throughput

honeybadgermpc/hyper_invertible.py
```diff
@@ -0,0 +1,104 @@
...
+
+class HyperInvMessageType(object):
+    SUCCESS = "S"
+    ABORT = "A"
+
+
+async def _recv_loop(n, recv, s=0):
+    results = [None]*n
+    for _ in range(n):
+        sender_id, value = await recv()
```
Drake-Eidukas, line 17: could we do this with a gather?

honeybadgermpc/hyper_invertible.py
```diff
@@ -0,0 +1,104 @@
...
+        results[sender_id-s] = value
+    return results
+
+
+async def generate_double_shares(n, t, my_id, _send, _recv, field):
+    poly = polynomials_over(field)
+    eval_point, algorithm = EvalPoint(field, n, use_fft=True), Algorithm.FFT
+    big_t = n-2*t-1  # This is same as `T` in the HyperMPC paper.
```
Drake-Eidukas:
```suggestion
big_t = n - 1 - (t<<1) 
```
Even if you stick with `2*t`, I think this expression is hard to read without spacing

honeybadgermpc/hyper_invertible.py
```diff
@@ -0,0 +1,104 @@
...
+    # Send the verification response.
+    if my_id > big_t:
+        for i in range(n):
+            send(i, response)
+
+    responses = await response_recv_task
+    subscribe_recv_task.cancel()
+    # If any of [T+1, N] parties say that the shares are inconsistent then abort.
+    if responses.count(HyperInvMessageType.SUCCESS) != n-big_t-1:
+        raise HoneyBadgerMPCError("Aborting because the shares where inconsistent.")
```
Drake-Eidukas: typo
```suggestion
        raise HoneyBadgerMPCError("Aborting because the shares were inconsistent.")
```

Drake-Eidukas (approved):
Code is overall pretty clean, just a few nitpicks

honeybadgermpc/hyper_invertible.py
smkuls: I could, but this will make the code a bit less readable since I will have to gather and then iterate again to assign the values correctly. This way it is also easier to debug since you can add a log to view what values are being received. I think it is better kept this way.

honeybadgermpc/hyper_invertible.py
smkuls: Fixed!

honeybadgermpc/hyper_invertible.py
smkuls: Added more spaces. `2*t` i s more readable than `t<<1`. This operation is being done only once and using bit shift will hardly have any impact, in fact, the interpreter might even execute it like a shift.

PR #272, opened 2019-04-27
Noticeably log when decoding fails

amiller:
We should not raise an exception when decoding fails, because the code is written to tolerate Byzantine failures. However, it indicates either actual Byzantine behavior, or more likely, an error.


Drake-Eidukas (approved):
LGTM, waiting on tests

Drake-Eidukas:
Rebased the PR

PR #274, opened 2019-04-28
Optimize fnt_decode

rahulgovind:
Underestimated the cost of division. Significant speedup by changing a division to multiplication in `fnt_decode_step2`.


amiller:
:O speedups are cool

smkuls (approved):
So an `inv` and `mul` is faster than `div`? By how much?

The changes look good to me.

PR #275, opened 2019-04-28
Refactor jubjub and shares (stability improvements)

Drake-Eidukas:
This refactors jubjub to explicitly use `ShareFuture`s, which helps clean up its interface and usage greatly. This accomplishes this my making its methods no longer async, but adding in `open` and `equals`, which are async and provide boilerplate code that was otherwise needed. This also updates how we assign ids to incoming shares (there were some issues with it).

Finally, this updates the existing code to use `create_task` instead of `ensure_future`, and updates existing jubjub code (mimc, etc.)

### DIFF
--- a/honeybadgermpc/mpc.py
+++ b/honeybadgermpc/mpc.py
@@ -13,6 +13,7 @@ from .batch_reconstruction import batch_reconstruct
 from .elliptic_curve import Subgroup
 from .preprocessing import PreProcessedElements
 from .config import ConfigVars
+from .exceptions import HoneyBadgerMPCError
 
 
 class Mpc(object):
@@ -42,9 +43,9 @@ class Mpc(object):
         self.prog = prog
         self.prog_args = prog_args
 
-        # A task representing the opened values
-        # { shareid => Future (field list(field)) }
-        self._openings = {}
+        # Counter that increments by 1 every time you access it
+        # This will be used to assign ids to shares.
+        self._share_id = 0
 
         # Store opened shares until ready to reconstruct
         # playerid => { [shareid => Future share] }
@@ -64,25 +65,50 @@ class Mpc(object):
         self.GFElementFuture = type(
             'GFElementFuture', (GFElementFuture,), {'context': self})
 
+    def _get_share_id(self):
+        """Returns a monotonically increasing int value
+        each time this is called
+        """
+        share_id = self._share_id
+        self._share_id += 1
+        return share_id
+
     def call_mixin(self, name, *args, **kwargs):
+        """Convenience method to check if a mixin is present, and call it if so
+        args:
+            name(str): Name of the mixin to call
+            args(list): arguments to pass to the call to the mixin
+            kwargs(dict): named arguments to pass to the call to the mixin
+
+        outputs:
+            future that resolves to the result of calling the mixin operation
+        """
         if name not in self.config:
             raise NotImplementedError(f"Mixin {name} not present!")
 
-        return asyncio.ensure_future(self.config[name](self, *args, **kwargs))
+        return asyncio.create_task(self.config[name](self, *args, **kwargs))
 
-    async def open_share(self, share):
+    def open_share(self, share):
         """ Given secret-shared value share, open the value by
         broadcasting our local share, and then receive the likewise
         broadcasted local shares from other nodes, and finally reconstruct
         the secret shared value.
+
+        args:
+            share (Share): Secret shared value to open
+
+        outputs:
+            Future that resolves to the plaintext value of the share.
         """
 
+        res = asyncio.Future()
+
         # Choose the shareid based on the order this is called
-        shareid = len(self._openings)
+        shareid = self._get_share_id()
         t = share.t if share.t is not None else self.t
 
         # Broadcast share
-        for j in range(self.N):
+        for dest in range(self.N):
             value_to_share = share.v
 
             # Send random data if meant to induce faults
@@ -92,7 +118,7 @@ class Mpc(object):
                 value_to_share = self.field.random()
 
             # 'S' is for single shares
-            self.send(j, ('S', shareid, value_to_share))
+            self.send(dest, ('S', shareid, value_to_share))
 
         # Set up the buffer of received shares
         share_buffer = [self._share_buffers[i][shareid] for i in range(self.N)]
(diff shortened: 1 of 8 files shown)


Drake-Eidukas:
Though this seemed to have a speedup for `prog` tests, overall, the impact is only around 10s total across all the unit tests when running with `-n0`. With `-nauto`, sometimes I see pretty good speedups, but that has more to do with the execution order of tests more than anything else (our fastest tests run first, which limits the utility of parallelizing our tests).

hanyunx (approved):
Looks good to me!

honeybadgermpc/mpc.py
```diff
@@ -64,11 +64,27 @@ def __init__(self, sid, n, t, myid, pid, send, recv, prog, config, **prog_args):
...
+            kwargs(dict): named arguments to pass to the call to the mixin
+
+        outputs:
+            future that resolves to the result of calling the mixin operation
+        """
         if name not in self.config:
             raise NotImplementedError(f"Mixin {name} not present!")
 
-        return asyncio.ensure_future(self.config[name](self, *args, **kwargs))
+        return asyncio.create_task(self.config[name](self, *args, **kwargs))
```
smkuls, line 89: why one over the other?

honeybadgermpc/mpc.py
```diff
@@ -64,11 +64,27 @@ def __init__(self, sid, n, t, myid, pid, send, recv, prog, config, **prog_args):
         self.GFElementFuture = type(
             'GFElementFuture', (GFElementFuture,), {'context': self})
 
+    def _get_share_id(self):
```
smkuls, line 68: Ummmm, there are some serious issues in the way we assign the `ShareId`. @rahulgovind and I had discussed an example in which assigning the order this way won't be consistent across all parties since the tasks are not executed in a deterministic order.

Why did you see a need to change this?

honeybadgermpc/mpc.py
```diff
@@ -92,47 +108,65 @@ def call_mixin(self, name, *args, **kwargs):
...
+            self.send(dest, ('S', shareid, value_to_share))
 
         # Set up the buffer of received shares
         share_buffer = [self._share_buffers[i][shareid] for i in range(self.N)]
 
         point = EvalPoint(self.field, self.N, use_fft=False)
 
         # Create polynomial that reconstructs the shared value by evaluating at 0
-        opening = robust_reconstruct(
+        p, _ = await robust_reconstruct(
```
smkuls: what's the second return value? Why are we ignoring it? Should we log it?

honeybadgermpc/mpc.py
```diff
@@ -92,47 +108,65 @@ def call_mixin(self, name, *args, **kwargs):
...
             share_buffer, self.field, self.N, t, point)
-        self._openings[shareid] = opening
-
-        p, _ = await opening
 
         # Return reconstructed point
         return p(self.field(0))
 
-    def open_share_array(self, sharearray):
+    async def open_share_array(self, sharearray):
```
smkuls: Why is this change required?

honeybadgermpc/mpc.py
```diff
@@ -92,47 +108,65 @@ def call_mixin(self, name, *args, **kwargs):
...
+        # bgtask should not exit early-- this should correspond to an error
+        if bgtask.done():
+            logging.error('Background task finished before prog')
+
+            bg_exception = bgtask.exception()
+            if not result.done():
+                result.cancel()
+
+            if bg_exception is None:
+                raise Exception('background task finished before prog!')
```
smkuls: `raise HoneyBadgerMpcException`, avoid raising generic exceptions!

Can you think of a case when this can happen?
As a rule, always log exceptions, we have seen things hang when there are exceptions but it is easier to spot a log.

honeybadgermpc/progs/jubjub.py
```diff
@@ -198,15 +194,24 @@ def __str__(self):
...
         if not isinstance(n, int):
             raise Exception("Can't scale a point by something which isn't an int!")
 
         return self
 
-    async def double(self):
+    def double(self):
         return self
 
+    async def equals(self, other):
```
smkuls: why is this async?

honeybadgermpc/progs/jubjub.py
```diff
@@ -198,15 +194,24 @@ def __str__(self):
...
+        if not isinstance(other, SharedIdeal):
+            return False
+
+        return self.curve == other.curve
+
+    async def open(self):
```
smkuls: Same here

tests/progs/test_jubjub.py
```diff
@@ -84,24 +80,19 @@ def test_basic_point_functionality():
...
-        p3 = await SharedPoint.create(context, context.Share(
-            0), context.Share(1), Jubjub(Jubjub.Field(-2)))
 
-        assert await shared_point_equals(p1, p1)
-        assert not await shared_point_equals(p1, p2)
-        assert not await shared_point_equals(p1, p3)
-
-    await run_test_program(_prog, test_runner)
+        # Different curve
+        p3 = deepcopy(p1)
```
smkuls: A deepcopy would be really slow, why do you need this? Can you replace this with something else?

smkuls (changes requested):
I am not aware of the JubJub and MIMC code, @hanyunxu would be the best person to review that (looks like she has already approved these changes)

honeybadgermpc/mpc.py
Drake-Eidukas: The asyncio documentation specifies that [`create_task`](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task) is preferred over `ensure_future` in python3.7. Furthermore, we mostly already mostly used `create_task`, so this makes it consistent throughout. 

When we call `ensure_future` with a coroutine,[ it simply calls `create_task` anyways](https://github.com/python/cpython/blob/56065d4c8ac03042cb7e29ffda9b1ac544a37b4d/Lib/asyncio/tasks.py#L616).

honeybadgermpc/mpc.py
Drake-Eidukas: Currently, shares are already assigned ids in this order-- we get the length of `self._openings` and assign `share_id` that way, then assign an opening to that slot of `self._openings` before ever yielding execution to another task. This simply adds an extra layer of overhead to understanding the code given that we use `self._openings` for no other reason today.

Conversely,  the new code makes the assignment of ids much more explicit, and also saves the memory that `self._openings` took up. We never deleted entries once they finished opening, and never actually use `self._openings` other than to assign these IDs.

 I think in the current state of `Mpc`, this change is better.

smkuls: Sounds good!

honeybadgermpc/mpc.py
Drake-Eidukas: The second return value is actually `_confirmed_errors` from reconstruction when successful. I don't think this is worth logging here, but this makes it clearer in the code that the function is actually returning two things. 

That said, there is some error handling to be had here-- this returns `None` in the case of failed reconstruction. We should throw an exception here.

smkuls: Do we add a log when there is an error? If not, then can you please add a log?

honeybadgermpc/mpc.py
smkuls: Shouldn't this return a `Future` which resolves to the opened values later on - as per our data flow?

Drake-Eidukas: It keeps this in line with `open`. It was weird how one returned a future but the other didn't. 

Perhaps returning a `GFElementFuture` from `open` and a list of `GFElementFuture`s for `open_share_array` and making both of these non-async would be a better idea.

tests/progs/test_jubjub.py
Drake-Eidukas: I just replaced it with `copy`, which executes much faster.

honeybadgermpc/progs/jubjub.py
Drake-Eidukas: It seems cleaner than returning a `Future` which will evaluate to `True` or `False`, as it's nonsensical to call either this or `open` without awaiting immediately (or gathering).

honeybadgermpc/progs/jubjub.py
Drake-Eidukas: See prior comment, this also is to match style of `Share` objects today.

honeybadgermpc/mpc.py
Drake-Eidukas: With the current `_recvloop`, no, but I think it's best to have this in here for completeness.

honeybadgermpc/progs/jubjub.py
smkuls: I think I should have phrased my concern better, there is no `await` in this method and the one below, so why do you even need the `async` keyword?

Drake-Eidukas: I think the semantics of SharedPoint and SharedIdeal should be the same, having this be async makes the code itself cleaner. Perhaps returning a future is better here.

Drake-Eidukas:
notable changes-- 
- `Mpc`'s `open_share`, `open_share_array` methods, as well as `SharedPoint`'s `open` and `equals` methods now return futures and are not async
- spacing edits that I found helpful when reading through the failing `test_hbavss_light_parallel_share_array_open`. 

Also, it seems that `test_hbvavss_light_parallel_share_array_open` is failing every other time that I run this due to a reconstruction error, but this persists even on `dev`, so I'm chalking it up to something other than these changes.

smkuls (approved):

PR #276, opened 2019-04-29
Update apps/ to use new mixins

Drake-Eidukas:
@smkuls  mentioned that I broke this at standup today-- here's the fix.

### DIFF
--- a/apps/shuffle/butterfly_network.py
+++ b/apps/shuffle/butterfly_network.py
@@ -1,8 +1,8 @@
 import asyncio
 import logging
 from math import log
-from honeybadgermpc.preprocessing import PreProcessedElements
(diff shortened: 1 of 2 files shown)


honeybadgermpc/utils.py
```diff
@@ -143,7 +143,7 @@ def type_check(*types, **kwtypes):
     return static_type_check(object, *types, **kwtypes)
 
 
-@static_type_check(str, 'callable(send)')
+@static_type_check(object, 'callable(send)')
```
smkuls: lets stick to `str` only? To make things work, just change line 76 in `butterfly_network.py` from `0` to `"0"`.

Drake-Eidukas: Sure thing

smkuls (approved):

PR #277, opened 2019-04-30
Refactor simple_router 

Drake-Eidukas:
This makes our `Router` code cleaner / more modular, and also slightly improves test time (~6-10 seconds in single-execution mode). This is going to facilitate implementing batched share opening.

### DIFF
--- a/honeybadgermpc/router.py
+++ b/honeybadgermpc/router.py
@@ -1,35 +1,100 @@
 import asyncio
-# import logging
+from abc import ABC, abstractmethod
+from functools import partial
+import logging
 
 
-def simple_router(n):
+class Router(ABC):
     """
-    Builds a set of connected channels
-    @return (receives, sends)
+    Base class for Router objects defining the recv and send methods.
     """
-    # Create a mailbox for each party
-    mbox = [asyncio.Queue() for _ in range(n)]
 
-    def make_send(i):
-        def _send(j, o):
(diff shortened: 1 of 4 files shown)


honeybadgermpc/router.py
```diff
@@ -1,35 +1,67 @@
...
+
+
+class SimpleRouter(Router):
+    def __init__(self, num_parties):
+        super().__init__(num_parties)
+
+        # Mailboxes for each party
+        self._queues = [asyncio.Queue() for _ in range(num_parties)]
+
+    async def _recv(self, player_id: int) -> object:
```
smkuls: why the `_` prefix? These are supposed to be public methods.

tests/test_simple_router.py
```diff
@@ -17,7 +17,7 @@
 
 
 @mark.asyncio
-async def test_simple_router(test_router):
+async def test_test_router(test_router):
```
smkuls: nit: `test_router`

smkuls (approved):
Looks good to me, except the `._` prefix in the name which I don't think should be there. `_` prefix afaik is to denote private methods, but these are public methods.

honeybadgermpc/router.py
Drake-Eidukas: At the time, the intention was to access these via the `.recvs` and `.sends` attributes, but I agree now.

PR #278, opened 2019-04-30
Benchmark jubjub operations

Drake-Eidukas:
This is reliant on #275 , and does 2 things 
- Adds benchmark tests for `Jubjub` operations. The main one of interest is for `share_mul`, which is parametrized by `bit_length`.
- Moves `random_refinement` into `honeybadgermpc/progs`

As of now, the benchmarks for the multiplication operations are hanging which is odd given that the normal tests for multiplication work just fine.


Drake-Eidukas:
So, it turns out that the hanging I was experiencing was actually batch reconstruction issues

Drake-Eidukas:
Everything works now! Should be ready for review.

benchmark/test_benchmark_jubjub.py
```diff
@@ -0,0 +1,151 @@
...
+COUNT_MAX = 2
+
+
+def test_benchmark_shared_point_add(test_preprocessing, benchmark):
+    async def _prog(context):
+        result = SharedPoint.from_point(context, TEST_POINT)
+        result = result.add(result)
+        await result.open()
+
+    for kind in TEST_PREPROCESSING:
```
smkuls: Is it possible that we can move the code from below this line to the end of this method in a separate method and then reuse it?

benchmark/test_benchmark_jubjub.py
```diff
@@ -0,0 +1,151 @@
...
+            multiplier_ += (2**idx) * m
+
+        # Compute share_mul
+        p1_ = await share_mul(context, m_bits, p)
+        result = await p1_.open()
+
+        # Assertation
+        multiplier = await multiplier_.open()
+        if multiplier == Jubjub.Field(0):
+            assert result == p
```
smkuls: You should run the benchmark code (to get the best results) with the `-O` flag which disables the asserts so I would advise not having asserts here. If you want to test something then you should add it as a test and not as a benchmark.

smkuls (changes requested):
Why does the `__init__.py` show as `No changes`? If there are no changes then maybe remove it from this commit?

Drake-Eidukas:
The `__init__.py` is empty because it's an empty file-- I was using it to make some import work correctly, but everything seems to be working just fine now. I'll remove it.

benchmark/test_benchmark_jubjub.py
Drake-Eidukas: Good point-- stripped them out.

benchmark/test_benchmark_jubjub.py
Drake-Eidukas: Good idea-- I'll add a fixture for it.

smkuls (approved):

PR #280, opened 2019-04-30
Fix client mode for batch hbavss and add a test

nqyy:
Fix some issues that came along with the client mode and add the test for the client mode for Batch HBAVSS.

There were original logical issues of AVID about the checking of certain enough message. They are moved to the back of the while loop instead of staying in the handling of a certain message.

In the client mode, the n is different. So when constructing the dispersal list for AVID, the size should be different, so we pass the ``n`` from ``avss`` to ``_get_dealer_msg`` instead of using ``self.n`` there.

### DIFF
--- a/honeybadgermpc/protocols/avid.py
+++ b/honeybadgermpc/protocols/avid.py
@@ -189,32 +189,16 @@ class AVID:
                 if sender in echo_set:
                     logger.warning("[%d] Redundant ECHO", pid)
                     continue
-
                 # Update
                 echo_set.add(sender)
-
-                if len(echo_set) >= echo_threshold and not ready_sent:
-                    ready_sent = True
-                    self.broadcast((sid, AVIDMessageType.READY))
 
             elif msg[1] == AVIDMessageType.READY:
                 # Validation
                 if sender in ready_set:
                     logger.warning("[%d] Redundant READY", pid)
                     continue
-
                 # Update
                 ready_set.add(sender)
-
-                # Amplify ready messages
-                if len(ready_set) >= ready_threshold and not ready_sent:
-                    ready_sent = True
-                    self.broadcast((sid, AVIDMessageType.READY))
-
-                if len(ready_set) >= output_threshold and len(echo_set) >= k:
-                    # update ok future indicating ready for retrieve
-                    if not self.ok_future.done():
-                        self.ok_future.set_result(True)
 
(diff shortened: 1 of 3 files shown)


honeybadgermpc/protocols/avid.py
```diff
@@ -225,3 +209,17 @@ def broadcast(self, o):
             elif msg[1] == AVIDMessageType.RESPONSE:
                 # put in the queue for retrieve
                 self.retrieval_queue.put_nowait((sender, msg))
+
+            if len(echo_set) >= echo_threshold and not ready_sent:
```
smkuls, line 213: Why was this bug not caught via the tests?

Can you add a test so that if this happens again we catch it?

smkuls (changes requested):

smkuls:
Rest of the PR looks good to me, we should catch such bugs via test - lets add tests to capture this scenario so that this does not happen again.

honeybadgermpc/protocols/avid.py
nqyy: I don't think that is necessary. There were some logical misunderstandings while I wrote the code and it's not a bug to catch. And the AVID is fully tested by all different kinds of batch HBAVSS tests anyway.

smkuls: I disagree, unit tests are meant to test things independent of other components. There is not even a single test for AVID. If I modify AVID I don't want to run HbAVSS tests to see if something is broken. You should write tests to test the AVID code.

nqyy: AVID is part of HBAVSS here. It is just put in a separate file. Modifying AVID means modifying HBAVSS. Plus, I don't think I'll have time for that in the final's week.

nqyy:
Right now, AVID is part of HBAVSS. Tests for HBAVSS are good enough for AVID currently.

honeybadgermpc/protocols/avid.py
smkuls: AVID (similar to RBC) as far as I understand is a separate protocol and must have tests. Nonetheless, it is in a separate file implying it is a separate module which can be reused.

If you don't have time, then you can open an issue and add the issue link in this PR so that this can be taken up later. I will approve this PR post that.

nqyy: Sure. Will do.

smkuls (approved):

PR #282, opened 2019-05-02
Implementation of comparison operation

hanyunx:
Not ready for merge.


amiller:
This is a really nice program... it's for "full field" comparison, one of the more complicated MPC operations.  This is in comparison to compare the "Fixpoint" comparison operations, which are a bit simpler but only work for a narrow range (i.e., 64 bit numbers). It's often useful to have comparison for the full field though, for example to perform a range check on untrusted shares.

The tests includes comparisons of random elements from a range of cross-over points, like near `p//2` and `p//4`... since some of the algorithm has special cases to handle when the leading couple bits of the numbers are 0.

amiller (changes requested):
I am excited for this comparison algorithm, it is a really interesting stand alone application since it's such a tricky algorithm. It should be moved into `progs`, and the documentation could be expanded a bit. These remaining changes are for a maintainer to take up. It's a great from yuguang and hanyun to get this to work.

Drake-Eidukas:
@amiller Once this gets merged I can move it into progs.

Drake-Eidukas:
https://github.com/initc3/HoneyBadgerMPC/pull/302 This work is continued here.

PR #286, opened 2019-05-03
Batches in HBAVSS-lite

tyurek:
allows the use of hbavss-lite with arbitrarily large batches.


nqyy:
What's the point of this given we already have the batch hbavss?

amiller:
We need this for comparison since we are not sure for what concrete values of batch size and n the batch version is concretely faster

nqyy:
LGTM.

Drake-Eidukas:
Restarted jenkins build-- not sure why it's been hanging on step 19...

nqyy (approved):

Drake-Eidukas:
https://github.com/initc3/HoneyBadgerMPC/pull/287

Issue #288, opened 2019-05-04
Investigate failing travis builds


Drake-Eidukas:
Currently, sometimes travis will fail on step 19 by failing to make progress. When restarted, it works... normally.
I will attach a log file next time I see the issue.

Drake-Eidukas:
[log.txt](https://github.com/initc3/HoneyBadgerMPC/files/3144406/log.txt)

PR #289, opened 2019-05-05
Removed docs from travis

Drake-Eidukas:
Currently, travis builds are failing. While investigation / improvements of docker/travis occur, this is a stopgap so that our builds will pass. This is part of the effort to address #288


smkuls (approved):
Can you add a comment in the code as well so as to why you are commenting this?

Drake-Eidukas:
now tests fail to build instead of docs, so I don't think this will work as a stopgap even.

Issue #239, opened 2019-04-09
Improve Travis Build Times


Drake-Eidukas:
We should be able to get some quick improvements if we take @smkuls's approach of having a base docker image with dependencies installed, and then basing our travis builds off of that.

Currently, our builds take around 11-12 minutes to run (22-24 minute total time).

Drake-Eidukas:
Currently, builds are down to 8 minutes including testing time. For docs and flake8, we see 2 minutes and 1 minute respectively.

Drake-Eidukas:
#310 Should bring this down to the actual minimal time to build on travis.

amiller:
Closing, just refer to #310 for now

PR #290, opened 2019-05-05
Refactoring travis and docker

Drake-Eidukas:
This PR does 2 things at once-- it fixes the current issue where Travis builds break, and also cuts the build times down to 1-2 min for docs and 8-9 min for tests.

We achieve this by leveraging multi-stage docker builds such that we can create an image with all of our dependencies installed, and use it in later stages to build on top of it. Thus, much of the work of the final stage of building the docker image is just downloading the image. 

However, now we have an added task-- when we install new dependencies, we need to push to dockerhub, which requires someone to build the image locally. If I find anything quick that can avoid that I'll add it in.

This tackles https://github.com/initc3/HoneyBadgerMPC/issues/288, as well as https://github.com/initc3/HoneyBadgerMPC/issues/239

### DIFF
--- a/.travis.Dockerfile
+++ b/.travis.Dockerfile
@@ -1,0 +1,8 @@
+FROM dsluiuc/honeybadgermpc-docker-base
+
+WORKDIR /usr/src/HoneyBadgerMPC
+
+COPY . /usr/src/HoneyBadgerMPC
+
+RUN pip install -e .
+RUN make -C apps/shuffle/cpp -j
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,64 +1,51 @@
-FROM python:3.7.1-stretch
+# Dockerfile contains multiple levels of images--
+# First, a base image containing dependencies shared by all other images:
+#   - apt dependencies
+#   - rust
+#   - ntl
+#   - pbc
+#   - charm
+#   - base pip dependencies (cython & setup.py)
+# 
+# Thereafter, it adds ever increasing levels of dependencies-- 
+#   - Test requirements (including doc requirements)
+#   - Dev requirements (including aws)
+#
+# In order to build and push this to dockerhub, run:
+# docker build . --target test-image --tag dsluiuc/honeybadgermpc-docker-base
+# docker push dsluiuc/honeybadgermpc-docker-base:latest
+FROM python:3.7.3-slim AS base-image
 
 # Allows for log messages to be immediately dumped to the 
 # stream instead of being buffered.
 ENV PYTHONUNBUFFERED    1 
 
-# Adds cargo to path for rust dependencies
-ENV PATH                /root/.cargo/bin:$PATH
-
 # Path variables needed for Charm
 ENV LIBRARY_PATH        /usr/local/lib
 ENV LD_LIBRARY_PATH     /usr/local/lib
-
-# Downloads rust and sets it up
-RUN curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain nightly-2018-10-24
-
-# Install apt dependencies
-RUN apt-get update && apt-get install -y \
-    vim \
-    tmux \
-    bison \
-    flex \
-    libgmp-dev \
-    libmpc-dev \
-    libmpfr-dev \
-    libflint-dev
-# Download and build NTL from source
-# Shoup recommends not using O3
-RUN wget -qO- https://www.shoup.net/ntl/ntl-11.3.2.tar.gz | tar xzvf - \
-         &&  cd ntl-11.3.2/src \
-         && ./configure CXXFLAGS="-g -O2 -fPIC -march=native -pthread -std=c++11" \
-         && make -j4 \
-         && make install
-
-# Sets default directory for running the rest of the commands
-WORKDIR /usr/src/HoneyBadgerMPC
-
-RUN pip install --upgrade pip
-
-COPY . /usr/src/HoneyBadgerMPC
-
-RUN pip install pairing/
-
-RUN wget -qO- https://crypto.stanford.edu/pbc/files/pbc-0.5.14.tar.gz | tar xzvf - \
-    && cd pbc-0.5.14 \
-    && ./configure \
-    && make \
-    && make install
-
 
 # Make sh point to bash
 # This is being changed since it will avoid any errors in the `launch_mpc.sh` script
 # which relies on certain code that doesn't work in container's default shell.
 RUN ln -sf bash /bin/sh
 
-# Downloads and installs charm
-RUN git clone https://github.com/JHUISI/charm.git \
-    && cd charm \
-    && git reset --hard be9587ccdd4d61c591fb50728ebf2a4690a2064f \
-    && ./configure.sh \
-    && make install
+# Install apt dependencies
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    bison \
+    curl \
+    flex \
+    g++ \
+    git \
+    libflint-dev \
+    libgmp-dev \
+    libmpc-dev \
+    libmpfr-dev \
+    libssl-dev \
+    make \
+    openssl \
+    tmux \
+    wget \
+    vim 
 
 # This is needed otherwise the build for the power sum solver will fail.
 # This is a known issue in the version of libflint-dev in apt.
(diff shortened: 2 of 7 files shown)


amiller:
OK this is fine as long as we have some clear documentation somewhere about what steps are needed to update the build, what account it is under, etc. No idea who all has privs to push updates to dockerhub/dsluiuc

tests/fixtures.py
```diff
@@ -171,3 +171,7 @@ def test_runner(test_preprocessing):
         return await program_runner.join()
 
     return _test_runner
+
+
+if '__main__' == __name__:
+    print('main')
```
amiller: looks like leftover debugging info?

tests/fixtures.py
```diff
@@ -1,7 +1,7 @@
-from honeybadgermpc.router import SimpleRouter
-import asyncio
-import random
 from pytest import fixture
+import random
+import asyncio
+from honeybadgermpc.router import SimpleRouter
```
amiller: any reason for these changes?

setup.py
```diff
@@ -65,17 +64,19 @@
...
 if not VERSION:
+    g = {}
+
+    # TODO: consolidate how we do this
     with open(os.path.join(here, NAME, '__version__.py')) as f:
-        exec(f.read(), about)
-else:
-    about['__version__'] = VERSION
+        exec(f.read(), g)
+        VERSION = g['__version__']
```
amiller, line 78: is this really how to do this? `exec` statements make me anxious

Dockerfile
```diff
@@ -1,78 +1,138 @@
...
+#   - pbc
+#   - charm
+#   - base pip dependencies (cython & setup.py)
+# 
+# Thereafter, it add ever increasing levels of dependencies-- 
+#   - Test requirements (including doc requirements)
+#   - Dev requirements (including aws)
+#
+# In order to build and push this to dockerhub, run:
+# docker build . --target test-image --tag dsluiuc/honeybadgermpc-docker-base
```
amiller, line 15: OK I see here is where the comments are..... I guess the Dockerfile is a natural enough place, but it should also be in a readme and/or wiki and/or the developer guide documentation

.travis.Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM dsluiuc/honeybadgermpc-docker-base
```
amiller, line 1: okay so this is essentially only for travis to use right? We don't use `honeybadgermpc-docker-base` in the dev image itself

Dockerfile
```diff
@@ -1,78 +1,138 @@
...
-    && make install
 
 
-# Make sh point to bash
-# This is being changed since it will avoid any errors in the `launch_mpc.sh` script
-# which relies on certain code that doesn't work in container's default shell.
-RUN ln -sf bash /bin/sh
+# Installs test dependencies
+# For now, upload this to docker-hub
+FROM base-image AS test-image
```
amiller: I am not familiar with this Dockerfile syntax. How do multiple `FROM` statements work? Do you really just get different images at the end?

amiller: https://docs.docker.com/develop/develop-images/multistage-build/ 
OK this documentation does clearly describe what's going on.

Dockerfile
amiller: It could be nice to add these statements to a script, like `scripts/build-and-push-to-dockerhub.sh` that also includes this documentation

setup.py
amiller: I guess it's fine

amiller (changes requested):
let's remove the (apparently?) unnecessary changes to fixtures.py and go for it

.travis.Dockerfile
smkuls: This change is not development friendly, if I modify the Dockerfile in `honeybadgermpc` then it requires me to modify this file as well (which I might not even know that I have to do).

We had earlier made an effort to unify our Dockerfiles (one file for both Travis and dev). Can we keep it that way?

Dockerfile
```diff
@@ -1,78 +1,138 @@
...
+# Dockerfile contains multiple levels of images--
+# First, a base image containing dependencies shared by all other images:
+#   - apt dependencies
+#   - rust
+#   - ntl
+#   - pbc
+#   - charm
+#   - base pip dependencies (cython & setup.py)
+# 
+# Thereafter, it add ever increasing levels of dependencies--
```
smkuls: nit: typo

Dockerfile
```diff
@@ -1,78 +1,138 @@
...
+# Thereafter, it add ever increasing levels of dependencies-- 
+#   - Test requirements (including doc requirements)
+#   - Dev requirements (including aws)
+#
+# In order to build and push this to dockerhub, run:
+# docker build . --target test-image --tag dsluiuc/honeybadgermpc-docker-base
+# docker push dsluiuc/honeybadgermpc-docker-base:latest
+
+
+FROM python:3.7.3-slim AS base-image
```
smkuls, line 17: Neat! What's the size of the new image?

Dockerfile
```diff
@@ -1,78 +1,138 @@
-FROM python:3.7.1-stretch
+# Dockerfile contains multiple levels of images--
```
smkuls, line 1: Is the `base-image` cached then? I mean if I run this on my local machine and modify a layer after the base image has been created will it skip the creation of the base image?

Dockerfile
```diff
@@ -1,78 +1,138 @@
...
+WORKDIR /
+
+
+# Below derived from https://pythonspeed.com/articles/multi-stage-docker-python/
+WORKDIR /usr/src/HoneyBadgerMPC
 COPY . /usr/src/HoneyBadgerMPC
 
+RUN pip install --upgrade pip
+RUN pip install Cython
+RUN pip install -e .
```
smkuls, line 99: While you are at it, can you look into `requirements.txt` or some other way in which we can skip downloading the packages for every call to `docker-compose build`?

You can take it up in a separate PR if you want.

Dockerfile
```diff
@@ -1,78 +1,138 @@
...
+
 # Download and build NTL from source
 # Shoup recommends not using O3
-RUN wget -qO- https://www.shoup.net/ntl/ntl-11.3.2.tar.gz | tar xzvf - \
-         &&  cd ntl-11.3.2/src \
-         && ./configure CXXFLAGS="-g -O2 -fPIC -march=native -pthread -std=c++11" \
-         && make -j4 \
-         && make install
+RUN curl -so - https://www.shoup.net/ntl/ntl-11.3.2.tar.gz | tar xzvf -
+WORKDIR /ntl-11.3.2/src
```
smkuls, line 68: Doesn't each line act as a new layer? Why are we going back to what we had earlier?

Dockerfile
```diff
@@ -1,78 +1,138 @@
...
-# This has been fixed if we pull the latest code from the repo. However, we want
-# to avoid compiling the lib from the source since it adds 20 minutes to the build.
-RUN sed -i '30c #include "flint/flint.h"' /usr/include/flint/flintxx/flint_classes.h
 
-ARG BUILD
-RUN pip install Cython
 
-# Runs setup.py
-RUN pip install --no-cache-dir -e .[$BUILD]
+# TODO: see if we can shrink this image size more.
```
smkuls, line 107: Can you get rid of commented code?

smkuls (changes requested):

Dockerfile
Drake-Eidukas: It starts really small... then when everything else gets installed to get the tests running, we sit at a bit under 2GiB

Dockerfile
Drake-Eidukas: Yep! Unless you run with `--no-cache`

Dockerfile
Drake-Eidukas: Sure-- I think it'll be better.

Dockerfile
Drake-Eidukas: So, mostly this is more development friendly. I've been going through the docs, and having many layers isn't really problematic--

1. It can cache up until the first thing that invalidates the cache, so builds are faster
2. Modern docker isn't impacted much from it performance / memory wise.
3. Easier to debug the dockerfile.

tests/fixtures.py
Drake-Eidukas: I was debugging an issue that came about importing ntl code, must have accidentally checked in this file

setup.py
Drake-Eidukas: I'm not sure this is the right way to do this, but this is how the existing setup was. I think it makes sense to just put the version tag in this file.

.travis.Dockerfile
Drake-Eidukas: Let's address this in the next PR.

Dockerfile
Drake-Eidukas: Will document this more thoroughly tonight, let's focus on getting this merged first.

Dockerfile
```diff
@@ -1,78 +1,129 @@
...
+# 
+# Thereafter, it adds ever increasing levels of dependencies-- 
+#   - Test requirements (including doc requirements)
+#   - Dev requirements (including aws)
+#
+# In order to build and push this to dockerhub, run:
+# docker build . --target test-image --tag dsluiuc/honeybadgermpc-docker-base
+# docker push dsluiuc/honeybadgermpc-docker-base:latest
+
+FROM dsluiuc/honeybadgermpc-docker-base AS travis-image
```
smkuls: I know you said you will take it up later so just mentioning -> we should not have any reference to Travis here - not even in the name.

smkuls (approved):
LGTM! Since this has become critical lets get this in and we can take up comments in the following PRs.

PR #291, opened 2019-05-06
AVSS work in process mode

hybridNeo:
AVSS works in process mode. Future changes will add flag for type of AVSS.


scripts/launch.sh
```diff
@@ -35,4 +35,5 @@ tmux new-session     "${CMD} -d -f ${CONFIG_PATH}.0.json; sh" \; \
      splitw -h -p 50 "${CMD} -d -f ${CONFIG_PATH}.1.json; sh" \; \
      splitw -v -p 50 "${CMD} -d -f ${CONFIG_PATH}.2.json; sh" \; \
      selectp -t 0 \; \
-     splitw -v -p 50 "${CMD} -d -f ${CONFIG_PATH}.3.json; sh"
+     splitw -v -p 50 "${CMD} -d -f ${CONFIG_PATH}.3.json; sh" \; \
```
smkuls: This and the config change will break the other existing scripts. Can you create separate configs for AVSS? Can you also make this script generic by having default number of panes as 4 unless passed from the command line?

honeybadgermpc/protocols/avid.py
```diff
@@ -108,7 +108,7 @@ def broadcast(self, o):
 
                     return decoded_output
 
-    async def disperse(self, sid, pid, input_list):
+    async def disperse(self, sid, pid, input_list, client_mode=False):
```
smkuls, line 111: Update the docstring to say what a client mode is.

Basically in the client mode, the dealer just sends the first message to all the receivers and returns.

honeybadgermpc/poly_commit_const.py
```diff
@@ -68,13 +68,13 @@ def gen_pc_const_crs(t, alpha=None, g=None, h=None, ghat=None):
     assert type(h) in (G1, nonetype)
     assert type(ghat) in (G2, nonetype)
     if alpha is None:
-        alpha = ZR.random()
+        alpha = ZR.random(0)
```
smkuls, line 88: These shouldn't be by default hard coded. The seeds should be set from the command line.

honeybadgermpc/hbavss.py
```diff
@@ -538,22 +541,71 @@ def _get_dealer_msg(self, values, n):
...
             return
 
+        # start disperse in the background
+        self.avid_msg_queue.put_nowait((avid, tag, dispersal_msg_list))
+
         # avss processing
         await self._process_avss_msg(avss_id, dealer_id, rbc_msg, avid)
+
+
+def get_avss_params(n, t):
```
smkuls, line 592: Do not duplicate code, modify the existing method to take a seed.

honeybadgermpc/hbavss.py
```diff
@@ -538,22 +541,71 @@ def _get_dealer_msg(self, values, n):
...
+
+async def _run(peers, n, t, my_id):
+    g, h, pks, sks = get_avss_params(n + 1, t)
+    async with ProcessProgramRunner(peers, n+1, t, my_id) as runner:
+        send, recv = runner.get_send_recv('1')
+        crs = gen_pc_const_crs(t, g=g, h=h)
+        values = None
+        dealer_id = n
+        if my_id == n:
+            # Dealer
```
smkuls: You can remove this comment. The log makes it clear.

honeybadgermpc/hbavss.py
```diff
@@ -538,22 +541,71 @@ def _get_dealer_msg(self, values, n):
...
+        private_keys[i] = ZR.random(0)
+        public_keys[i] = pow(g, private_keys[i])
+    return g, h, public_keys, private_keys
+
+
+async def _run(peers, n, t, my_id):
+    g, h, pks, sks = get_avss_params(n + 1, t)
+    async with ProcessProgramRunner(peers, n+1, t, my_id) as runner:
+        send, recv = runner.get_send_recv('1')
+        crs = gen_pc_const_crs(t, g=g, h=h)
```
smkuls: move this till `592` out of the `with`. Basically start with `HbAVSS` directly in the `with` (no need to unnecessarily setup network connection if we don't want to use them right away.

honeybadgermpc/hbavss.py
```diff
@@ -538,22 +541,71 @@ def _get_dealer_msg(self, values, n):
...
+
         logger.debug("[%d] Starting AVID disperse", self.my_id)
         avid = AVID(n, self.t, dealer_id, recv, send, n)
-        # start disperse in the background
-        self.avid_msg_queue.put_nowait((avid, tag, dispersal_msg_list))
 
         if client_mode and self.my_id == dealer_id:
             # In client_mode, the dealer is not supposed to do
             # anything after sending the initial value.
+            await avid.disperse(tag, self.my_id, dispersal_msg_list, client_mode=True)
```
smkuls, line 581: Update the comment to say that the dealer will send the initial message and then can terminate.

honeybadgermpc/hbavss.py
```diff
@@ -17,7 +20,7 @@
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.ERROR)
 # Uncomment this when you want logs from this file.
-# logger.setLevel(logging.NOTSET)
+logger.setLevel(logging.NOTSET)
```
smkuls, line 21: Comment this again and change the `logger.info` in main to a print.

smkuls (changes requested):

honeybadgermpc/hbavss.py
```diff
@@ -538,22 +541,71 @@ def _get_dealer_msg(self, values, n):
             # dispersal_msg_list: the list of payload z
             broadcast_msg, dispersal_msg_list = self._get_dealer_msg(values, n)
 
-        tag = f"{dealer_id}-{avss_id}-B-AVID"
+        tag = f"{dealer_id}-{avss_id}-B-RBC"
```
nqyy, line 565: Nice catch. I should have updated this.

honeybadgermpc/protocols/avid.py
```diff
@@ -144,6 +144,8 @@ def broadcast(self, o):
                 # send each person the column of stripes
                 self.send(i, (sid, AVIDMessageType.VAL, roothash_list,
                               branch_list, stripes_list_per_party[i]))
+            if client_mode:
```
nqyy, line 147: Add a comment for this.

amiller:
needs flake8 fix

smkuls (approved):

honeybadgermpc/hbavss_light.py
```diff
@@ -0,0 +1,54 @@
...
+from .hbavss import get_avss_params, HbAvssLight
+from honeybadgermpc.betterpairing import ZR
+import asyncio
+import time
+import logging
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.ERROR)
+# Uncomment this when you want logs from this file.
+logger.setLevel(logging.NOTSET)
```
smkuls, line 12: You don't need this and the two lines before this.

smkuls: Same in the other file

honeybadgermpc/hbavss_batch.py
```diff
@@ -0,0 +1,56 @@
...
+    async with ProcessProgramRunner(peers, n+1, t, my_id) as runner:
+        send, recv = runner.get_send_recv('HBAVSS_BATCH')
+        crs = gen_pc_const_crs(t, g=g, h=h)
+        values = None
+        dealer_id = n
+        if my_id == dealer_id:
+            # Dealer
+            values = [ZR.random(0)] * batch_size
+            logger.info("Starting DEALER")
+            logger.info(f"Dealer timestamp: {time.time()}")
```
smkuls, line 27: You should use `%s` to log and not format strings

smkuls (commented):
Can you open an issue so that we fix all the comments on this PR? This PR makes changes which can affect other protocols [hard coding the seed for polynomial commitments] and is also not in a great shape overall.

PR #292, opened 2019-05-07
Run HyperMPC random generation in process mode

smkuls:


Drake-Eidukas (approved):
I think adding some more documentation to hyper_invertible would be a good idea, otherwise LGTM

smkuls:
> I think adding some more documentation to hyper_invertible would be a good idea, otherwise LGTM

The algo is commented well, this is just wrapper code.

PR #295, opened 2019-05-08
Reimplement Triple Refinement

smkuls:
* Reimplement Triple Refinement using Vandermonde matrix
* Do interpolations and evaluations only at the required number of
points
* Update the tests to reflect the same


amiller (approved):
OK This is fine and i recognize it from thesis time

PR #301, opened 2019-05-08
Refactor and document type checking

Drake-Eidukas:
This does 2 main things:
- Refactors type checking to be more readable and maintainable
- Makes type checking assertion-based, which means that when we run with `python -O`, type checking is not run.


amiller (approved):
This looks really useful. Does this idea come from anywhere in particular? This approach to python optional runtime checking seems quite generally useful. This could make sense to put online as a separate library outside hbmpc altogether.

Nit: StaticTypeCheck seems to imply static type checking, i.e. type checking analysis done prior to runtime, but actually it seems like this is all at runtime

Drake-Eidukas:
Part of my intern project last summer was to work on [something along these lines for ruby](https://sorbet.org/), so that's where the idea came from. I agree it could be a decent library idea  for python (although it's somewhat already a thing with mypy and python type hints). 

The reason I have it named `StaticTypeCheck` is because it's used for "static" methods (`@staticmethod`), but I think I might be able to do some other introspection to consolidate this into one decorator.

Drake-Eidukas:
Alright, so I still need to port this over where we use typechecking in the codebase, but I just added the *new and improved* typechecking! It does all of the things it used to do, but now also checks the return type, and has a far less clunky interface. You plop the decorator down on a function, and now it performs optional typechecking based on the function signature and associated annotations, which is more in line with what developers will expect.

Drake-Eidukas:
This is completely ported over now with unit tests for the new functionality, and the rest of the dataflow functions now have annotations.

In single threaded mode, with typechecking on, it took `425.87s` to run all of the unit tests sequentially.
With typechecking off, it took `411.17s`. You can turn off typechecking as if run with `-O` by setting the environment variable `DISABLE_TYPECHECKING`

PR #302, opened 2019-05-09
Implementation of comparison algorithm

Drake-Eidukas:
This contains a stripped down and refactored version of comparison. 
Currently, this algorithm fails every so often (about every other time I run the tests, which run 18 comparisons each time). I assume the pre-existing assertions check for conditions that cause this to fail, so i can re-add those if desired.


Drake-Eidukas:
The only difference to the algorithm other than refactoring names / style is that there was [this line](https://github.com/initc3/HoneyBadgerMPC/pull/282/files#diff-d1ea8ffaf5118b3cd6de8b67906d4b7aR74) in the previous algorithm, which goes against the algorithm in the paper. With or without that line, there are still intermittent failures.

Drake-Eidukas:
(this is a continuation of https://github.com/initc3/HoneyBadgerMPC/pull/282)

Drake-Eidukas:
> The only difference to the algorithm other than refactoring names / style is that there was [this line](https://github.com/initc3/HoneyBadgerMPC/pull/282/files#diff-d1ea8ffaf5118b3cd6de8b67906d4b7aR74) in the previous algorithm, which goes against the algorithm in the paper. With or without that line, there are still intermittent failures.

After further investigation, that line was indeed necessary.

Drake-Eidukas:
Tests now pass in travis... but fail every once in a while. It's pretty consistent now that I've fixed the algorithm.

amiller:
What's the status of this? If I understand it correctly, the tests sometimes still fail? This seems to indicate a flaw in the implementation that only occurs for some values to compare and for some random choices of bits. Can we investigate this further by finding a set of random seeds / values that reproduces the problem?

Drake-Eidukas:
I haven't been able to get that to happen consistently-- it's an issue with share reconstruction. I left comments with values that have failed before, but then when I hard-code those numbers, it passes... I'll take another stab at it.

Drake-Eidukas:
At LONG LAST, comparison should be done. We no longer leak information, and this passed all 250 times I've run it so far locally. Once reviewed, I'll squash the commits. This does a couple of things in addition to simply re-implementing comparison--
1. Moves `Equality` into `share_comparison.py` alongside `LessThan`, and adds preprocessing for generating what I call `share_bits`, which is simply a share with it's bitwise representation in shares. 
2. This also slightly changes the dockerfile, as i had to remove the `-j` to get this to build locally, and I know @hanyunxu and @rahulgovind both have had issues related to that flag as well (seems to mostly affect macs?). 

This operation is somewhat slow, and when testing this is compounded by the fact that since we do 18 comparisons, we need 10,000 preprocessing elements, as well as it just takes a while to run the tests. That said, on my machine, all of the tests finish in about 3 minutes when run with 8-way parallelism, which just about double what it used to be.

Drake-Eidukas:
Seems like the dockerfile for travis wasn't updated for asynchromix-- this is also updated by this PR

Drake-Eidukas:
So, I had to up `n` to 4 to get this to stop hanging on travis, but I suspect the issue was a `StopIteration` exception getting eaten up.

honeybadgermpc/preprocessing.py
```diff
@@ -4,6 +4,7 @@
 from uuid import uuid4
 from random import randint
 from os import makedirs
+from gmpy2 import num_digits
```
amiller: This is an overkill dependency just to use `num_digits` given that we are otherwise only using an int

honeybadgermpc/preprocessing.py
```diff
@@ -22,17 +23,20 @@ class PreProcessingConstants(object):
     ONE_MINUS_ONE_FILE_NAME_PREFIX = f"{SHARED_DATA_DIR}one_minus_one"
     DOUBLE_SHARES_FILE_NAME_PREFIX = f"{SHARED_DATA_DIR}double_shares"
     READY_FILE_NAME = f"{SHARED_DATA_DIR}READY"
+    SHARE_BITS_FILE_NAME_PREFIX = f"{SHARED_DATA_DIR}share_bits"
```
amiller: `ONE_MINUS_ONE` is also really the same as `BIT` except for a small local operation, I think we should only have one or the other

amiller (approved):
This looks great. Very nice refactor and bugfixing

honeybadgermpc/preprocessing.py
Drake-Eidukas: Good call-- TIL `int` has a `bit_length()` method already.

PR #305, opened 2019-05-10
Avid fix

amiller:
AVID currently fails if a RETRIEVE request is received before OK is finished. A solution is to enqueue retrieval requests.


tyurek:
Looks good to me

tyurek (approved):

PR #306, opened 2019-05-11
Localbenchmarks

tyurek:
adds a script that can do local benchmarking for avss. Allows for precomputation with polycommits. Also adds earlier termination for rbc with client mode

### DIFF
--- a/benchmark/test_benchmark_hbavss.py
+++ b/benchmark/test_benchmark_hbavss.py
@@ -1,0 +1,167 @@
+from pytest import mark
+from contextlib import ExitStack
+from random import randint
+from honeybadgermpc.poly_commit_const import gen_pc_const_crs, PolyCommitConst
+from honeybadgermpc.poly_commit_lin import PolyCommitLin
+from honeybadgermpc.betterpairing import G1, ZR
+from honeybadgermpc.hbavss import HbAvssLight, HbAvssBatch
+from honeybadgermpc.field import GF
+from honeybadgermpc.elliptic_curve import Subgroup
+import asyncio
+
+
+def get_avss_params(n, t):
+    g, h = G1.rand(), G1.rand()
+    public_keys, private_keys = [None] * n, [None] * n
+    for i in range(n):
+        private_keys[i] = ZR.random()
+        public_keys[i] = pow(g, private_keys[i])
+    return g, h, public_keys, private_keys
+
+
+@mark.parametrize("t, k", [(1, 5), (3, 5), (5, 5), (16, 5), (33, 5), (85, 5),
+                           (1, 25), (3, 25), (5, 25), (16, 25), (33, 25), (85, 25),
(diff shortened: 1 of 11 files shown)


tests/test_hbavss.py
```diff
@@ -603,6 +647,12 @@ def _get_dealer_msg(self, values, n):
 
 @mark.asyncio
 async def test_hbavss_light_client_mode(test_router):
+    def callback(future):
```
amiller: We can now import this as:
`from honeybadgermpc.utils.misc import print_exception_callback`

pairing/src/lib.rs
```diff
@@ -611,13 +611,15 @@ impl PyFr {
     //    })
     //}
     //fn __new__(obj: &PyRawObject, s1: u32, s2: u32, s3: u32, s4: u32) -> PyResult<()>{
-    fn __new__(obj: &PyRawObject, s: &str) -> PyResult<()>{
-        let f =  Fr::from_str(s).unwrap();
```
amiller, line 615: let's get rid of the commented out code

benchmark/test_benchmark_hbavss.py
```diff
@@ -0,0 +1,167 @@
...
+
+    (t, n, g, h, pks, sks, crs, pc, values, field) = params
+    sends, recvs, _ = test_router(n+1)
+    dealer_id = n
+
+    hbavss = HbAvssLight(pks, None, crs, n, t, dealer_id, sends[dealer_id], recvs[dealer_id], pc=pc, field=field)  # (# noqa: E501)
+    await asyncio.create_task(hbavss.avss(0, value=values, client_mode=True))
+
+
+async def hbavss_batch_batch(test_router, params):
```
amiller: Can't help but think `batch_batch` sounds redundant. Let's consider an alternate name for a future commit

amiller (changes requested):
This should ideally be two commits. One is about adding benchmarks, the other is a bunch of changes/improvements to the AVSS implementation.

honeybadgermpc/betterpairing.py
```diff
@@ -111,7 +111,7 @@ def __pow__(self, other):
             try:
                 intother = int(other)
                 exponend = ZR(intother)
-            except:
+            except Exception:
```
amiller: This is too broad - let's figure out what a type error would throw and just catch that

tests/test_hbavss.py
tyurek: oh cool. that's a good change

benchmark/test_benchmark_hbavss.py
tyurek: going with multibatch for now

amiller (approved):
This looks ready to go. Approve +1

Drake-Eidukas:
Luckily there's no merge conflicts. I'll update travis.

PR #307, opened 2019-05-13
Batch implementation of RanDouSha

amiller:
This is a batched version of RanDouSha (hyper-invertible.py). It now takes an additional parameter `k`, and outputs `(n-2t)k` random double sharings. By using the batch version of `decode`/`encode` and sending only one message for the batch it can avoid the startup cost of the inverse vandermonde matrix, etc., reduce the number of messages sent, python function calls, etc.

### DIFF
--- a/tests/test_hyper_invertible.py
+++ b/tests/test_hyper_invertible.py
@@ -7,14 +7,15 @@ from honeybadgermpc.reed_solomon import Algorithm, DecoderFactory
 
 @mark.asyncio
 @mark.parametrize("n", [4, 7])
-async def test_generate_double_shares(test_router, polynomial, galois_field, n):
+@mark.parametrize("k", [1, 10])
+async def test_generate_double_shares(test_router, polynomial, galois_field, n, k):
     t = (n-1)//3
     sends, receives, _ = test_router(n)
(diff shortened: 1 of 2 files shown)


Drake-Eidukas (approved):
I don't know much about the actual math to comment on that, but stylistically this is fine.

tests/test_hyper_invertible.py
```diff
@@ -7,20 +7,21 @@
...
     assert len(shares_per_party) == n
-    assert all(len(random_shares) == n-2*t for random_shares in shares_per_party)
-    random_values = [None]*(n-2*t)
+    assert all(len(random_shares) == (n-2*t)*k for random_shares in shares_per_party)
+    random_values = []
     eval_point = EvalPoint(galois_field, n, use_fft=True)
     decoder = DecoderFactory.get(eval_point, Algorithm.FFT)
     for i, shares in enumerate(zip(*shares_per_party)):
         shares_t, shares_2t = zip(*shares)
         r_t = polynomial(decoder.decode(list(range(n)), shares_t))(0)
```
rahulgovind, line 23: Is this required (`
        r_t = polynomial(decoder.decode(list(range(n)), shares_t))(0`)? This looks like a fairly expensive operation.

Issue #297, opened 2019-05-08
Further refactor Travis builds / docker setup.


Drake-Eidukas:
Investigate means of consolidating docker files. The optimal way is to pull down a docker image, mount code, then just run tests. Furthermore, this will require updating the way that we track dependencies.

Drake-Eidukas:
First refactor https://github.com/initc3/HoneyBadgerMPC/pull/300

Drake-Eidukas:
Mostly finished by #351

PR #310, opened 2019-05-16
Decrease docker image size and speedups for travis

Drake-Eidukas:
Currently, the docker image we use for testing is ~2.5 GB. This cuts the size down to ~550 MB, and also makes it so that we don't have to install dependencies on travis. This allows our CI/CD pipeline to finish around 7-8 minutes. However, this will require some manual processes when we update our dependencies-- someone will have to re-run the script to build and push the docker image to dockerhub.

### DIFF
--- a/.ci/Dockerfile
+++ b/.ci/Dockerfile
@@ -1,8 +1,3 @@
-FROM dsluiuc/honeybadgermpc-docker-base AS travis-image
+FROM dsluiuc/honeybadgermpc-test-image
 
-WORKDIR /usr/src/HoneyBadgerMPC
 COPY . /usr/src/HoneyBadgerMPC
-
-RUN pip install -e .["tests,docs"]
-RUN make -C apps/shuffle/cpp -j
-
--- a/scripts/build_dockerhub_image.sh
+++ b/scripts/build_dockerhub_image.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+# This script is used to create the docker image we push to dockerhub.
+# This requires that you have access to the dsluiuc dockerhub page. 
+# Currently, only Drake and Samarth have access to this.
+# Usage: sh scripts/build_dockerhub_image.sh
+
+docker login
+docker build . --target test-image --tag dsluiuc/honeybadgermpc-test-image
+docker push dsluiuc/honeybadgermpc-test-image:latest
--- a/setup.py
+++ b/setup.py
@@ -13,6 +13,7 @@ REQUIRES_PYTHON = '>=3.7.0'
 VERSION = None
 
 REQUIRED = [
+    'cython',
     'gmpy2',
     'zfec',
     'pycrypto',
@@ -21,10 +22,9 @@ REQUIRED = [
     'pyzmq',
 ]
 
-TESTS_REQUIRES = [
-    'flake8',
+
+TEST_REQUIRES = [
     'pep8-naming',
-    'pytest',
     'pytest-asyncio',
     'pytest-mock',
     'pytest-cov',
@@ -32,33 +32,36 @@ TESTS_REQUIRES = [
     'pytest-xdist',
     'pytest-benchmark',
     'pytest-benchmark[histogram]',
-    'pyyaml',
 ]
 
 DEV_REQUIRES = [
     'ipdb',
-    'ipython',
 ]
 
-DOCS_REQUIRE = [
-    'Sphinx',
+DOCS_REQUIRES = [
     'sphinx-autobuild',
     'sphinx_rtd_theme',
     'sphinx_tabs',
     'm2r',
-    'doc8',
+    'doc8'
 ]
 
-ETH_REQUIRES = ['web3', 'ethereum']
+ETH_REQUIRES = [
+    'web3',
+    'ethereum',
+]
 
-AWS_REQUIRES = ['boto3', 'paramiko']
+AWS_REQUIRES = [
+    'boto3',
+    'paramiko',
+]
 
 EXTRAS = {
-    'tests': TESTS_REQUIRES,
-    'dev': DEV_REQUIRES + TESTS_REQUIRES + DOCS_REQUIRE + ETH_REQUIRES,
-    'docs': DOCS_REQUIRE + ETH_REQUIRES,
+    'tests': TEST_REQUIRES,
+    'dev': DEV_REQUIRES + TEST_REQUIRES + DOCS_REQUIRES + ETH_REQUIRES,
+    'docs': DOCS_REQUIRES + ETH_REQUIRES,
     'eth': ETH_REQUIRES,
-    'aws': AWS_REQUIRES,
+    'aws': AWS_REQUIRES
 }
 
 here = os.path.abspath(os.path.dirname(__file__))
@@ -99,7 +102,10 @@ setup(
     long_description=long_description,
     long_description_content_type='text/markdown',
     python_requires=REQUIRES_PYTHON,
-    setup_requires=['cffi>=1.0.0', 'Cython'],
+    setup_requires=[
+        'cffi>=1.0.0',
+        'cython'
+    ],
     install_requires=REQUIRED,
     cffi_modules=['apps/shuffle/solver/solver_build.py:ffibuilder'],
     extras_require=EXTRAS,
--- a/.travis.Dockerfile
+++ b/.travis.Dockerfile
@@ -1,8 +1,0 @@
-FROM dsluiuc/honeybadgermpc-docker-base
-
-WORKDIR /usr/src/HoneyBadgerMPC
-
-COPY . /usr/src/HoneyBadgerMPC
-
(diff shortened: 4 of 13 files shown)


Drake-Eidukas:
This is ready for review / merge now, I think. The commit message has a good description of what all is done by this, but essentially, this further trims the docker image that we push to dockerhub and adds/updates documentation about the changes. Additionally, I have charm as a git submodule now. Whether or not this is optimal is up for debate-- I think it makes sense as it feels more natural, however, it very slightly increases the build time of the docker image locally. That said, I think we can massively cut down this time using auto-builds through docker-hub, and using those builds for either caching, or simply just pulling those images for local use.

This tackles both #297 and #239.

Drake-Eidukas:
WOW dockerhub is slow... but at least it works! It takes 45 minutes to build the image.

amiller (approved):
This looks good to me, the changes make sense and from iteration I believe they have the intended effect. The updated `ci.rst` gives a good description of what's necessary for the workflow. So I think this looks good.

amiller:
I would like @sbellem to sign off on this as sane before merging it though, could you add a review?

.ci/Dockerfile
```diff
@@ -1,8 +1,3 @@
-FROM dsluiuc/honeybadgermpc-docker-base AS travis-image
-
-WORKDIR /usr/src/HoneyBadgerMPC
-COPY . /usr/src/HoneyBadgerMPC
-
-RUN pip install -e .["tests,docs"]
-RUN make -C apps/shuffle/cpp -j
+FROM dsluiuc/honeybadgermpc-test-image
```
smkuls, line 1: Is this what we want to call it?

setup.py
```diff
@@ -21,44 +22,46 @@
...
     'sphinx-autobuild',
     'sphinx_rtd_theme',
     'sphinx_tabs',
     'm2r',
-    'doc8',
+    'doc8'
 ]
 
-ETH_REQUIRES = ['web3', 'ethereum']
+ETH_REQUIRES = [
```
smkuls, line 49: @amiller do we have any code right now which uses these?

scripts/build_dockerhub_image.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+# This script is used to create the docker image we push to dockerhub.
+# This requires that you have access to the dsluiuc dockerhub page. 
+# Currently, only Drake and Samarth have access to this.
+# Usage: sh scripts/build_dockerhub_image.sh
+
+docker login
+docker build . --target test-image --tag dsluiuc/honeybadgermpc-test-image
+docker push dsluiuc/honeybadgermpc-test-image:latest
```
smkuls, line 10: I don't think this will work...

What happens when two people update the code and want to upload different images to the same endpoint? One of them might pick up the other version..

smkuls (changes requested):
I would like to discuss some things before this gets merged.

.ci/Dockerfile
Drake-Eidukas: I named it this to prevent a clash with the existing docker image. I think this problem ties in pretty well with your concerns about two people updating this.

smkuls (approved):

smkuls:
Approving this for now, @Drake-Eidukas and I discussed a bunch of things that need to be fixed and he is going to make issues for all of them and take them up when he gets time. He is also blocked on this to be checked in.

sbellem:
Great effort!

### tl;dr
I am concerned with the heavy reliance on dockerhub and would like that we try to reduce that reliance as much as possible by separating the dependencies into those that can be dealt well by travis by using [travis' caching mechanism](https://docs.travis-ci.com/user/caching/) from other dependencies that cannot be cached by travis. 

### Small recommendation
As I understood it the PR addresses a few concerns:

1. Image size
2. Slow travis builds -- pre-build dependencies on dockerhub
3. Style changes
4. Removal of explicit dependencies in `setup.py`

It can be a good exercise to try to limit a PR to one concern as it simplifies and speeds up the reviewing process.

### Image size
Just as a note: if the quest for a smaller image goes on, some work was done in the past to use [alpine](https://github.com/docker-library/python/blob/4baab390a3ba83908b869277491c4ce2ca283373/3.8-rc/alpine3.9/Dockerfile) which I think is probably the smallest base image one can use. Some dependencies needed for the python pairing lib required a dev version of Rust (at the time) and it made it tricky to use alpine, although still possible but it was decided to stick to simplicity as image size was not so much a concern. 

### Travis build time
I think we should try to categorize the dependencies here and leave some to travis given that travis has a [caching mechanism](https://docs.travis-ci.com/user/caching/).  At the very least we should try to see if the speed up is good enough -- especially for [`pip`](https://docs.travis-ci.com/user/caching/#pip-cache).

It seems to me that it would be preferable *not to* depend on dockerhub for the dependencies that are installed via `pip` because it seems to me that the need for a new dependency is not so infrequent in general and having to go through a manual process to build an entire new docker image on dockerhub seems to me more inconvenient in comparison to waiting for pip to install un-cached dependencies.

### Style changes
These are subjective and it re-enforces the motivation for using a code formatter tool: `[black](https://github.com/python/black)`. A PR should appear soon! :smile: 

### Removal of explicit dependencies in `setup.py`
Hmmmm ... I wonder ... the zen of python does say that "explicit is better than implicit" ... I don't mind either way here except that I see this habit as potentially problematic if in the future someone was to remove an explicit dependency that was the sole responsible for an implicit dependency. In other words say `pytest` is removed because `pytest-asyncio` installs it and then later on `pytest-asyncio` is no longer needed and removed but now `pytest` is missing. So perhaps the saying from the zen of python: "explicit is better than implicit" is relevant here ... :smile:

sbellem:
_draft_

@Drake-Eidukas nailed it! 

We’ll simply refine what he’s done! One proposal is to maintain a core/base image in a separate repository and possibly, if that makes any sense, to break down that core image into small specialized ones (e.g.: pairing, ntl, pbc, charm, etc). The Dockerfile(s) under the main repository, HoneyBadgerMPC will be building `FROM` that core image, just like it is done in this PR. The small tweak is that the Dockerfile(s) under HoneyBadgerMPC will allow for additional dependencies to be added (via `setup.py` extras, `requirements.txt`, `Pipfile`, or via direct additions into the Dockerfile(s)), thus allowing a developer to make a pull request without needing to deal with the base docker image hosted on dockerhub. Modifications to the base docker image will need to be done via separate pull requests. Versioning of the base Dockerfile should ease that workflow. For instance, `FROM dsluiuc/honeybadgermpc:alpine-0.2.3` would point to a specific base image available on dockerhub with which a developer could implement new features into the HoneyBadgerMPC codebase. New dependencies can be freely added as “experimental”  in the Dockerfile under the HoneyBadgerMPC, and moved into the base image much later on whenever it makes sense.

### Bullet point summary
- Maintain a base image under a separate repository, strictly used to maintain and version base images.
- Explore whether breaking down the base image into smaller specialized ones makes  sense. For instance, one image for pairing, one for ntl, one for pbc, one for charm, etc.
- Just like Drake did in his PR, the Dockerfiles under HoneyBadgerMPC are built `FROM` the base one and explicitly point to a version (to avoid inconsistencies, and to make it clearer what base image is being used).
- We should probably need one Dockerfile for day-to-day development and one for Travis CI, under `.ci` like we already have. Or is this overkill? Is there a simpler approach?
- The Dockerfiles under HoneyBadgerMPC will `pip install` extras for `setup.py` , `requirements.txt`, or `Pipfile` to allow for additional dependencies to be easily added and built locally and on travis. The simplest for now would be to add an `experimental` extra in `setup.py` and to add the line `pip install -e .[experimental]` in Dockerfiles. Also, the Dockerfiles will  be modifiable to add library dependencies not present in the base image. We could add a shell script, e.g.: `install_experimental.sh` that would always be there in the Dockerfiles, i.e.: `RUN ./install_experimental`. A developer would simply need to add additional (non-pip installable) dependencies under that shell script.
- PRs could then be merged with experimental dependencies and in a separate step experimental dependencies could be added to the base image via the regular pull request workflow. The resulting base docker image with new dependencies would have a new version and in order for it to be used would need to be explicitly specified in the Dockerfiles under HoneyBadgerMPC.
- This is more or less what this PR does with the addition that the base image is isolated from the development repository and dependencies can be added as necessary within the development repository without the need to do any manual update relating to dockerhub. Basically, an “experimental” intermediate stage is introduced to enable developers to not have to deal with dockerhub at all and develop whatever is needed with whatever dependencies. Changes to the base image are managed via pull requests in a separate repository, allowing for “asynchronous development” in the sense that one SHOULD NOT need to wait for dockerhub for a new image version to check the work on Travis because that work has already been checked via the “experimental stage”. Moving the experimental dependencies into the base image and updating the Dockefiles under HoneyBadgerMPC to use a new base image version is just a formality and should not interfere with the day-to-day development efforts.

Drake-Eidukas:
Closing in lieu of #351

PR #318, opened 2019-05-24
more renaming [READY]

amiller:
- [x] renames ntl files to be systematic
  - `ccobject.pxd`, `ccobject_impl.h` defines generic C++ object to python mapping
  - `ntlwrapper.pxd`, `ntlwrapper_impl.h` defines python versions of general NTL functions
  - `rsdecode.pxd`, `rsdecode_impl.h` defines interpolation routines, with a `_c` suffix 
  - `hbpmc_ntl_helpers.pyx` defines all of the in-cython functions we use
    `setup.py` builds this as `ntl._hbpmc_ntl_helpers`
  - `__init__.py` imports `*` from `_hbmpc_ntl_helpers` into `ntl.` so we do not need to use `ntl.helpers`
- ~[x] more to come....~


amiller:
We should just take this now, no need to wait for more

PR #319, opened 2019-05-24
Update CHANGELOG.md

amiller:
A list of main components included in v0.1, including an initial "contributors" listing

This list is probably incomplete so far, discuss below!


Drake-Eidukas:
Just about everything I can think of that I've done is included here.

Drake-Eidukas:
Should we wait on more feedback with this or should we just rebase this and merge it in?

PR #325, opened 2019-05-26
Asynchromix coordinator using Ethereum

amiller:
This PR creates an end-to-end driver for the Asynchromix app. It is based around the idea of using a Solidity smart contract as a coordinator. 

At a high level the coordinator is responsible for:
Coordinator:
1. Keeps track of offline preprocessed values. 
    - bits/triples/powers for the mixing epoch
    - input masks for client inputs
    - requires all `n` servers to report on how many preprocess values are available, taking the `min` of these
2. Accept client inputs
      - Reserves an input mask for each client
      - Accept the published "masked input" from each client
3. Initiate mixing epochs

The Asynchromix Client does:
1. Wait until an input mask is available
2. Reserve an input mask `contract.reserve_inputmask()`
3. Request shares of the input mask from the servers and reconstructs
4. Publish the masked input `contract.submit_message()`

The Asynchromix Server does:
1. Offline phase  (1a Bits/triples, 1b input masks)
  I. Wait until more input masks or more mix preprocessing is needed (try to maintain a buffer of at least 2 mixes worth)
  II. Run the offline phase (`randousha`, plus an MPC program for `generate_triples` or `generate_bits`)
  III. Submit an updated preprocessing report `contract.preprocess_report()`
2. Handling client input. If a client requests an input mask share, check that the client has already reserved the corresponding input mask `contract.client_authorized()`
3. Participating in sequential mixing `epoch`s
   a. Wait for the next mix to be initiated `contract.epochs_initiated()`
   b. Client inputs `inputs[K*epoch:K*(epoch+1)]`
   c. Preprocessing `triples[PER_MIX_TRIPLES*epoch:PER_MIX_TRIPLES*(epoch+1)]`
   d. Call the Mpc program in a fake context

Some of the challenges that had to be addressed:
- Running multiple Mpc contexts sequentially, we need to be able to specify where the preprocessing comes from. Currently we have a workaround, but the current approach is not sustainable. The workaround is to clear the `del MixinBase.pp_elements[key]` where `key` contains one server's id and then overwrite the file. A more sustainable solution would be to make the source of preprocessing a parameter to the `Mpc` context object itself. This way concurrent instances of Mpc could use different preprocessing values.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,7 @@
 #   - pbc
 #   - charm
 #   - base pip dependencies (cython & setup.py)
+#   - ethereum
 # 
 # Thereafter, it adds ever increasing levels of dependencies-- 
 #   - Test requirements (including doc requirements)
@@ -87,6 +88,25 @@ RUN git clone https://github.com/JHUISI/charm.git
 WORKDIR /charm/
 RUN git reset --hard be9587ccdd4d61c591fb50728ebf2a4690a2064f
 RUN ./configure.sh
+RUN make install
+WORKDIR /
+
+
+# Ethereum .[eth] extras
+RUN apt-get install -y --no-install-recommends \
+    git cmake g++ \
+    libffi-dev libssl-dev sudo
+RUN curl -sL https://deb.nodesource.com/setup_8.x | bash
+RUN apt-get install -y --no-install-recommends nodejs npm
+RUN npm install -g ganache-cli
+RUN git clone --recursive https://github.com/ethereum/solidity.git
+WORKDIR /solidity/
+RUN git checkout v0.4.24 # Old version necessary to work???
+RUN git submodule update --init --recursive
+RUN ./scripts/install_deps.sh
+RUN mkdir build/
+WORKDIR build
+RUN cmake ..
 RUN make install
 WORKDIR /
 
--- a/apps/asynchromix/asynchromix.py
+++ b/apps/asynchromix/asynchromix.py
@@ -1,0 +1,518 @@
+"""
+Implementation of Asynchromix MPC Coordinator using an EVM blockchain
+"""
+from web3 import Web3, HTTPProvider
+import time
+import asyncio
+import logging
+from contextlib import contextmanager
+import subprocess
+from honeybadgermpc.router import SimpleRouter
+from honeybadgermpc.utils.misc import subscribe_recv, wrap_send
+from honeybadgermpc.utils.misc import print_exception_callback
+from honeybadgermpc.field import GF
+from honeybadgermpc.polynomial import EvalPoint, polynomials_over
+from honeybadgermpc.elliptic_curve import Subgroup
+from honeybadgermpc.offline_randousha import randousha
+from honeybadgermpc.offline_randousha import generate_triples
+from honeybadgermpc.offline_randousha import generate_bits
+from honeybadgermpc.mpc import Mpc
+from honeybadgermpc.progs.mixins.share_arithmetic import BeaverMultiplyArrays
+from honeybadgermpc.progs.mixins.constants import MixinConstants
+from honeybadgermpc.progs.mixins.base import MixinBase
+from butterfly_network import iterated_butterfly_network
+from ethereum.tools._solidity import compile_code as compile_source
+from web3.contract import ConciseContract
+import os
+
+
+# For fake preprocessing
+from honeybadgermpc.preprocessing import (
+    PreProcessedElements, clear_preprocessing)
+
+field = GF(Subgroup.BLS12_381)
+
+
+async def wait_for_receipt(w3, tx_hash):
+    while True:
+        tx_receipt = w3.eth.getTransactionReceipt(tx_hash)
+        if tx_receipt is not None:
+            break
+        await asyncio.sleep(5)
+    return tx_receipt
+
+
+########
+# Client
+########
+
+class AsynchromixClient(object):
+    def __init__(self, sid, myid, send, recv, w3, contract, req_mask):
+        self.sid = sid
+        self.myid = myid
+        self.contract = contract
+        self.w3 = w3
+        self.req_mask = req_mask
+        self._task = asyncio.ensure_future(self._run())
+        self._task.add_done_callback(print_exception_callback)
+
+    async def _run(self):
+        contract_concise = ConciseContract(self.contract)
+        await asyncio.sleep(60)  # give the servers a head start
+        # Client sends several batches of messages then quits
+        for epoch in range(1000):
+            logging.info(f'[Client] Starting Epoch {epoch}')
+            receipts = []
+            for i in range(32):
+                m = f'message:{epoch}:{i}'
+                task = asyncio.ensure_future(self.send_message(m))
+                task.add_done_callback(print_exception_callback)
+                receipts.append(task)
+            receipts = await asyncio.gather(*receipts)
+
+            while True:  # wait before sending next
+                if contract_concise.outputs_ready() > epoch:
+                    break
+                await asyncio.sleep(5)
+
+    async def _get_inputmask(self, idx):
+        # Private reconstruct
+        contract_concise = ConciseContract(self.contract)
+        n = contract_concise.n()
+        poly = polynomials_over(field)
+        eval_point = EvalPoint(field, n, use_omega_powers=False)
+        shares = []
+        for i in range(n):
+            share = self.req_mask(i, idx)
+            shares.append(share)
+        shares = await asyncio.gather(*shares)
+        shares = [(eval_point(i), share) for i, share in enumerate(shares)]
+        mask = poly.interpolate_at(shares, 0)
+        return mask
+
+    async def join(self):
+        await self._task
+
+    async def send_message(self, m):
+        # Submit a message to be mixed
+        contract_concise = ConciseContract(self.contract)
+
+        # Step 1. Wait until there is input available, and enough triples
+        while True:
+            inputmasks_available = contract_concise.inputmasks_available()
+            # logging.infof'inputmasks_available: {inputmasks_available}')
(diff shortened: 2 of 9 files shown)


apps/asynchromix/butterfly_network.py
```diff
@@ -9,8 +9,9 @@
 async def batch_switch(ctx, xs, ys, n):
     pp_elements = PreProcessedElements()
     sbits = [pp_elements.get_one_minus_one_rand(ctx).v for _ in range(n//2)]
-    ns = [1 / ctx.field(2) for _ in range(n)]
+    ns = [1 / ctx.field(2) for _ in range(n//2)]
 
+    assert len(xs) == len(ys) == len(sbits) == n // 2
     xs, ys, sbits = list(map(ctx.ShareArray, [xs, ys, sbits]))
```
amiller, line 15: This fix was necessary - initially xs and ys were half the size of the sbits. This indicates a possible issue, if the ShareArray multiplication just truncates rather than reporting incompatible dimensions

honeybadgermpc/preprocessing.py
```diff
@@ -56,6 +75,128 @@ def _write_shares_to_file(self, f, degree, myid, shares):
...
+    def get_one_minus_one_rand(self, ctx):
+        file_suffix = f"_{ctx.N}_{ctx.t}-{ctx.myid}.share"
+        fpath = f"{PreProcessingConstants.ONE_MINUS_ONE_FILE_NAME_PREFIX}{file_suffix}"
+        key = (ctx.myid, ctx.N, ctx.t)
+        if key not in self._one_minus_one_rands:
+            self._one_minus_one_rands[key] = iter(
+                self._read_share_values_from_file(fpath))
+        try:
+            return ctx.Share(next(self._one_minus_one_rands[key]))
+        except StopIteration:
```
Drake-Eidukas, line 180: 1. We should turn this print statement into a logging statement
2. I think it's a great change that now the exception mentions which preprocessing is running out
3. I think we should extract this into some sort of a helper method such that all of the getters will annotate the exception with the type of preprocessing missing.

apps/asynchromix/asynchromix.sol
```diff
@@ -0,0 +1,252 @@
+pragma solidity >=0.4.22 <0.6.0;
```
Drake-Eidukas, line 1: code in this file is very well documented! I'm not even sure what language this is though, nor do I really know exactly what `AsynchromixCoordinator` is truly all about. Perhaps some documentation that describes the big picture here would be good.

Dockerfile
```diff
@@ -91,6 +92,25 @@ RUN make install
 WORKDIR /
 
 
+# Ethereum .[eth] extras
+RUN apt-get install -y --no-install-recommends \
```
Drake-Eidukas, line 96: So `git, `g++`, and `libssl-dev` are already installed by this point. I think the rest of the apt dependencies should be installed there.

Also, I think it would be best if the apt install commands were consolidated to the beginning.

Dockerfile
```diff
@@ -91,6 +92,25 @@ RUN make install
...
+    libffi-dev libssl-dev sudo
+RUN curl -sL https://deb.nodesource.com/setup_8.x | bash
+RUN apt-get install -y --no-install-recommends nodejs npm
+RUN npm install -g ganache-cli
+RUN git clone --recursive https://github.com/ethereum/solidity.git
+WORKDIR /solidity/
+RUN git checkout v0.4.24 # Old version necessary to work???
+RUN git submodule update --init --recursive
+RUN ./scripts/install_deps.sh
+RUN mkdir build/
```
Drake-Eidukas, line 107: Might be cleaner to do something like
```
RUN mkdir build/
RUN cmake .
RUN make install -c build/
```
But this is getting into the territory of super nitpicky.

honeybadgermpc/offline_randousha.py
```diff
@@ -182,8 +184,43 @@ def _get_send_recv(tag):
...
+    subscribe_recv_task, subscribe = subscribe_recv(_recv)
+
+    def _get_send_recv(tag):
+        return wrap_send(tag, _send), subscribe(tag)
+
+    # Start listening for my share of t and 2t shares from all parties.
+    send, recv = _get_send_recv("randousha")
+    rs_t2t = await randousha(n, t, 2 * k, my_id, send, recv, field)
+
+    us_t2t = rs_t2t[0*k:1*k]
```
Drake-Eidukas: I think it would be a bit more readable to have 
```
us_t2t = rs_t2t[0:k]
rs_t2t = rs_t2t[k:2*k]
```

Also, I'm not really sure what each of these lists are from the variable names... perhaps this code could use some more documentation as to what exactly is going on for future readers.

honeybadgermpc/offline_randousha.py
```diff
@@ -182,8 +184,43 @@ def _get_send_recv(tag):
...
+    rs_t2t = rs_t2t[1*k:2*k]
+
+    us_t, _ = zip(*us_t2t)
+    us_t = list(map(field, us_t))
+    rs_t, rs_2t = zip(*rs_t2t)
+
+    # Config just has to specify mixins used by switching_network
+    config = {}
```
Drake-Eidukas: Does this need more mixins? The comment makes it seem like this needs mixins.

Drake-Eidukas (changes requested):
Seems pretty good, but I don't really have much depth of knowledge about asynchromix and what that's about... perhaps it would be a good idea to have someone with a bit more knowledge double check since I feel like I can't give a good review of this. 

That said, I think the dockerfile changes should be incorporated to keep that in a clean state. In terms of changes to preprocessing, I think that print statement should either become a logging statement, or stripped out, with anything else being left for a future refactor of our test preprocessing system, as it currently contains quite a few pain points for a new developer.

Finally, could you squash the commits and rebase?

apps/asynchromix/asynchromix.sol
Drake-Eidukas: Specifically, moving some of the docs from the PR description here might be a good idea, as those docs are pretty good at describing what this file's all about.

apps/asynchromix/asynchromix.py
```diff
@@ -0,0 +1,542 @@
+"""
+Implementation of Asynchromix MPC Coordinator using an EVM blockchain
+"""
+from web3 import Web3, HTTPProvider
+import time
+import asyncio
+import logging
+from contextlib import contextmanager
+import subprocess
+# from honeybadgermpc.exceptions import HoneyBadgerMPCError
```
Drake-Eidukas: Can you remove the commented out imports?

apps/asynchromix/asynchromix.py
```diff
@@ -0,0 +1,542 @@
...
+async def wait_for_receipt(w3, tx_hash):
+    while True:
+        tx_receipt = w3.eth.getTransactionReceipt(tx_hash)
+        if tx_receipt is not None:
+            break
+        await asyncio.sleep(5)
+    return tx_receipt
+
+
+def print_exception_callback(future):
```
Drake-Eidukas: This could probably be placed in the `utils` files.

apps/asynchromix/asynchromix.py
```diff
@@ -0,0 +1,542 @@
...
+        loop.close()
+
+
+def test_asynchromix():
+    import time
+    # cmd = 'testrpc -a 50 2>&1 | tee -a acctKeys.json'
+    # with run_and_terminate_process(cmd, shell=True,
+    # stdout=sys.stdout, stderr=sys.stderr) as proc:
+    cmd = "ganache-cli -p 8545 -a 50 -b 1 > acctKeys.json 2>&1"
+    print("Running", cmd)
```
Drake-Eidukas: These print statements should probably be moved over to logging statements.

apps/asynchromix/asynchromix.py
```diff
@@ -0,0 +1,542 @@
...
+    # Create the coordinator contract and web3 interface to it
+    compiled_sol = compile_source(open(os.path.join(os.path.dirname(
+        __file__), 'asynchromix.sol')).read())  # Compiled source code
+    contract_interface = compiled_sol['<stdin>:AsynchromixCoordinator']
+    contract_class = w3.eth.contract(abi=contract_interface['abi'],
+                                     bytecode=contract_interface['bin'])
+    # tx_hash = contract_class.constructor(w3.eth.accounts[:7],2).transact(
+    #   {'from':w3.eth.accounts[0]})  # n=7, t=2
+
+    tx_hash = contract_class.constructor(w3.eth.accounts[:4], 1).transact(
```
Drake-Eidukas, line 437: It would probably be better if 4 and 1 were based on some `n`, `t` constants or parameters being passed in.

Drake-Eidukas (changes requested):
Missed this file

honeybadgermpc/offline_randousha.py
amiller: good catch, the comment comment was copypasta. the inner program here actually uses no mixins

apps/asynchromix/asynchromix.py
amiller: k

apps/asynchromix/asynchromix.py
amiller: fixed, also changed the reference in test_hbavss which is the only other place it's used

honeybadgermpc/offline_randousha.py
amiller: Add a descriptive comment

apps/asynchromix/asynchromix.py
amiller: they only show up once, this is the place to change it

amiller:
This addressed all the comments, should be mergeable

Drake-Eidukas (approved):

PR #326, opened 2019-05-28
Implement tree-like fold function for ShareArray

Drake-Eidukas:
Share Equality and Comparison both require a tree-like fold for maximum
efficiency. This implements it on the ShareArray class to be as DRY as
possible. This allows operations such as summing and taking the product sum of a sharearray to proceed as parallel as possible.

### DIFF
--- a/honeybadgermpc/progs/mixins/dataflow.py
+++ b/honeybadgermpc/progs/mixins/dataflow.py
@@ -228,6 +228,58 @@ class ShareArray(ABC):
 
     __truediv__ = __floordiv__ = __div__
 
+    async def _tree_fold(self, op):
+        """ Apply a provided operation in a 'tree'-like fashion--
+        Instead of folding sequentially across the shares of the array which
+        creates a linked-list like chain of operations as shares resolve, this
+        applies them in a tree-like fashion, which involves only log(n) levels of
+        application. This requires the array to be non-empty.
+
+        example:
+            A regular reduce would proceed as follows:
+                sum([1,2,3,4,5]) => 1 + 2 + 3 + 4 + 5
+            Instead, this will proceed as follows:
+                sum([1,2,3,4,5]) => ((1+2) + (3+4)) + 5
+
+        args:
+            op (function): A binary function that takes two shares and returns a share
+                or sharefuture. This is required to be commutative. For this to be
+                useful, it should also be non-strict / lazy (i.e. the result arrives
+                asynchronously)
+
+        returns:
+            A Share or ShareFuture representing the iterated binary operation of op on
+            the shares of this array
+        """
+        shares = self._shares
+        assert len(shares) > 0
+
+        while len(shares) > 1:
+            left, right = shares[::2], shares[1::2]
+            extra = None
+            if len(left) > len(right):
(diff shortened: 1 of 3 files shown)


Drake-Eidukas:
This also fixes a bug where equality could break if the security parameter was odd, as the zip operation truncates the last element of x if x had an odd length.

honeybadgermpc/progs/mixins/dataflow.py
```diff
@@ -228,6 +228,50 @@ def __div__(self, other):
...
+
+        returns:
+            A Share or ShareFuture representing the iterated binary operation of op on
+            the shares of this array
+        """
+        shares = self._shares
+        assert len(shares) > 0
+
+        while len(shares) > 1:
+            results = [op(a, b) for a, b in zip(shares[::2], shares[1::2])]
```
rahulgovind: Unfortunately, I had observed that share reconstructions do not work very well when done in parallel. The reconstructions need to be executed in the same order across all nodes, and I believe this condition can be violated when multiple reconstructions are carried out in parallel. 

Additionally, there is a pretty high overhead with calling the operations on individual shares. Instead, for each level, you can apply the operation on arrays instead. I have versions of this in https://github.com/rahulgovind/HoneyBadgerMPC/blob/fixed-point-order/honeybadgermpc/fixedpoint.py. binary_addition has a prefix sum, get_carry_bits has a reduction.

rahulgovind (changes requested):

honeybadgermpc/progs/mixins/dataflow.py
Drake-Eidukas: 1. Not sure how to address this point, as I don't think this changes the way we do share reconstruction for the arrays... but maybe I'm misunderstanding your point.
2. That's a good idea-- I'll update this to do that.

Drake-Eidukas: Yep, about 10 seconds faster for an array with 1000 elements using number 2! good catch :D

rahulgovind: I'm not too sure how point 1 can be addressed either. Just that the await statement is required for now. The changes look great! How much time does it take to do the 1000 element array operation right now?

rahulgovind (approved):

honeybadgermpc/progs/mixins/dataflow.py
Drake-Eidukas: ~So, the total time to run all the tests in the file with preprocessing 10000 elements took about 38 seconds with this new version, the old version took 53 seconds for the whole file, which includes the same operation 3 times on 80 elements for the equality test. not a very scientific approach to testing the time... I'd keep the test in but I was having difficulty with `test_preprocessing` to get the test case of 1000 elements to work when I ran the tests over all of the files (the tests worked perfectly when only running on this one file... but there's issues where test_preprocessing will cache calls to generate more elements but then not actually reset the elements in the preprocessing to contain enough elements. Perhaps I can add it back in once that gets refactored).~
With the exact amount of preprocessing required, it takes ~1.3 seconds on my machine

Issue #327, opened 2019-05-28
Investigate failing `test_decoding_all_zeros` in travis


Drake-Eidukas:
Sometimes, this test fails in travis and requires a restart of travis.
```
=================================== FAILURES ===================================
___________________________ test_decoding_all_zeros ____________________________
[gw1] linux -- Python 3.7.3 /usr/local/bin/python
    def test_decoding_all_zeros():
        int_msg = [0, 0, 0, 0, 0, 0, 0, 0]
        k = len(int_msg)  # length of message
        n = 22  # size of encoded message
        p = 53  # prime
        t = k - 1  # degree of polynomial
    
        enc, dec, _ = make_wb_encoder_decoder(n, k, p)
        encoded = enc(int_msg)
    
        # Check decoding with no errors
        # https://github.com/initc3/HoneyBadgerMPC/issues/143
        with pytest.raises(IndexError):
            # If this error is not raised then the bug has been fixed :D
            _ = dec(encoded, debug=False)
    
        # Corrupt with maximum number of erasures:
        cmax = n - 2 * t - 1
    
        corrupted = corrupt(encoded, num_errors=0, num_nones=cmax)
        with pytest.raises(IndexError):
            _ = dec(corrupted, debug=False)
    
        # Corrupt with maximum number of errors:
        emax = (n - 2 * t - 1) // 2
        corrupted = corrupt(encoded, num_errors=emax, num_nones=0)
        coeffs = dec(corrupted, debug=False)
    
        # This also showcases inconsistency in polynomial functions
        # poly([]) should be equal to poly([0])
        assert coeffs == []
    
        # Corrupt with a mixture of errors and erasures
        e = emax // 2
        c = cmax // 4
        corrupted = corrupt(encoded, num_errors=e, num_nones=c)
>       coeffs = dec(corrupted, debug=False)
tests/test_reed_solomon_wb.py:76: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
honeybadgermpc/reed_solomon_wb.py:146: in decode
    q_, e_ = solve_system(enc_m, max_e=e, debug=debug)
honeybadgermpc/reed_solomon_wb.py:106: in solve_system
    solution = some_solution(system, free_variable_value=1)
honeybadgermpc/reed_solomon_wb.py:253: in some_solution
[...]

Issue #143, opened 2018-11-25
wb_interpolate can't decode a message that is all zeros


amiller:
This commit adds a test case for wb_interpolate that seems to indicate a bug. Usually when we've encountered this case, it's because of some other mistake (we usually aren't decoding all 0's). But there's nothing wrong with a sharing that is all 0's, it's the constant polynomial f(x) = 0 which should represent a value 0.

https://github.com/initc3/HoneyBadgerMPC/commit/20620fd69f97797d93908006abc5f21b350b4e65
https://github.com/initc3/HoneyBadgerMPC/tree/asyncmix-master-wbinterpolatefix

This test fails (TODO: update when travis finishes)
https://travis-ci.org/initc3/HoneyBadgerMPC/branches

The error in this case is something like:
```
File "/usr/src/HoneyBadgerMPC/honeybadgermpc/linearsolver.py", line 65, in isPivotColumn
    while A[i][j] == 0 and i < len(A):
IndexError: list index out of range
```

The problem seems to be that `wb_interpolate` fails to find a pivot point. Is this the only case where that can happen? Would a special check be sufficient?

smkuls:
This does not happen with Gao's algorithm. I have asked @rahulgovind to write a test case for this for both Gao and `wb_interpolate` so that we keep track of this issue.

amiller:
I think this is no longer an issue, I'm closing as all our batch decoding has been significantly changed and there are no outstanding problems

PR #328, opened 2019-05-28
Fix issue with wb_interpolate exceeding array bounds

Drake-Eidukas:
#143 alluded to this, but was closed. It occurred again recently, (#327), and this fixes it.

### DIFF
--- a/honeybadgermpc/reed_solomon_wb.py
+++ b/honeybadgermpc/reed_solomon_wb.py
@@ -213,7 +213,7 @@ def no_solution(a):
 # and return the row index of the 1 if it exists
 def is_pivot_column(a, j):
     i = 0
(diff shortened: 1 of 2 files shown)


honeybadgermpc/reed_solomon_wb.py
```diff
@@ -213,7 +213,7 @@ def no_solution(a):
 # and return the row index of the 1 if it exists
 def is_pivot_column(a, j):
     i = 0
-    while a[i][j] == 0 and i < len(a):
+    while i < len(a) and a[i][j] == 0:
```
rahulgovind, line 216: That's awesome! The fix was way more straightforward than I thought it would be.

rahulgovind (approved):

amiller (approved):
Awesome. A nice relief

Issue #320, opened 2019-05-24
Fixed point and comparison


amiller:
@rahulgovind has already made a pretty decent implementation of an MPC fixed point object [1]. This stores a 64-bit number in one field element, with the remaining bits used as padding (we could push it these constants a bit further if needed). 
https://github.com/rahulgovind/HoneyBadgerMPC/blob/fixed-point-order/honeybadgermpc/fixedpoint.py
The linked code is embedded within a more complicated project implementing some machine learning algorithm demos. This issue can be closed just by extracting the fixed-point logic itself.

[1] Catrina and Saxena 2010 https://link.springer.com/chapter/10.1007/978-3-642-14577-3_6

Issue #29, opened 2018-08-24
Machine Learning


amiller:
Someone needs to take an ML class and implement an MPC app for training a neural network

amiller:
MPyC has a machine learning demo
https://github.com/lschoe/mpyc/blob/master/demos/cnnmnist.py

smkuls:
Removing the benchmarking label from this since we don't have an implementation.

rahulgovind:
I have some code ready for fixed-point arithmetic and basic linear regression at https://github.com/rahulgovind/HoneyBadgerMPC/tree/fixed-point in case anyone is interested. Here are some problems I see so far
- Random bit generation is very slow and is most definitely going to be the bottleneck. Need to come up with ways to speed this up
- Free dot product-like operations are integral to optimising the operations. Unsure of how exactly to expose an interface to this at this point. Any suggestions are welcome. 
- Need to think of an interface to batch operations or a way to automatically batch independent operations

amiller:
@rahulgovind Can you raise a PR including the fixed point operations you implemented before we forget?

PR #334, opened 2019-06-06
Fixed point program

amiller:
This is a subset of the functionality from @rahulgovind's machine-learning project #29 #320  https://github.com/rahulgovind/HoneyBadgerMPC/blob/fixed-point/honeybadgermpc/fixedpoint.py

There are three tests so far: 
- [x] addition and subtraction
- [x] multiplication and reduction
- [x] comparison

~Comparison currently is broken, so needs to be debugged~
Now works :+1:

### DIFF
--- a/honeybadgermpc/progs/fixedpoint.py
+++ b/honeybadgermpc/progs/fixedpoint.py
@@ -1,0 +1,223 @@
+import logging
+import asyncio
+import time
+from honeybadgermpc.mpc import TaskProgramRunner
+from honeybadgermpc.elliptic_curve import Subgroup
+from honeybadgermpc.field import GF
+from honeybadgermpc.preprocessing import (
+    PreProcessedElements as FakePreProcessedElements)
+from honeybadgermpc.progs.mixins.share_arithmetic import (
+    MixinConstants, BeaverMultiply)
+
+
+config = {MixinConstants.MultiplyShare: BeaverMultiply(), }
+
+
+# Fixed Point parameters
+
+F = 32  # The precision (binary bits)
+KAPPA = 32  # Statistical security parameter
+K = 64  # Total number of padding bits (?)
+p = modulus = Subgroup.BLS12_381
+Field = GF(p)
+
+
+# General (non MPC) fixed point functions
+def to_fixed_point_repr(x, f=F):
+    return int(x * 2 ** f)
+
+
+def from_fixed_point_repr(x, k=K, f=F, signed=True):
+    x = x.value
+    if x >= 2 ** (k - 1) and signed:
+        x = -(p - x)
+
+    return float(x) / 2 ** f
+
+
+def binary_repr(x, k):
+    """
+    Convert x to a k-bit representation
+    Least significant bit first
+    """
+    def _binary_repr(v):
+        res = []
+        v = int(v)
+        for i in range(k):
+            res.append(v % 2)
+            v //= 2
+        return res
+
+    assert type(x) is int
+    return _binary_repr(x)
+
+
+# MPC operations for fixed point
+async def random2m(ctx, m):
+    result = ctx.Share(0)
+    bits = []
+    for i in range(m):
+        bits.append(ctx.preproc.get_bit(ctx))
+        result = result + Field(2) ** i * bits[-1]
+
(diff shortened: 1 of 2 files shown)


honeybadgermpc/progs/fixedpoint.py
```diff
@@ -0,0 +1,229 @@
...
+
+    carry_bits = [(await (ai * bi)) for ai, bi in zip(a_bits, b_bits)]
+    all_one_bits = [ctx.Share(ai.v + bi.v - 2 * carryi.v) for ai, bi, carryi in
+                    zip(a_bits, b_bits, carry_bits)]
+    carry_bits.append(ctx.Share(low_carry_bit))
+    all_one_bits.append(ctx.Share(0))
+    return (await _bit_ltl_reduce(list(zip(carry_bits, all_one_bits))))[0]
+
+
+async def bit_ltl(ctx, a, b_bits):
```
Drake-Eidukas, line 101: What does `ltl` stand for? This might make sense to document / rename.

honeybadgermpc/progs/fixedpoint.py
```diff
@@ -0,0 +1,229 @@
...
+        return await self.sub(x).ltz()
+
+    async def div(self, x):
+        if type(x) in [float, int]:
+            return await self.mul(FixedPoint(self.ctx, 1. / x))
+        raise NotImplementedError
+
+
+async def _prog(ctx):
+    ctx.preproc = FakePreProcessedElements()
```
Drake-Eidukas, line 186: I don't think this is a good idea-- won't this file not otherwise work with other programs unless they also modify context like this? At the least we should document this.

honeybadgermpc/progs/fixedpoint.py
```diff
@@ -0,0 +1,229 @@
...
+    result = ctx.Share(0)
+    bits = []
+    for i in range(m):
+        bits.append(ctx.preproc.get_bit(ctx))
+        result = result + Field(2) ** i * bits[-1]
+
+    return result, bits
+
+
+async def trunc_pr(ctx, x, k, m):
```
Drake-Eidukas, line 66: What does `trunc_pr` do? I think this needs either a better name or some documentation.

honeybadgermpc/progs/fixedpoint.py
```diff
@@ -0,0 +1,229 @@
...
+            res.append(v % 2)
+            v //= 2
+            return res
+
+    assert type(x) is int
+    return _binary_repr(x)
+
+
+# MPC operations for fixed point
+async def random2m(ctx, m):
```
Drake-Eidukas, line 56: why is this async?

Drake-Eidukas (changes requested):
I think in general, we should move these methods to be helper methods on the class, or move them into some shared utility file if they can see use in other contexts. I feel like this would help the code be less reliant on constants in the file and be able to be changed grammatically.

amiller:
All the above comments are good. I will probably leave this open for a while so we can use it, but not merge it right away until it's refactored a bit. Also we should sort out 'division' and implement the operator functions like `__mul__`. @rahulgovind has a more elaborate version that works on arrays as well for performance

honeybadgermpc/progs/fixedpoint.py
rahulgovind: Followed notation/function names in the paper. A link to the paper in the header should address this and other similar reviews.

honeybadgermpc/progs/fixedpoint.py
rahulgovind: Was async before since the random bits were generated on the fly. Doesn't need to be async anymore though.

sanket1729: I don't understand this, but making this async fails the test :(

sanket1729: Although I see no reason why this should be async

sanket1729:
Overriding __mul__ is somehow not working. 

`await a.__mul__(b)` works fine, but somehow `await a*b` is not cool

amiller:
Closed in favor of #368

PR #336, opened 2019-06-07
Refactor preprocessing

Drake-Eidukas:
This PR refactors preprocessing into a more plugin based model. This also changes the way we cache elements-- when asked to generate elements, we simply make sure that the cache has at least that amount of elements. If asked to append, this will append to the existing preprocessing file, otherwise, it will overwrite the existing data.

We also turn preprocessing into a singleton to prevent overuse of memory.

A nice side effect is that since this doesn't needlessly generate values, we were able to cut testing time by 33% on my machine (300s -> 200s). 

Finally, this fixes issues where the preprocessing would run out due to the execution of other tests as the caching wasn't being properly updated.

### DIFF
--- a/apps/asynchromix/powermixing.py
+++ b/apps/asynchromix/powermixing.py
@@ -4,8 +4,7 @@ from time import time
 from honeybadgermpc.mpc import TaskProgramRunner
 from honeybadgermpc.field import GF
 from honeybadgermpc.elliptic_curve import Subgroup
-from honeybadgermpc.preprocessing import PreProcessedElements, PreProcessingConstants
-from honeybadgermpc.preprocessing import wait_for_preprocessing, preprocessing_done
+from honeybadgermpc.preprocessing import PreProcessedElements
 import logging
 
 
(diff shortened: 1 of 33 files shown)


amiller:
What's the rationale for changing the `get_rand(ctx, ...)` into `get(ctx, 'rand', ...)`? This seems even less explicit. What I'd really prefer is `ctx.pp.get_rand()`.

amiller:
The most important thing is to probably check that the blockchain example `python apps/asynchromix/asynchromix.py` still works, because this does not use the `FakePreProcessedElements`. It's important to distinguish between the fake preprocessing files we use for tests and the general use of preprocessing, so this refactoring would ideally make that clear

Drake-Eidukas:
With the new changes, we can now run without `--dist=loadfile`, as now most tests use proper isolation with their preprocessing (each runner uses their own directory). This all cuts testing time down from `300`s -> `120`s on my machine (running with `-n8`)

Drake-Eidukas:
On travis, the testing time goes from ~20min => ~14min

apps/asynchromix/powermixing.py
```diff
@@ -166,9 +165,14 @@
 if __name__ == "__main__":
     from honeybadgermpc.config import HbmpcConfig
 
+    HbmpcConfig.load_config()
```
Drake-Eidukas, line 166: Not exactly sure how to fix this code-- I couldn't figure out how to get it running.

Drake-Eidukas:
Now, testing takes `100`s

Drake-Eidukas:
See #339

PR #339, opened 2019-06-22
Refactor preprocessing to improve caching and use mixins

Drake-Eidukas:
# Improving preprocessing

Moving the previous PR (#336) to this so I can base a pr for batched share opening against this.

## What this PR does: 
- Refactors preprocessing to use a mixin model to improve testability and modularity of preprocessing
- Reimplements caching of preprocessing elements to improve performance and reduce testing bugginess
- Refactors the codebase to only access preprocessing elements through a context.
- General refactors to clean up code I came across

## Benefits from this PR:
- Testing time in travis went from ~20-21 minutes down to ~13-14 minutes (approx 2-3 minutes of both measurements are bootstrapping travis, so 18 min => 11 min is the actual testing speedup). 
- Testing time locally on my machine (`pytest -n8 -v`) down from 270-315s => 90-110s
- Tests are much less brittle, with the exception being `asynchromix` tests. All test workers follow good test data isolation practices, as each test worker creates a new temporary directory inside of `sharedata` to store data in that worker to prevent concurrent accesses and improve caching.
  - Added benefit of this is that we are able to preprocess far fewer elements than before (we preprocess about 60k fewer elements of several types if I remember correctly)
  - Far lower memory overhead since the entire preprocessing is only stored in memory once for each worker 
- Cleaner and more intuitive means of interacting with preprocessing.

## Work still to be done:
- Investigate the multiton pattern used for the `PreProcessedElements` class to see if there's a better method of keeping good cache performance (memory + execution time) while being more explicit rather than implicit.
- Unify shared setup and teardown boilerplate throughout remaining tests to use existing fixtures.
- Add documentation on how to use and modify preprocessing in the future
- Work more closely on investigating the differences between preprocessing and fake preprocessing and implementing them.
- Auto-cleanup the `sharedata/` temp directories after all tests have run so this is friendlier on people running outside of docker.

### DIFF
--- a/.ci/travis-script.sh
+++ b/.ci/travis-script.sh
@@ -10,8 +10,7 @@ if [ "${BUILD}" == "tests" ]; then
         --cov-report=term-missing \
         --cov-report=xml \
         -Wignore::DeprecationWarning \
-        -nauto \
-        --dist=loadfile
+        -nauto 
 
     IMAGE_NAME=$(docker ps -alq --format "{{.Names}}")
     docker cp $IMAGE_NAME:/usr/src/HoneyBadgerMPC/coverage.xml .
--- a/apps/asynchromix/asynchromix.py
+++ b/apps/asynchromix/asynchromix.py
@@ -8,9 +8,14 @@ import logging
 from contextlib import contextmanager
 import subprocess
 from honeybadgermpc.router import SimpleRouter
-from honeybadgermpc.utils.misc import subscribe_recv, wrap_send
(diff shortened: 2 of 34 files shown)


Drake-Eidukas:
New benchmark: On my macbook, running all the tests goes from 30-32 minutes down to 15-16

amiller (approved):
Ok, iterated to my satisfaction. I still want to refactor this to make it possible to use actual preprocessing more easily not just fake preprocessing and to make that more clear/explicit..... All that for later

PR #340, opened 2019-06-22
Batched Share Opening

Drake-Eidukas:
This creates a more modular approach to opening shares through the use of a `ShareManager` class. This will be the interface I plan to use when implementing batched share openings. 

- [x] Step 1: Implement Share Manager
- [ ] Step 2: Implement Batching
  - [ ] Step 2a: Implement BatchingShareManger
  - [ ] Step 2b: Refactor codebase to use BatchingShareManager somehow
- [ ] Step 3: Writeup about Batching Share Opening


amiller (dismissed):
~I don't want to slow down your ind. study work, but I do not want to merge. The automatic batching is too complicated of a feature to make mandatory. This implementation strategy moves ALL of the logic of share opening into the batch manager, whereas my goal is to have the `mpc.py` be a simple self-contained implementation of share opening, with ShareManager entirely optional as a programming convenience. So my recommendation is that we just let these diverge for your ind. study and I can reimplement this for merging based on your proof of concept~
I'll review this later, I dont think its meant to be ready yet

amiller:
Closing this until someone wants to revisit it

PR #348, opened 2019-07-07
Hashing to ZR/Group

tyurek:
Adds functions to hash into group and generate fiat-shamir challenges. Fixes underlying rust code to use a random number generator that can be seeded with a 256 bit seed (instead of 128, which wouldn't be enough entropy for a ~2^256 sized group)


tyurek:
This one will also require rust to be recompiled in order for tests to pass.

honeybadgermpc/betterpairing.py
```diff
@@ -394,15 +409,28 @@ def rand(seed=None):
...
-            assert len(seed) == 4
-            out.rand(seed[0], seed[1], seed[2], seed[3])
+            assert len(seed) <= 8
+            out.rand(seed)
         return G2(out)
 
+    # length determines how many G2 values to return
+    @staticmethod
+    def hash(*args, length=1):
+        bytebuf = pickle.dumps(args)
```
amiller: I'm wary of using pickle for this. I think there's no guarantee that pickle gives canonical encodings, there may be multiple valid encodings produced at different times or on different platforms for "equal" objects.

amiller: Suggest making this hash explicitly require a bytes-like object as input

tyurek: I'd counter that this would make the interface more cumbersome while not necessarily reducing errors. The result I think would be requiring a pickle import on every file that calls for hashing and the call would just be G1.hash(pickle.dumps(...)) anyway. 

Regardless, there shouldn't be encoding differences for G1 and G2, as they are encoded using their affine coordinates. ZR is straightforward, so they one thing I wouldn't be completely sure about is GT, but it looks like all the values are compared directly, which would mean that an encoding difference would imply inequality.

If we're talking about serializing objects outside of the betterpairing library, then of course I can't speak for that. I'm not familiar with real world examples of how this can come up. We probably should look into something more secure than pickle anyway though.

tyurek:
whoops, didn't realize I had a flake8 error still

tyurek:
bump

Drake-Eidukas:
See #354

PR #351, opened 2019-07-12
Reimplement Dockerfile / Travis to use better caching

Drake-Eidukas:
We discussed changes to docker / travis / dockerhub many times, this is the culmination of my research into best practices and experimentation. I believe this will give us the best flexibility on travis, better views on travis performance metrics, and a smoother dev experience surrounding travis.

### TL;DR:
- Rewrote dockerfile for better cache performance
- re-implemented travis to properly use dockerhub and build our dependencies intelligently. 
  - If dependencies change, docker image is rebuilt, and when merged into dev then dockerhub gets updated.
  - Due to caching, this is very quick in the best-case (~ 2 minutes)
  - Broken into jobs to make a clearer view of the current travis workload.
- Implemented `scripts/stager.py` which facilitates cache-friendly builds and pulls via dockerhub.

### TODO: 
- Look into intelligently building / rebuilding NTL code through the docker image.
- More separation on the build stage for better caching.
- I hardcoded my (encrypted) credentials to the `travis.yml` file. I'd like to have one of you guys use yours instead for when I leave.
-----

### `black` stage added to travis
- I ran black to get this PR to pass, which is why there are so many unrelated style changes
- There was an existing stage but it didn't perform any checks.

### Reimplement Dockerfile:
- I reordered and added targets to the dockerfile.
  - `base` 
    - Contains essential environment variables and packages required by all other build targets
  - `build` 
    - Contains a heavier toolchain, used to build statically compiled dependencies which are later copied into `prod` and `dev` targets.
  - `pre-prod` 
    - This is where the rest of the installation to get this prod-ready are contained.
    - *NOTE*: We DO NOT do `COPY . .` in this target which significantly improves cache performance when building `pre-dev`.
  - `prod` 
    - "terminal" target, simply copies over the build context over `pre-prod` for use in testing.
    - This stage gets uploaded by travis. It is tagged with the commit hash for PR runs, and the branch name for push runs.
  - `pre-dev` 
    - Similar to `pre-prod`, this is where any dev-level dependencies are installed. Essentially a catch-all stage before `dev`, which similarly to `prod` just copies files over.
  - `dev`
    - Same as `prod`.

### Reimplement Travis:
- Travis was rebuilt from the ground up using stages.
- Differences:
  - More descriptive travis webpage
  - Now builds using the same dockerfile as dev use
  - Utilizes new `scripts/stager.py` script to fetch, build, tag, and push docker images.
  - Uploads newly built dockerhub images on push builds 
  - When dependencies change, travis builds the changes and uploads to dockerhub instead of failing. 

### Create `scripts/stager.py`
- This script provides "caching superpowers" when building docker images. It parses the passed in docker compose file for a docker file and its associated targets. It then downloads caches from dockerhub and builds the docker images. In the case when no dependencies change, this can be as fast as a minute.
  - This allows for great caching performance on travis when doing a rebuild without causing deadlock situations with conflicting pull requests.
- Perhaps this belongs in its own repository.


Drake-Eidukas:
In travis, installation took about 3 minutes total. It seems to have stopped caching once it copied over NTL and thus installed pip dependencies despite that stage being cached. It's possible the cache image is stale from when I was testing stuff.

Drake-Eidukas:
Furthermore, the docker script is still a bit rough around the edges-- this will be updated as well.

I personally don't use docker-compose much in my daily workflow, so I'm still investigating how to get that to cache as gracefully as just using docker.

Drake-Eidukas:
As of now: when the dockerfile changes, travis will rebuild the image and upload to dockerhub. Right now, there's some funky stuff going on with caching caused by NTL, but I believe it has to do with stale caches due to `__pycache__` directories.

There's nuance to it all, I'll explain in the morning, but this is basically ready to go.

.travis.yml
```diff
@@ -1,21 +1,45 @@
...
 sudo: true
-
 services:
-  - docker
-
-matrix:
+- docker
+env:
+  global:
+  - secure: p2k4l8RmT5JX81rmesKh8MQoJiO5LowU+vfQxeuzvilFAByuf7FaShaV0TMCM2rt3pUa7YnbklMfHI8KA3y1D2D0ShYXIO7YKmpQNYRmlV7hbfnWE8ySeVtuSopwQObivZjfeSbtJfpd4yJHLTxOoicja9LaMRU/UvUgZX6qeAE6ITzZZBe+cT7b4ekD2CGPrsdpCdU3IVxfZjup9jSVzwZOBgZ3PJmTtoI5a4V9jz5SCohUXcV5vq63SkHaKleHEaYpZjdddVrQKtXdfjYB/DDqAgOyYMqY6lAVk2fdRHE+U3oobcuoTa7TP7Gdg/zRvgq9rHCad270duGgLkmqwrCPaCB+QMyEQ3P4t+PYp3mUg4H8xwYPBiW1T0aY3S3ZqnTF1hPbs1TK7Fy8I7RAoroGZyTRlICbeghUul6OzDHXyu+fbe05Jw/qIGbe7ko5vc1bfNotRmbI16KlLhCdXeWWRtv8UWpVIIGVpMLmeOW3vvp1/kz8LUhKg+eXhxOxQpybh5M/z8+pr8L1ZZzvhCu7cVycmKMU1toh5nvHQopEwJ2l8KMe/h3X9DNL0r/OQNke0KhPbhGhiXdZIKxPdH08kyZx/TldA6IK4k/1Ze+CssUSCI3MqkjqkQeqlFB3ciwo4xX1Oiycd1v63+fHyVQpIexrdS3MSeyQgTRc59M=
```
Drake-Eidukas, line 11: These were generated as follows: 
```
$ gem install travis
$ travis encrypt DOCKER_USER=<my_dockerhub_username> --add
$ travis encrypt DOCKER_PASS=<my_dockerhub_password> --add
```

Drake-Eidukas:
I think I've figured out how to use my script to support docker-compose with good caching as well-- I'll test it out and see how it fares in travis.

Drake-Eidukas:
Hmm... everything works just fine locally with docker compose and caching, but this isn't working on travis. I will likely just keep this using plain docker on travis then.

Drake-Eidukas:
Hmm... everything works just fine locally with docker compose and caching, but wasn't working on travis. Travis is now using plain docker instead of docker-compose for that reason. I have left the functionality in the script for use when building docker-compose images, however, I don't think it will use caching that well. 

For development, I personally just use raw docker images instead of compose.

Drake-Eidukas:
PR looks big, but 400 of the lines of the diff are whitespace changes due to black so at least this is only 600 lines of a diff to go through :/

14/21 files changed are also either just deleted files or reformatted files as well.

Drake-Eidukas:
Example Builds:
- Only pythonic changes: https://travis-ci.org/initc3/HoneyBadgerMPC/builds/559202753
- Rust changes: https://travis-ci.org/initc3/HoneyBadgerMPC/builds/559695413

.travis.yml
amiller: OK This seems like a comment that needs to be added to the documentation/workflow somewher

amiller (approved):
I think I follow this strategy and it looks reasonable. Happy to approve it now

Drake-Eidukas:
Comments from Sylvain w/ comments from me:
- `docker-compose` would be preferable to use in travis
  - Caching performance doesn't work with `docker-compose`, so I recommend against it. `stager.py` currently has code that can be used with `docker-compose` as a starting point if someone wants to try to fix it via the `ComposeStager` class.
  - Main pain point lies with the `COPY pairing/ pairing/` step.
- `doc8` checks might belong in the linting stage
  - Difficult since sphinx needs docker
- It would be best for uploading of the docker images and coverage results were in their own stage
  - Although this would be better for visual clarity and transparency on the travis page, the limitations of travis re: persistent file storage make this complicated and probably not worthwhile.
  - Taking out the coverage uploading from the script could be a good idea though.
- Docs need to be updated
  - Focus on documenting travis, docker, and stager, as these are the harder parts to maintain.

sbellem (approved):
Great work! I learned new things in reviewing this PR! I like the usage of [Travis's `jobs` and `stages`](https://docs.travis-ci.com/user/build-stages/)!

Issue #315, opened 2019-05-19
Batch decoding python bypass


amiller:
Right now there's a lot of overhead in python conversion, that outweighs (by nearly a 10x factor at last estimate) the actual cost of batch reconstruction. The current workflow is essentially:
- A `ShareArray` contains a batch of GFElement objects, represented as python ints
- batch encoder `.encode()` translates each one of these into an NTL formatted object `Vec<ZZ_p>`
- The NTL format is encoded back into an array of ints (at least it's not GFElements)
- The python ints  are pickled and sent over a socket
- The socket receive is unpickled into an array of ints
- The ints are converted back into NTL objects for `interpolate`
- Back from `Vec<ZZ_p>` to python int
Clearly that's a lot of back and forth, even in the optimistic case that no flow control is needed! 

To close this issue:
- create an `OpaqueShareArray` / `OpaqueGFElementArray` object, which is backed by in memory data that is already in the exact format of a `Vec<ZZ_p>`
- `OpaqueShareArray` backing is sent directly over the socket
- batch reconstruction (both rounds) sends this backing array directly to ntl

A folllow up or related issue will be to implement (some forms of) python programs that go straight to C++ while avoiding python decoding altogether. In an ideal world, the optimistic phase bypasses python entirely except for flow control and dispute handling

PR #352, opened 2019-07-12
Batch decoding python bypass #315

chirantan31:


honeybadgermpc/ntl/rsdecode.pxd
```diff
@@ -30,3 +30,27 @@ cdef extern from "rsdecode_impl.h":
                                                           vec_ZZ_p y_vec,
                                                           ZZ_p omega,
                                                           int k, int n, int order)
+    
+<<<<<<< HEAD
```
Drake-Eidukas: Looks like you have some merge conflicts here btw

Drake-Eidukas (changes requested):
Just a few questions--
- What is the new library and why do we need it?
  - On a side note, we'll probably want to make this a git submodule. 
  - What sort of improvements does this provide?
 
Other than that I'm not too well versed in those parts of the codebase, so I'll leave reviewing that up to @amiller .

Some things to do before merging this:

- [x] Squash this down to one commit using `git rebase -i dev`
  - Make sure that you pulled in the newest version of dev
- [x] Remove remaining merge conflict fragments
- [x] Add unit tests for the new functionality (@amiller can probably give insight on that, not sure how it's done in this part of the codebase if at all)
- [x] Add a concisely headed but descriptive commit message summarizing the changes here.

Drake-Eidukas:
As for getting the tests to work on travis, this should be working very soon. In the meantime, work on those tasks.

rahulgovind:
Quickly went through the code and just had a few high-level comments/questions

- You are using variable-length encoding here where you first encode the number of limbs `s`, and then you serialize `s` limbs (which are unsigned longs?) after that. However, since the intermediate values are uniformly distributed, they will mostly occupy as many bytes as `p`. Therefore, It is not clear to me why this variable-length encoding would be better than a fixed-length encoding where you just transmit 32 bytes per ZZ_p

- It looks like you are assuming you will receive valid serialized data from the other parties which might not be the case. For example, a large value of `size` would just crash your node.  I would suggest adding additional checks while deserializing wherever possible.

chirantan31:
Closed in lieu of #353

amiller:
> Just a few questions--
> 
> * What is the new library and why do we need it?
>   
>   * On a side note, we'll probably want to make this a git submodule.
>   * What sort of improvements does this provide?
> 
> Other than that I'm not too well versed in those parts of the codebase, so I'll leave reviewing that up to @amiller .
> 
> Some things to do before merging this:
> 
> * [x]  Squash this down to one commit using `git rebase -i dev`
>   
>   * Make sure that you pulled in the newest version of dev
> * [x]  Remove remaining merge conflict fragments
> * [x]  Add unit tests for the new functionality (@amiller can probably give insight on that, not sure how it's done in this part of the codebase if at all)
> * [x]  Add a concisely headed but descriptive commit message summarizing the changes here.

I think the existing unit tests actually cover all of this. It is not actually new functionality, rather it is a different method (avoids serialization) for the existing functionality.

honeybadgermpc/ntl/rsdecode_impl.h
```diff
@@ -401,4 +407,110 @@ bool gao_interpolate_fft(vec_ZZ_p &res_vec, vec_ZZ_p &err_vec,
...
+            buf << size;
+            for(int i =0; i<size; i++){
+                buf << l[i];
+            }
+        }
+
+        template <class B>
+        void parse(B& buf) {
+            buf >> size;
+            l = new ZZ_limb_t[size];
```
rahulgovind, line 459: There should be an extra check for size here (size <= 32?). Also, is there a way to check if HPS serialization failed for any reason?

rahulgovind (commented):
The issues in https://github.com/initc3/HoneyBadgerMPC/pull/352#issuecomment-511182296 might still be relevant here.

honeybadgermpc/ntl/rsdecode_impl.h
rahulgovind: Meant to review #353. You can ignore this.

amiller: great point, i'd want to be able to catch (in python) any exception due to serialization failure

PR #353, opened 2019-07-15
Avoid conversion of ZZ_p's into PyInts during Batch Reconstruction

chirantan31:
This feature is written to address Issue #315, in the earlier flow ZZ_p elements were converted
to PytInts and then pickled to bytes before sending, and then reconverted back on receiving.
To avoid this, the HPS library is used to serialize a ZZ_p directly to and from bytes within cython.
This is accomplished by extracting the limbs of a ZZ_p (an array of longs, which is the internal
representation of a field element in ntl)


chirantan31:
Continuing from #352

amiller (approved):
I think this looks great and ready to approve

honeybadgermpc/ntl/rsdecode_impl.h
```diff
@@ -401,4 +406,110 @@ bool gao_interpolate_fft(vec_ZZ_p &res_vec, vec_ZZ_p &err_vec,
...
+            buf << size;
+            for(int i =0; i<size; i++){
+                buf << l[i];
+            }
+        }
+
+        template <class B>
+        void parse(B& buf) {
+            buf >> size;
+            l = new ZZ_limb_t[size];
```
rahulgovind, line 436: Copying over from #352. There should be an extra check for size here (size <= 32?). Also, is there a way to check if HPS serialization failed for any reason?

rahulgovind (changes requested):
The issues in #352 (comment) might still be relevant here.

honeybadgermpc/ntl/rsdecode_impl.h
amiller: Hm. Both of these comments look like they should still be fixed before merging. I especially would like to make sure that an invalid serialization actually raises an exception.

chirantan31: I'm not sure on what kind of check should I put on "size", size is the number of limbs(longs) a ZZ_p is represented by internally. We can get the value of "size" from NTL only if it's a valid ZZ_p object we're trying to serialize in the first place.

chirantan31: Also, can I have any pointers on how I should take care of the exception, if it's an invalid object we try to de-serialize HPS raises an "std::bad_alloc".
On the cython side of things should I just allow the methods to throw exceptions, as shown here
https://cython.readthedocs.io/en/latest/src/userguide/wrapping_CPlusPlus.html#exceptions

amiller: On line 457 you're parsing an untrusted string into a signed long, then allocating based on it. Isn't size at this point guaranteed to be between 0 and 16? Why use a long for that? And in any case sanity check its value

chirantan31: Oh okay, I will add that check, also why is it guaranteed to be between 0 and 16?
I use long, because NTL gives me the size as a `long` and also needs size as a `long` when converting from limbs back to ZZ_p.

amiller: I think you can figure out the maximum possible number of limbs based on the ZZ_p rep info, I forget the exact name of the class. It seems like a waste of space to use 4 bytes for that when 1 is enough, but if it helps stay consistent with the rest of ntl then that's ok

chirantan31: The maximum number of limbs could be however large, I don't see NTL having a limit on the size of the field element so no limit on the number of limbs. However, if we having a fixed modulus. I could set the max number of limbs according to that.

chirantan31: When using 
`bls12_381_r` (52435875175126190479447740508185965837690552500527637822603658699938581184513) as the modulus, the `size` is 4. So, I can check to see if `size` is under 4.

rahulgovind: > You are using variable-length encoding here where you first encode the number of limbs s, and then you serialize s limbs (which are unsigned longs?) after that. However, since the intermediate values are uniformly distributed, they will mostly occupy as many bytes as p. Therefore, It is not clear to me why this variable-length encoding would be better than a fixed-length encoding where you just transmit 32 bytes per ZZ_p

As I previously mentioned here, it's not clear to me why we need such involved serialization/deserialization here. When receiving data here we already know how many ZZ objects we need to receive and we already know which field we are going to use. So If we use a fixed-sized encoding scheme we can encode each ZZ with a fixed number of bytes (like 32 for bls12_381_r), and if you are sending over 100 ZZ objects, you can send it as 3200 bytes of data in this case. You would need to take care of a few things like endianness with this approach.

chirantan31: I may not have understood this clearly, I'll just go through what I've done and the motivation for it. Earlier how we were converting from PyInt to ZZ and vice versa was using the ZZFromBytes() method by ntl. Now instead of doing that we're getting the limbs (building blocks of a ZZ) and serializing it to bytes. Using this approach we're able to increase the speed of conversion by around ~25-30%. The reason there is a variability to the size is because it represents the number of limbs of a ZZ which depends on its magnitude.

If I'm maybe still missing something, we could get on slack and further discuss this!

amiller:
I'm going to close this until someone wants to revisit and integrate it

PR #368, opened 2019-10-01
Fixed point implementation

sanket1729:
This uses a new PR instead of #334.


honeybadgermpc/progs/fixedpoint.py
```diff
@@ -0,0 +1,248 @@
...
+    r2 = ctx.Share(r2.v * Field(2) ** m)
+
+    c = await (x + r2 + r1 + Field(2) ** (k - 1)).open()
+    c2 = int(c) % (2 ** m)
+    u = await bit_ltl(ctx, c2, r1_bits)
+    a2 = ctx.Share(Field(c2) - r1.v + (2 ** m) * u.v)
+    return a2
+
+
+async def trunc(ctx, x, k, m):
```
amiller, line 194: Could we add some comments to these functions explaining their arguments and intended input/output? It's not so self explanatory

honeybadgermpc/progs/fixedpoint.py
```diff
@@ -0,0 +1,248 @@
...
+    b: List of private bit shares. Least significant digit first
+    """
+    b_bits = [ctx.Share(Field(1) - bi.v) for bi in b_bits]
+    a_bits = [ctx.Share(ai) for ai in binary_repr(int(a), len(b_bits))]
+
+    carry = await get_carry_bit(ctx, a_bits, b_bits)
+    return ctx.Share(Field(1) - carry.v)
+
+
+async def mod2m(ctx, x, k, m):
```
amiller: Mod 2 m needs an explanation, which values are secret shares?

honeybadgermpc/progs/fixedpoint.py
```diff
@@ -0,0 +1,248 @@
...
+KAPPA = 32  # Statistical security parameter
+K = 64  # Total number of padding bits (?)
+p = modulus = Subgroup.BLS12_381
+Field = GF(p)
+
+"""
+Library for fixed point operations:
+Throughout the code:
+`x` stands for private floating point number
+`k` is the total of number of bits we are repersenting
```
amiller: representing

tests/progs/test_fixedpoint.py
```diff
@@ -0,0 +1,82 @@
...
+    async def _prog(context):
+        context.preproc = FakePreProcessedElements()
+        for i in range(0, iters):
+            a = FixedPoint(context, aval[i])
+            b = FixedPoint(context, bval[i])
+
+            assert await (await a.ltz()).open() == 0
+            assert await (await b.ltz()).open() == 1
+
+    await run_test_program(_prog, test_runner)
```
amiller, line 82: Tests all look good

amiller (changes requested):
I'm only requesting improved comments explaining the functions, which I think will make it much easier for other folks to use the routines from here

sanket1729:
Added more comments to every function

honeybadgermpc/progs/fixedpoint.py
```diff
@@ -0,0 +1,303 @@
...
+    ]
+    carry_bits.append(ctx.Share(low_carry_bit))
+    all_one_bits.append(ctx.Share(0))
+    return (await _bit_ltl_reduce(list(zip(carry_bits, all_one_bits))))[0]
+
+
+"""
+This is used in substracting numbers `a` and `b_bits`.
+a is a known public number and b_bits is a secret shared array.
+This algorithmc computes whether a + int(1^n - b) has a carry bit.
```
amiller: algorithm

honeybadgermpc/progs/fixedpoint.py
```diff
@@ -0,0 +1,303 @@
...
+    """
+    b_bits = [ctx.Share(Field(1) - bi.v) for bi in b_bits]
+    a_bits = [ctx.Share(ai) for ai in binary_repr(int(a), len(b_bits))]
+
+    carry = await get_carry_bit(ctx, a_bits, b_bits)
+    return ctx.Share(Field(1) - carry.v)
+
+
+"""
+Given the secret shared [x] calcuate the secret shares of [x//2^m] for known  public m.
```
amiller: What is //? Is that division mod p, or division with truncation? Does m have to be one of the constants above?

amiller (changes requested):
Since this is the first substantive PR in a while I want to make it good. The comments seem to be in an inconsistent style, can we look at what this file looks like when rendered for documentation? @sbellem this would be a good time to revisit documentation system

sbellem:
> Since this is the first substantive PR in a while I want to make it good. The comments seem to be in an inconsistent style, can we look at what this file looks like when rendered for documentation? @sbellem this would be a good time to revisit documentation system

Yes, will have a look. The documentation generation is in the process of being fixed as well via #388.

honeybadgermpc/progs/fixedpoint.py
```diff
@@ -0,0 +1,303 @@
...
+    Check whether the self is than is x.
+    """
+
+    async def lt(self, x):
+        return await (self - x).ltz()
+
+    async def div(self, x):
+        if type(x) in [float, int]:
+            return await self.__mul__(FixedPoint(self.ctx, 1.0 / x))
+        raise NotImplementedError
```
sbellem, line 266: ### Comments for the `FixedPoint` class.

Classes in Python 3 extend `object` by default, so you can omit it:

```python
class FixedPoint:
```

To check the type of an argument, `isinstance()` is the preferred way:

```python
 if isinstance(x, (float, int)):
```

#### About dispatching based on the type of an argument
Just as a note, since the behaviour of some methods depends on the type you may find [`functools.singledispatchmethod`](https://docs.python.org/3/library/functools.html#functools.singledispatchmethod) interesting. In order to use it we would need to upgrade to Python 3.8 though, or use an approach inspired by https://stackoverflow.com/a/24064102/2769475.

honeybadgermpc/progs/fixedpoint.py
```diff
@@ -0,0 +1,303 @@
...
+    def _binary_repr(v):
+        res = []
+        v = int(v)
+        for i in range(k):
+            res.append(v % 2)
+            v //= 2
+        return res
+
+    assert type(x) is int
+    return _binary_repr(x)
```
sbellem: Just to point out that an alternative way to get the binary repr of an integer can be done using the Python's builtin [`bin()`](https://docs.python.org/3/library/functions.html#bin) or [`format()`](https://docs.python.org/3/library/functions.html#format) such as in:

```python
def binary_repr(x, k):
    return [int(i) for i in f'{x:0{k}b}'[::-1]]
```

in a less compact form:


```python
def binary_repr(x, k):
    bin_str = f'{x:0{k}b}'
    reversed_bin_str = bin_str[::-1]
    return [int(i) for i in reversed_bin_str]
```

As for checking the type, the above will raise a `ValueError` if `x` is not an integer. If you prefer to catch the error and raise a different error:

```python
def binary_repr(x, k):
    try:
        bin_str = f'{x:0{k}b}'
    except ValueError:
        raise TypeError('x must be an integer')
     return [int(i) for i in bin_str[::-1]]
```

### About checking the type of an object
In general to check for the type of an argument `isinstance()` is preferred:

```python
assert isinstance(x, int)
```

See [PEP 8 Programming Recommendations](https://www.python.org/dev/peps/pep-0008/#programming-recommendations):

> Object type comparisons should always use [`isinstance()`](https://docs.python.org/3/library/functions.html#isinstance) instead of comparing types directly.
>
> Yes: if `isinstance(obj, int):`
> 
> No:  `if type(obj) is type(1):`

### About using `assert`
It is important to keep in mind that if optimization is required (via [`-O`](https://docs.python.org/3/using/cmdline.html#cmdoption-o) or [`PYTHONOPTIMIZE=1`](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONOPTIMIZE) `assert` statements will be removed:

> The current code generator emits no code for an assert statement when optimization is requested at compile time.

source: https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement

`assert` statements are definitely useful to check for something that should not occur when troubleshooting so it depends on the intent.

sanket1729:
Ready for review again

amiller (approved):

PR #372, opened 2019-10-02
Update hbmpc-tutorial-1.py

hugo-tardiou:
ctx.preproc.get_triple does not exist while ctx.preproc.get_triples does

-- Hugo


amiller:
Thank you! We're having trouble with the .ci but should be able to take this patch

sanket1729 (approved):

sanket1729:
Thanks for the contribution,

PR #376, opened 2019-10-17
NTL fix

sanket1729:


amiller (commented):
This looks fine by inspection. Clearly it still works with the .ci though I'm not sure how to check that it has the intended effect in other environments

amiller (approved):

PR #386, opened 2019-11-02
Update formatting

sbellem:
* Update code format to match latest `black` requirements as it has received some [updates](https://github.com/psf/black#change-log) since we started using it.
* Update `black` repo url in pre-commit config (repo was moved under the Python Software Foundation Github org (@psf)).
* Put back `black` in `setup.py`.

PR #384, opened 2019-11-01
Fix codecov upload

sbellem:
~~Will only be ready after #386 is merged.~~

**UPDATE**: #386 has been merged. Now ready if all checkpoints (Travis CI, Codecov) pass.

This will allow us to re-introduce the code coverage check.

Issue #382, opened 2019-11-01
Review configuration mechanism


sbellem:
The current configuration mechanism has some strength and weaknesses. One of its current strengths appears to be that it is very convenient as it happens automatically in the root `__init__.py`:

https://github.com/initc3/HoneyBadgerMPC/blob/0914c0269a5d3636b01f40557e8ea432f721bd04/honeybadgermpc/__init__.py#L23

This comes at a cost though: building documentation that has docstrings fails (see #378). Perhaps it could also be a cause to other problems in the future but for now that seems to be the only problem. So one may object that it is a very low cost in exchange for the convenience it provides. Perhaps. Nevertheless, conceptually speaking it seems that we could improve the current mechanism in place to not only restore the capability of building documentation but to also make the configuration mechanism clearer, more explicit, and re-usable as needed in the modules that require it.

### Rough Planning Notes
- [x] Review and understand how the current mechanism works.
- [x] Identify what commands/entrypoints require loading the configuration from a file.
- [x] Clarify whether there is one or multiple entrypoints to the system. Does one always starts the system via the same entrypoint or is the current code base more like a framework/toolbox with multiple independent component each having its one "entry point" but nevertheless sharing the same configuration object/mechanism.
- [x] Implement a different mechanism so that #378 can be resolved.
~~- [ ] Address #222.~~

### Observations
The main observation one may make is that the "[loading config](https://github.com/initc3/HoneyBadgerMPC/blob/dev/honeybadgermpc/config.py#L64)" code that reads command line arguments does not need to be invoked by default in the project root `__init__.py`. 

### Questions
* Would it make sense to transfer the responsibility of [`load_config`](https://github.com/initc3/HoneyBadgerMPC/blob/dev/honeybadgermpc/config.py#L64) to a more general command line argument?

* Would it make sense to have one top command, say `hbmpc` that can take in subcommands? For example: `hbmpc run appname ...`.

* If the current code base has multiple standalone "apps" each with its own command that depends on the current [`HbmpcConfig`](https://github.com/initc3/HoneyBadgerMPC/blob/dev/honeybadgermpc/config.py#L54) then would it make sense to have an overall general command line interface that includes the option of loading a config as it is currently done, and each "app" that requires its own command line could inherit from the general command line interface with the already existing configuration mechanism "baked in". This approach would achieve a very similar convenience as the current approach but would limit the loading of the config to actual cases where it is indeed needed. For non-command line cases a default config may do the work.

[...]

PR #388, opened 2019-11-02
Explicit configuration

### LINKED ISSUE #378: Fix sphinx-build error to generate docs
The error:

```shell
root@15a832adeb00:/usr/src/HoneyBadgerMPC# make -C docs html
usage: sphinx-build [-h] [-d] [-f CONFIG_FILE_PATH]          
sphinx-build: error: unrecognized arguments: -M html . _build
```

This perhaps is caused by a combination of using `autodoc` to generate documentation from the docstrings and not having an `if __name__ == "__main__"` clause in a python module that could be run as a script. So the problem seems to be with the following code:

https://github.com/initc3/HoneyBadgerMPC/blob/6aec65f9a1e430316077bc98d548bda42571b4b3/honeybadgermpc/__init__.py#L16-L18

Now adding the `if` clause as in:

```python
if __name__ == "__main__":

    # Skip loading the config for tests since the would have different values for sys.argv.
    if "pytest" not in sys.modules:
        HbmpcConfig.load_config()
```

works as far as building the docs but it creates problems for actually running the code as in the case of tutorial 2 for instance, because the config does not get loaded as it is expected to be:

```shell
$ scripts/launch-tmuxlocal.sh apps/tutorial/hbmpc-tutorial-2.py conf/mpc/local
WARNING: the $CONFIG_PATH environment variable wasn't set. Please run this file with `scripts/launch-tmuxlocal.sh apps/tutorial/hbmpc-tutorial-2.py conf/mpc/local`
```

So it seems that to properly solve this issue the logic that handles the loading of the config may need to be reviewed (#382). The main observation one may make is that the "loading config" code that reads command line arguments does not need to be invoked by default in the project root `__init__.py`. Doing so is obviously convenient, so the proposed solution should aim at preserving the convenience of the current approach meanwhile resolving the problem it causes for generating documentation.

sbellem:
Addresses #382 and resolves #378.

This PR is for commit 139a149. The other commits belong to #386 and #384.

This is to be merged only after #386 and #384.

### DIFF
--- a/apps/asynchromix/butterfly_network.py
+++ b/apps/asynchromix/butterfly_network.py
@@ -84,6 +84,8 @@ async def _run(peers, n, t, my_id):
 if __name__ == "__main__":
     from honeybadgermpc.config import HbmpcConfig
 
+    HbmpcConfig.load_config()
+
     k = int(HbmpcConfig.extras["k"])
 
     pp_elements = PreProcessedElements()
--- a/apps/tutorial/hbmpc-tutorial-2.py
+++ b/apps/tutorial/hbmpc-tutorial-2.py
@@ -55,6 +55,8 @@ if __name__ == "__main__":
     from honeybadgermpc.config import HbmpcConfig
     import sys
 
+    HbmpcConfig.load_config()
+
     if not HbmpcConfig.peers:
         print(
             f"WARNING: the $CONFIG_PATH environment variable wasn't set. "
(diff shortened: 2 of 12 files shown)


sbellem:
Pull requests must be made from `initc3` org otherwise `docker` actions such as pushes do not work.

> Please note that encrypted environment variables are not available for [pull requests from forks](https://docs.travis-ci.com/user/pull-requests#pull-requests-and-security-restrictions).

-- https://docs.travis-ci.com/user/encryption-keys/

Will close this one and re-open from `initc3` repo.

PR #398, opened 2019-11-19
Remove unused equality; print errors on mixin failure

amiller:
This patch
a) gets rid of duplicate code for equality testing
b) makes sure that calls to mixins, which create a new task, print exceptions when they fail for easier debugging

Thanks to @amitgtx for encountering this!


sanket1729 (approved):
Ack

Issue #400, opened 2019-11-26
[rust/python] Building pairing lib fails


sbellem:
`pip install pairing/` currently fails. See below for the error output. The current quickest solution is to "pin" the nightly build to a recent working version such as `nightly-2019-11-15`. But it would be useful to figure out a better solution for the near future.

### Some observations
It may be useful to try to stay up-to-date with `pyo3` as they are fixing issues arising from changes to `rust` (e.g.: https://github.com/PyO3/pyo3/pull/615).

* Current pyo3 version in pairing: [0.4](https://github.com/PyO3/pyo3/releases/tag/v0.4.1) (Aug 20, 2018)

https://github.com/initc3/HoneyBadgerMPC/blob/350bfb1cf49491b5d9e40cc5c3140468fa699d68/pairing/Cargo.toml#L25-L26

* Latest pyo3 release version: [0.8.3](https://github.com/PyO3/pyo3/releases/tag/v0.8.3) (Nov 22, 2019)

### Error ouput
```shell
   error[E0277]: the trait bound `(): err::PyErrArguments` is not satisfied
       --> /root/.cargo/registry/src/github.com-1ecc6299db9ec823/pyo3-0.4.1/src/err.rs:521:64
        |                                       
    521 |                 PyErr::from_value::<$pyexc>(PyErrValue::ToArgs(Box::new(err)))
        |                                                                ^^^^^^^^^^^^^ the trait `err::PyErrArguments` is not implemented for `()`
    ...
    590 | impl_to_pyerr!(std::string::ParseError, exc::ValueError);
        | --------------------------------------------------------- in this macro invocation
        |
        = note: required for the cast to the object type `dyn err::PyErrArguments`
    
    error: aborting due to previous error
    
    For more information about this error, try `rustc --explain E0277`.
    error: could not compile `pyo3`.
    
    Caused by:
      process didn't exit successfully: `rustc --crate-name pyo3 /root/.cargo/registry/src/github.com-1ecc6299db9ec823/pyo3-0.4.1/src/lib.rs --error-format=json --json=diagnostic-rendered-ansi --crate-type lib -$
emit=dep-info,metadata,link -C opt-level=3 --cfg 'feature="default"' --cfg 'feature="extension-module"' --cfg 'feature="python3"' -C metadata=b089b812af986789 -C extra-filename=-b089b812af986789 --out-dir /tmp/p$
[...]

Issue #209, opened 2019-03-05
Undo using a specific commit of charm


smkuls:
As part #208, we had switched to a specific commit of charm. This was because the latest code was breaking the build. Ideally we should be using a `release` of charm but they haven't released a new version in ages.

amiller:
We're probably going to have to close this as non-actionable

Issue #100, opened 2018-10-29
Reproducible research


sbellem:
_draft_

This is an issue to note down required details to ensure that experiments along with their results can be reproduced by others. For instance, even though the code may be open source and one may provide the commit hash of the code that was used to conduct an experiment this may not guarantee that the experiment will be reproducible if different versions of some dependencies are used.

To help making things clearer: THE CODE means everything that is captured by a commit hash.

Here's a tentative list of things THE CODE of an experiment should provide:

- [ ] `Pipfile` and `Pipfile.lock` (see https://pipenv.readthedocs.io/en/latest/basics/#pipenv-lock)
- [x] Dockerfile(s)

amiller:
Here's a general concept to mention:
- when we do benchmarking, and read measurements from log file printouts (latency, bytes measured, etc.) we should make sure to save the raw files, commit them to a git repository of some kind
- we generally want to do analysis on a spreadsheet of just the most interesting results basically, so one of the steps is to by script (or by hand) parse the log file and copy out the relevant fields. However, we should not just delete the raw files we read, since otherwise we would waste an opportunity to catch or double check mistakes in transcription

sbellem:
> * when we do benchmarking, and read measurements from log file printouts (latency, bytes measured, etc.) we should make sure to save the raw files, commit them to a git repository of some kind
> 
> * we generally want to do analysis on a spreadsheet of just the most interesting results basically, so one of the steps is to by script (or by hand) parse the log file and copy out the relevant fields. However, we should not just delete the raw files we read, since otherwise we would waste an opportunity to catch or double check mistakes in transcription

Just some thoughts (_copied from the discussion on Slack_):

## For now
What if after each run a script (python or shell) was run to simply push the log files to a dedicated github repo? The commit message could contain the commit hash of the code that produced the log files. That should be relatively easy and quick to setup. The files could be compressed and archived in a tar ball before being committed to save space if needed.

I think that could be a quick and fast first improvement.

## For the near future
In a second iteration, I would look into whether the logs are structured or not and if not would structure them using, one of:

1. https://www.structlog.org/
2. https://docs.python.org/3/howto/logging-cookbook.html#implementing-structured-logging

and would automate the parsing and start looking into persisting them to some kind of datastore that is good for that. Eventually this approach probably would start mimicking what the ELK stack provides ... 

## For the long term
[...]

PR #401, opened 2019-11-27
Remove the need of docker credentials on Travis CI

### LINKED ISSUE #387: Running Travis CI jobs on forks & PRs from forks
**Tl;dr**: Without write access to Docker Hub, the code that is tested will not be the code from the PR. Write access to Docker Hub is reserved to PRs made from a branch under **initc3/HoneyBadgerMPC**. 

> Please note that encrypted environment variables are not available for [pull requests from forks](https://docs.travis-ci.com/user/pull-requests#pull-requests-and-security-restrictions).

-- https://docs.travis-ci.com/user/encryption-keys/

---

**UPDATE**: This issue was originally opened to address the problem of running Travis CI on one's own fork as this may be very useful for multiple reasons. But there is also the problem of being able to accept pull requests from forks. The two problems are somewhat related because in both cases Travis CI cannot access the encrypted credentials and therefore cannot login to Docker Hub and therefore cannot push built images which in the end results in Travis CI not being able to actually test the actual code of the PR, unless some changes are made.

Currently the two jobs under the **`Test`** stage, `Sphinx Tests` and `Unit Tests`:

image

depend on a docker image that is pulled from Docker Hub. That image is expected to have been pushed to Docker Hub during the previous **`Build`** stage.

Pull requests that are from forks or branches that are from forks cannot push to Docker Hub during the **`Build`** stage and consequently the image that is pulled from Docker Hub in the **`Test`** stage is from a previous merged PR not from the current PR.

# Problem & Possible Solutions
It may be worthwhile to review the motivations behind the current situation. As far as I know the main motivation was to speed the build phase on Travis CI, and also for local development.

**Problem**: Travis CI builds take too long.
**Solution**: Use Docker Hub as a caching system.

The current approach in place works well as long as PRs are from initc3/HoneyBadgerMPC. This leads to a new problem:

**Problem**: PRs must be from initc3/HoneyBadgerMPC, and more generally, Travis CI can only be used for branches under initc3/HoneyBadgerMPC. This limits contributions to people who have push access to initc3/HoneyBadgerMPC.

## Solution 1: Limit external PRs to specific files
That is, limit external PRs to Python code modifications and/or files that do not require to be compiled/built. Mount those modifiable files into the container to actually test against these files. Detect changes to "non-modifiable" files using `git diff` and `TRAVIS_COMMIT_RANGE`. This requires maintaining a list of files that most not be modified. If an external PR modifies such files the PR should fail.

**PROs**
* Quick to implement (at least that is the current guess)

**CONs**
* Requires maintaining a list of non-modifiable files to check against for external PRs.
* Prone to errors as it is easy to imagine the "black list" being out-of-date.
* Prevents external contributors from making contributions to other parts of the code or would require a different process in place, in which a member of initc3/HoneyBadgerMPC would somehow need to bring in the changes. 

## Solution 2: Use separate repositories
Move the code for the base layer into one (or more) separate repository and version that layer such that at any point in time one working on the HoneyBadgerMPC Python code will work against a versioned "base". Modifications to that "base" would need to be done in that dedicated base repository.

**PROs**
* Allows everyone to contribute on all parts of the code.
* Invites for a refinement of the modularization of the current code base. For instance, perhaps `pairing` could be in its repo, and the apps as well.
* Invites for having a base docker image that has the most basic dependencies such as `rust`, `flint`, `ntl`, `pbc` and `charm` (i.e.: [line 25-130 of the Dockerfile](https://github.com/initc3/HoneyBadgerMPC/blob/dev/Dockerfile#L25-L130)).

**CONs**
* Implementation effort more time-consuming (_perhaps_).
* May complicate the development workflow for those needing to modify various layers of the system.

# Some Details to Keep in Mind
It seems to me that it may be very useful to discern between dependencies or prerequisites that do not depend at all on the actual code found under the HoneyBadgerMPC repository. **`Flint`** and **`NTL`** are examples of such dependencies. On the other hand, building **`pairing`** or **`apps/asynchromix/cpp`** depend on the actual code found under HoneyBadgerMPC. This point is mentioned because dependencies that do not rely on the HoneyBadgerMPC code can be pre-built in a totally separate context outside the HoneyBadgerMPC context. 

## Separation of Concerns
The current setup addresses various concerns, such as image size, build speed, development environment, and testing on Travis CI. It may be useful, to think of each concern separately and to see what kind of solution is best. This thought exercise should not prevent to have an integrated solution in the end if that is what is best.

### Concern 1: Travis CI
Travis CI is used to run tests, build documentation and run code and documentation quality checks. Travis CI results and code reviews are used to help validate pull requests. Once a pull request has passed all checks on Travis Ci and has been positively reviewed it may be merged by someone who has write privileges to the repository.

### Concern 2: Managing the Development Lifecycle
One key issue that arises with managing the dev environment has to do with binaries and volume mounts (see #180 for instance).

Another issue, is that one needs to be able to relatively easily make changes to whatever part of the system, be it pure Python, C, C++, or Rust -based optimizations. Moreover, a developer needs to be capable to quickly re-build the system after modifications made to whatever part of the system. With pure Python this is somewhat trivial whereas with code that requires compilation a more substantial toolchain may be required.

**Requirements**:
1. The code under test must be the same as the code under review.
2. Make the builds fast.
3. Ideally support pull requests made from forks; from external contributors.

Currently, the above 3 are not possible. Builds are always fast but when they are from forks it is at the expense of (1) :fearful: .

# Possible Improvements to Dockerfile
- [ ] Use the Docker image of rustlang for nightly builds: https://hub.docker.com/r/rustlang/rust/.
- [ ] To build `pairing` use a pre-built image with a working version of `pyo3`. In order to do this look into setting up a docker image on dockerhub that will include pyo3 and will get built automatically when the nightly rust changes, but if the the overall build with pyo3 fails then the image on dockerhub should remain with a previously working nightly rust. In other words the nightly rust version in use should work with pyo3 and pairing. HoneyBadgerMPC image would then depend on this "working" build. Doing so, if possible would prevent cases in which building no longer works, and one needs to then manually downgrade the rust nightly version.
- [ ] For Travis CI it should not be required to install the project in development mode (with the `-e` or `--editable` option when using `pip` or equivalently using `python setup.py develop`. Currently if the editable is not used `ntl` cannot be imported. (`ModuleNotFoundError: No module named 'honeybadgermpc.ntl._hbmpc_ntl_helpers'` -- note that this problem is "strange" -- todo: provide example.)

---

**Original text for this issue**:
This happens when Travis CI runs the build on a branch of a fork. To work on this issue requires clarifying whether the current build process can actually be executed on a fork. Having a build process that is "reproducible" on forks may be useful and desirable as it can allow others to execute the builds/tests under their own chosen CI environment.

Example (https://travis-ci.org/sbellem/HoneyBadgerMPC/jobs/606265339#L177):

```shell
Setting environment variables from .travis.yml
$ export 9ak80ojFZdSe1ZHfJlG7cQQs9q8KTh4Y558YbkciXtw=[secure]
We were unable to parse one of your secure environment variables.
Please make sure to escape special characters such as ' ' (white space) and $ (dollar symbol) with \ (backslash) .
For example, thi$isanexample would be typed as thi\$isanexample. See https://docs.travis-ci.com/user/encryption-keys.
```

# Resources
* [Docker development best practices](https://docs.docker.com/develop/dev-best-practices/)
* [Use bind mounts](https://docs.docker.com/storage/bind-mounts/)
* [Best practices for writing Dockerfiles](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/)
* [Multi-stage builds in Docker](https://docs.docker.com/develop/develop-images/multistage-build/)
* [Managing Open-Source Docker Images on Docker Hub using Travis](https://medium.com/vaidikkapoor/managing-open-source-docker-images-on-docker-hub-using-travis-7fd33bc96d65)
* https://scikit-learn.org/stable/developers/advanced_installation.html
* https://github.com/pypa/packaging.python.org/issues/320

sbellem:
* Move base dependencies into separate docker image
* Resolves #387
---
**Main points**:
* Dependencies that _do not_ rely on the code base are packed into one docker image that will be maintained in an external repository.
* For the time being this parent [Dockerfile](https://github.com/sbellem/docker-honeybadgermpc-deps/blob/master/python3.7/buster/slim/Dockerfile) (for dependencies) is under [docker-honeybadgermpc-deps](https://github.com/sbellem/docker-honeybadgermpc-deps) on Github and the image is on Docker Hub at [honeybadgermpc-deps](https://hub.docker.com/repository/docker/sbellem/honeybadgermpc-deps/).
* Either before or right after this PR is merged the Dockerfile and image should be moved under **initc3** or **dsluiuc**.

Some of the dependencies for the external [honeybadgermpc-deps](https://hub.docker.com/repository/docker/sbellem/honeybadgermpc-deps) image, have been packaged into their own image:

* [FLINT](https://hub.docker.com/repository/docker/sbellem/flint2)
* [NTL](https://hub.docker.com/repository/docker/sbellem/ntl)
* [PBC](https://hub.docker.com/repository/docker/sbellem/pbc)
* [Charm-crypto](https://hub.docker.com/repository/docker/sbellem/charm-crypto)
* [Rust nightly frozen](https://hub.docker.com/repository/docker/sbellem/rust-frozen)

**Some positive side effects**:
* The problem with FLINT (https://github.com/wbhart/flint2/issues/217 / https://github.com/wbhart/flint2/issues/512) is resolved, as we can use the latest code that is on Github. 
* The problem with building pairing (#400) is taken care of by providing a frozen nightly build.
* The problem with Charm mentioned in #209, appears to be resolved with a more recent version of Charm. The image uses the latest code of Charm and the code could build and tests pass.
* **Precision**: The proposed approach should allow for much control over dependencies with respect to which versions are being used. This may help with the "reproducible" research concern (#100).
* **Speed**: Once a required specific version of a dependency is packed into an image it is not needed to re-build it as one simply copies the necessary files (e.g.: `.h` and `.so`) from the pre-built image. This can be done in the image used for development. 

**TODO**:

- [x] Explain the proposed workflow.
- [x] Explain how one would go about to experiment with a new dependency and eventually add the new dependency.
- [x] Explain how one would upgrade or downgrade a dependency.
- [x] Use **initc3** or **dsluiuc** as base org for docker base image `honeybadgermpc-deps`. I have used my account for the time being.
- [x] Document how to work with the images.
- [x] (`pip`) Install python dependencies for `tests` only for the unit test job, and `docs` only for the documentation job. Currently both `tests` and `docs` are installed for both jobs.

Will remove the `docker_*.sh` scripts in a subsequent step as they may be useful for build jobs done under the `initc3/HoneyBadgerMPC` repo, i.e.: `TRAVIS_REPO_SLUG == initc3/HoneyBadgerMPC`.


sanket1729:
Hi @sbellem, 
   This looks super clean and thanks for the prompt work. We would like to merge this as soon as possible. 

1) Can we still use Drake's worksflow for the second repo? That is, we should auto push to docker hub whenever the base image is updated. 
2) Can you also remove the scripts docker_*.sh inside the scripts folder as part of this PR. 
3) I suppose this will require an edit in the README for install instructions too.

sbellem:
> Can we still use Drake's workflow for the second repo? That is, we should auto push to docker hub whenever the base image is updated.

Docker Hub allows for [automated builds](https://docs.docker.com/docker-hub/builds/) to be configured such that any change on the linked Github repo will trigger a new build of the image on Docker Hub. 

I will look into the optimal way to [set this up for the team](https://docs.docker.com/docker-hub/builds/#autobuild-for-teams). It is also possible to set automated tests via what Docker calls [`Autotest`](https://docs.docker.com/docker-hub/builds/automated-testing/) and I also want to look into this.

amiller (approved):
We discussed this online and looks reasonable!

sanket1729:
I think before we merge this, we should at least update the install instructions

sbellem:
> I think before we merge this, we should at least update the install instructions

I'll be adding documentation early on Tuesday, now that the images have been moved under the initc3 org on both Github and Docker Hub.

sanket1729 (approved):
Awesome. Great job. One small question. Will we have to manually change the digest `HBMPC_DEPS_DIGEST` every time there is a new build image.

If yes, can we look into some automated way of doing this? It is not a priority thing, but just "good to have" thing

sbellem:
> Will we have to manually change the digest `HBMPC_DEPS_DIGEST` every time there is a new build image.

Yes, every time we wish to use a different version of the parent image.
  
> If yes, can we look into some automated way of doing this? It is not a priority thing, but just "good to have" thing

I would need to look into it. Perhaps with a [custom build phase hook](https://docs.docker.com/docker-hub/builds/advanced/#custom-build-phase-hooks) after the push (`hooks/post_push`).

If we change the digest the change should go through a CI phase, like on Travis CI, or perhaps using Docker Hub [`Autotest`](https://docs.docker.com/docker-hub/builds/automated-testing/#enable-automated-tests-on-a-repository).

An alternative would be to use tags instead of digests. Something like:

 ```dockerfile
FROM initc3/honeybadgermpc-deps:python3.7-slim-buster
```
instead of 

 ```dockerfile
FROM initc3/honeybadgermpc-deps@sha256:46902d869ea881d7b00b72ff6accf2558a5e15849da5fa5cc722b4ff82a507f8
```

Tags are mutable whereas digests are immutable. In other words, when you are pulling an image by its tag that image may change from one day to another whereas using the digest provides a guarantee that this image will not change.

So, if  `initc3/honeybadgermpc-deps:python3.7-slim-buster` has a digest equal to `46902d869ea881d7b00b72ff6accf2558a5e15849da5fa5cc722b4ff82a507f8` right now, and tomorrow `initc3/honeybadgermpc-deps:python3.7-slim-buster` is updated its digest will be different, so one could get the latest image by using the tag.

The advantage with using the image digest is that it helps with reproducible results, such as benchmarks and tests.
 
You may find this [blog post](https://engineering.remind.com/docker-image-digests/) interesting to read.

PR #406, opened 2019-12-04
Volume matching

lilione:
Implement the volume matching algorithm and test cases for it.

### DIFF
--- a/apps/auctions/volume_matching.py
+++ b/apps/auctions/volume_matching.py
@@ -1,0 +1,346 @@
+# Original code was written by Harjasleen Malvai
+
+"""
+In volume matching auction, buy and sell orders are matched only on volume
+while price is determined by reference to some external market.
+"""
+
+import asyncio
+import logging
+
+from honeybadgermpc.mpc import TaskProgramRunner
+from honeybadgermpc.preprocessing import (
+    PreProcessedElements as FakePreProcessedElements,
+)
+from honeybadgermpc.progs.fixedpoint import FixedPoint
+from honeybadgermpc.progs.mixins.share_arithmetic import (
+    BeaverMultiply,
+    BeaverMultiplyArrays,
+    MixinConstants,
+)
+
+config = {
+    MixinConstants.MultiplyShareArray: BeaverMultiplyArrays(),
+    MixinConstants.MultiplyShare: BeaverMultiply(),
+}
+
+
+async def compute_bids(ctx, balances, bids, price):
+    """Compute all valid bids for each user.
+    According to current market price, a bid becomes invalid
+    if the user doesn't have enough balance to pay for this bid.
+    We only keep valid bids and classify them into
+    buy bids(volume > 0) and sell bids(volume < 0).
+    Invalid bids will be discarded after executing this function.
+
+    Parameters
+    ----------
+    balances : dict of dict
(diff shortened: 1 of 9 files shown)


sanket1729:
We can merge fixed point first. Then we can get this in

sanket1729:
Would require a rebase now

apps/auctions/volume_matching.py
```diff
@@ -0,0 +1,243 @@
...
+
+        used_balances[addr]["erc20"] = used_balances[addr]["erc20"] + sell_vol
+        used_balances[addr]["eth"] = used_balances[addr]["eth"] + (
+            await buy_vol.__mul__(price)
+        )
+
+    return buys, sells
+
+
+async def volume_matching(ctx, bids):
```
amiller: This is missing comments explaining what this function interface is. What is "bids" supposed to contain? What is the type of it? Are there preconditions? What function does this compute?

amiller: If possible try to follow the documentation format guidelines in our contributing guidelines: 
https://github.com/initc3/HoneyBadgerMPC/blob/dev/docs/development/contributing.rst#documentation-conventions

amiller (changes requested):
Before I can review it further, this is missing any comments or documentation explaining what the code does

amiller:
We should try to use this PR as an example of the documentation style we want to see in contributed apps

sbellem:
@lilione when you have a chance, could you rebase the PR so that CircleCI will be used to generate the docs and we can then preview them ...

sbellem:
@yunqi I added some commits so that you can build and view the docs locally. To build and view the docs in your browser you can run:

```shell
$ make servedocs
```

To build the docs:

```shell
$ make docs
```

You then need to refresh the browser.

To view the auctions app docs go to http://localhost:58888/apps/auctions.html.

Let me know once you are setup!

apps/auctions/volume_matching.py
```diff
@@ -0,0 +1,340 @@
...
+    """Compute all valid bids for each user.
+    According to current market price, a bid becomes invalid
+    if the user doesn't have enough balance to pay for this bid.
+    We only keep valid bids and classify them into
+    buy bids(volume > 0) and sell bids(volume < 0).
+    Invalid bids will be discarded after executing this function.
+
+    Parameters
+    ----------
+    balances : {address -> {cointype -> balance}}
```
sbellem: For the type, I think the convention is usually to specify the type, such as `list`, `dict`, `str`, etc. From the [numpy docstring docs](https://numpydoc.readthedocs.io/en/latest/format.html#sections):


> For the parameter types, be as precise as possible. Below are a few examples of parameters and their types.
>
> ```rst
> Parameters
> ----------
> filename : str
> copy : bool
> dtype : data-type
> iterable : iterable object
> shape : int or tuple of int
> files : list of str
> ```


So in this case that would be `dict of dict`.

In the description, details about the keys and values can be added, such as keys are addresses, and values are dictionaries in which the keys are coin types and the values are balances ...

Just as a rough example.

sbellem (approved):

sbellem:
This is good to merge. The app works (`python apps/auctions/volume-matching.py`) and the documentation follows the numpy docstring style.

/cc @amiller

PR #409, opened 2019-12-08
Netting Project: PR 1 of n

sanket1729:
The project will probably have lots of PRs coming the way. I am making small PRs for fast development and easy review/merges. 

Although this PR does not add any functionality by itself, I think merging this will save development and review time in the future. Let me know if you have any issues with this approach. And we can decide on another way of doing this. 

I am also trying to maintain this project as a standalone project without reliance on this codebase, but I don't think that effort is warranted for now. I am using a custom .gitignore file for this project. 

As of now, this PR supports:

- [x] Creation of transaction and balances from initial values

- [x] Client reading the data from files and saving in-memory structures.

- [x] Tests for the correct functioning of these files

### DIFF
--- a/apps/netting/config/config.py
+++ b/apps/netting/config/config.py
@@ -1,0 +1,22 @@
+# Number of transactions
+NUMTX = 500000
+
+# Number of parties
+NUM_CLIENTS = 64
+
+# Average payment amount
+mu_tx = 10000.00
+mu_bal = 10000.00
+
+# Standard deviation
+sigma_tx = 100.00
+sigma_bal = 100.00
+
+# File name info
+tx_file_prefix = "party_"
+tx_file_suffix = "_tx_data.csv"
+
+bal_file_prefix = "party_"
+bal_file_suffix = "_bal_data.csv"
+
+NETTING_BASE_DIR = "/usr/src/HoneyBadgerMPC/apps/netting/"
--- a/apps/netting/makefile
+++ b/apps/netting/makefile
@@ -1,0 +1,7 @@
+all: clean gen_data
+
+clean:
+	rm -rf data/
+
+gen_data:
+	python3 scripts/generate_data.py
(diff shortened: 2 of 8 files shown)


sanket1729:
Tracked in #410

apps/netting/makefile
```diff
@@ -0,0 +1,7 @@
+all: clean gen_data
+
+clean:
+	rm -rf data/
+
+gen_data:
+	python3 scripts/generate_data.py
```
sbellem, line 7: Just to remain consistent with the code base, let's use `Makefile` instead of `makefile`. Although both work, the [GNU `make`](https://www.gnu.org/software/make/manual/make.html#Makefile-Names) manual recommends `Makefile`:

> We recommend `Makefile` because it appears prominently near the beginning of a directory listing, right near other important files such as `README`.

apps/netting/config/config.py
```diff
@@ -0,0 +1,22 @@
...
+sigma_bal = 100.00
+
+# File name info
+tx_file_prefix = "party_"
+tx_file_suffix = "_tx_data.csv"
+
+bal_file_prefix = "party_"
+bal_file_suffix = "_bal_data.csv"
+
+NETTING_BASE_DIR = "/usr/src/HoneyBadgerMPC/apps/netting/"
```
sbellem, line 22: Is there a plan to have more files under `apps/netting/config/`? If not, `apps/netting/config.py` is sufficient and makes the code simpler.

apps/netting/scripts/generate_data.py
```diff
@@ -0,0 +1,105 @@
...
+    close_files(f)
+
+
+def clean_files():
+    path = NETTING_BASE_DIR + "data/"
+    try:
+        shutil.rmtree(path)
+    except Exception as e:
+        logging.exception(e)
+        pass
```
sbellem, line 93: The `pass` is not needed :smile: .

amiller:
Closing as we're working on another branch

PR #416, opened 2019-12-16
Apply flake8 import order

### LINKED ISSUE #392: Require import style as per PEP 8
[PEP 8](https://www.python.org/dev/peps/pep-0008/#imports) has some guidelines for organizing the imports. One of the key things is:

> Imports should be grouped in the following order:
>
>    Standard library imports.
>    Related third party imports.
>    Local application/library specific imports.
>
> You should put a blank line between each group of imports.

This structure eases code reviews and provides some consistency for the code base.

You can read more about the reasoning behind the [Python Style Guide (PEP 8)](https://www.python.org/dev/peps/pep-0008/) in the section: [A Foolish Consistency is the Hobgoblin of Little Minds](https://www.python.org/dev/peps/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds). 

### Adding the check in Travis CI
The style checker [`flake8-import-order`](https://github.com/PyCQA/flake8-import-order) plugin from the [Python Code Quality Authority](https://github.com/PyCQA) can be used so that the `flake8` check on Travis CI will also check the import style.

The plugin supports multiple [styles](https://github.com/PyCQA/flake8-import-order#styles). The style used by the [`cryptography`](https://github.com/pyca/cryptography) project from the [Python Cryptographic Authority](https://github.com/pyca) is the default (see [example](https://github.com/PyCQA/flake8-import-order/blob/master/tests/test_cases/complete_cryptography.py)). This style is more strict than the [`pep8` style](https://github.com/PyCQA/flake8-import-order/blob/master/tests/test_cases/complete_pep8.py), which "only enforces groups without enforcing the order within the groups". For instance, running the check on the current code base returns **399** errors with the [default (`cryptography`) style](https://github.com/PyCQA/flake8-import-order/blob/master/tests/test_cases/complete_cryptography.py) and **191** with the [pep8 style](https://github.com/PyCQA/flake8-import-order/blob/master/tests/test_cases/complete_pep8.py).

### Editor plugins
Some documentation should be added about adding the plugin to the text editor to help developers catching up the errors in their text editor.

sbellem:
This will help having a consistent style and structure for imports.

The style used is the default one which is the one referred to as
`cryptography`.

See https://github.com/PyCQA/flake8-import-order for more information.

Resolves #392

**Editors**
`flake8-import-order` is added to `setup.py` and re-installing the project (e.g.: `docker-compose build` or `pip install -e .[dev]`) should make the plugin available to your editor.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -5,3 +5,4 @@ exclude =
 	.eggs/,
         apps/tutorial/
 ignore = E203, E266, E501, W503
+application_import_names = apps,honeybadgermpc,lib_solver
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,8 +13,8 @@ jobs:
     install: pip install black
     script: black --check .
   - name: Flake8
-    install: pip install flake8 pep8-naming
-    script: flake8
+    install: pip install flake8 pep8-naming flake8-import-order
+    script: flake8 --count
   - name: Doc8
     install: pip install doc8 pygments
     script: doc8 docs/
--- a/apps/asynchromix/asynchromix.py
+++ b/apps/asynchromix/asynchromix.py
@@ -1,34 +1,35 @@
 """
 Implementation of Asynchromix MPC Coordinator using an EVM blockchain
 """
-from web3 import Web3, HTTPProvider
-import time
 import asyncio
 import logging
-from contextlib import contextmanager
+import os
 import subprocess
+import time
+from contextlib import contextmanager
+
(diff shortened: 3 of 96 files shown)


amiller (approved):
Concept ack. Adding new things to flake8 seems to be necessary if we add new modules, e.g. even just in the apps/ directory, that may need to be imported. So there should be something about this in our dev docs or readme

PR #429, opened 2020-03-18
WIP: Hello Shard! Intershard communication example app

sbellem:
The relevant commit is 5941750da4d17e7b94d000173d1991b524a40801

_Far from elegant and very rudimentary first version that can for sure benefit from many improvements._

**Some of the current limitations are:**
1. The client sends messages to a gateway shard, meanwhile the other shard only receives masked messages from this gateway shard.
2. The complete message transmission, from client to gateway shard to "receiving" shard all happens within an epoch.
3. The generation of preprocessing elements for the intershard transfers is "fake", as it is done at start up time as though a trusted dealer would distribute field elements (masks shares) to each server.
4. The contract does not control access to the queue of intershard messages thus allowing "un-authorized" reads.

**From the above limitations, things to improve are roughly:**
- [ ] The message transmission flow can go from client to any shard, and the first receiving shard will forward it to the other.
- [ ] Not sure, but it seems to me that the "complete" message transmission (client -> shard 1 -> shard
2 does not need to be all happening within one epoch. That is, the shard that produces messages and the shard that consumes messages do not need to operate at the same rate, since the producer queues up messages and the consumer consumes the messages from the queue; they can be operating at different speeds.
- [ ] Implement randousha-based generate_intershard_masks() function.
- [ ] Tidy up the access control using a SecretCellArray data structure in the contract.


apps/helloshard/server.py
```diff
@@ -0,0 +1,422 @@
...
+    consume secret from client
+    ^^^^^^^^^^^^^^^^^^^^^^^^^^
+    1. (blockchain state read) consume a client's secret from a contract
+    2. (inner-shard communication) unmask the secret in a MPC with
+       nodes of its shard
+
+    produce masked message for other shard
+    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+    1. mask the client message with an intershard mask share
+    2. (inner-shard communication) open the masked share to get the
```
amiller: intra-shard or within-shard lets call it?

apps/helloshard/contract.sol
```diff
@@ -0,0 +1,291 @@
...
+        require(_shard_1.length == _shard_2.length);
+	    n = _shard_1.length;
+	    t  = _t;
+	    require(3*t < n);
+	    shard_1.length = n;
+	    shard_2.length = n;
+	    for (uint i = 0; i < n; i++) {
+	        shard_1[i] = _shard_1[i];
+	        shard_2[i] = _shard_2[i];
+	        servermap[_shard_1[i]] = i+1;   // servermap is off-by-one
```
amiller, line 41: I see why something like this is necessary. On the other hand, it might not be efficient if we have more than just 2 shards of different lengths. What else could we use here, mapping address => PartyStruct?

apps/helloshard/server.py
```diff
@@ -0,0 +1,422 @@
...
+            if self.is_gateway_shard:
+                client_msg_field_elem = self._collect_client_input(
+                    index=epoch, queue=contract_concise.input_queue
+                )
+
+                # 3.d. Call the MPC program
+                async def prog(ctx):
+                    logging.info(f"[{self.global_id}] Running MPC network")
+                    client_msg_share = ctx.Share(client_msg_field_elem)
+                    client_msg = await client_msg_share.open()
```
amiller, line 260: This MPC program looks unclear. Which shard runs it, and also isn't clients secret misnamed?

amiller (commented):
Looks great so far

apps/helloshard/contract.sol
sbellem: Yes definitely, it needs to be generalized/improved.

sbellem: Would it make sense to use eth addresses as node ids instead of integers (`i`)?

apps/helloshard/server.py
sbellem: Yes this needs to be improved, Ideally it would be run by any shard that has received a client secret. In this code one shard acts as a gateway shard (hence the condition at line 251 above: `if self.is_gateway_shard:`) and only the gateway shard is expected to receive inputs from the client. Whether a server is part of a gateway shard or not is set at start up time in the `main` module.

I guess that the `_mpc_initiate_loop` logic needs to be adapted to shards. That is, each shard needs to have an `inputs_ready` condition that when met can be use to trigger the start of the client's input processing by the relevant shard.

sbellem: As for the name, yeah ... I now wonder what happened ... I think I was trying to be better understand what type each variable is and tried to pick names that could reflect this. Definitely can be changed!

It is indeed confusing. The method `_collect_client_input()` is used to read the client secret from the input queue but rather than just returning the client secret (`m̅`) it subtracts the input mask share (`[r]`) and returns the result which is a `field` element that is then cast to a `Share` (`[m]`). The naming was trying to reflect this I think ... :smile:.

PR #432, opened 2020-03-26
Set the maximum line length for docstrings to 89

sbellem:
Python's style guide, PEP 8 recommends 72, but for now it is
simpler with this higher bound to limit the number of required
changes which would be 356.

### DIFF
--- a/honeybadgermpc/progs/fixedpoint.py
+++ b/honeybadgermpc/progs/fixedpoint.py
@@ -23,8 +23,11 @@ http://www.ifca.ai/pub/fc10/31_47.pdf
 
 F = 32  # The precision (binary bits)
 """
-This implementation of the library is not completely hiding. This leaks information about the bits used in computation which is determinied by the security parameter Kappa.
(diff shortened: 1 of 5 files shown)


honeybadgermpc/progs/fixedpoint.py
```diff
@@ -23,8 +23,11 @@
 
 F = 32  # The precision (binary bits)
 """
-This implementation of the library is not completely hiding. This leaks information about the bits used in computation which is determinied by the security parameter Kappa.
-In particular, we leak O(1/(2^Kappa)) information theorotic bits per operation on a floating point secret.
+This implementation of the library is not completely hiding. This leaks
+information about the bits used in computation which is determinied by
+the security parameter Kappa.
+In particular, we leak O(1/(2^Kappa)) information theorotic bits per
+operation on a floating point secret.
```
amiller, line 30: This comment looks wrong. What does it mean to leak an information theoretic bit? This is fixed point not floating point. This needs attention

amiller (approved):
fine

Issue #440, opened 2020-05-11
autodoc: failed to import module 'honeybadgermpc'


sbellem:
Related to the benchmark logger configuration.

Setting benchmark logs path to an absolute one as opposed to relative could fix this issue.

```shell
WARNING: autodoc: failed to import module 'honeybadgermpc'; the following exception was raised:                            
Traceback (most recent call last):                                                                           
  File "/usr/local/lib/python3.7/logging/config.py", line 563, in configure   
    handler = self.configure_handler(handlers[name])                             
  File "/usr/local/lib/python3.7/logging/config.py", line 736, in configure_handler                     
    result = factory(**kwargs)                                              
  File "/usr/local/lib/python3.7/logging/__init__.py", line 1087, in __init__      
    StreamHandler.__init__(self, self._open())                             
  File "/usr/local/lib/python3.7/logging/__init__.py", line 1116, in _open
    return open(self.baseFilename, self.mode, encoding=self.encoding)
FileNotFoundError: [Errno 2] No such file or directory: '/usr/src/HoneyBadgerMPC/docs/benchmark-logs/benchmark.log'

The above exception was the direct cause of the following exception:                                                                                                                                                
                                                                                                                                                                                                                    
Traceback (most recent call last):                                                                                                                                                                                  
  File "/opt/venv/lib/python3.7/site-packages/sphinx/ext/autodoc/importer.py", line 32, in import_module                                                                                                            
    return importlib.import_module(modname)                                                                                                                                                                         
  File "/usr/local/lib/python3.7/importlib/__init__.py", line 127, in import_module                                                                                                                                 
    return _bootstrap._gcd_import(name[level:], package, level)                                                                                                                                                     
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import                                                                                                                                                   
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load                                                                                                                                                 
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked                                                                                                                                        
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked                                                                                                                                                 
  File "<frozen importlib._bootstrap_external>", line 728, in exec_module                                                                                                                                           
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed                                                                                                                                      
  File "/usr/src/HoneyBadgerMPC/honeybadgermpc/__init__.py", line 16, in <module>                                                                                                                                   
[...]

PR #441, opened 2020-05-11
Documentation/bugs/438

### LINKED ISSUE #438: TypeError: add_source_parser() takes 2 positional arguments but 3 were given
Appears to be a problem with `m2r` (markdown to restructuredtext).

See https://github.com/sphinx-doc/sphinx/issues/7420.

Two possible fixes appear to be:
* Use patched `m2r` (`'m2r @ git+https://github.com/crossnox/m2r@dev#egg=m2r'` -- tip taken from https://github.com/sphinx-doc/sphinx/issues/7420#issuecomment-610958327)
* Switch to [`recommonmark`](https://github.com/readthedocs/recommonmark).

Switching to `recommonmark` appears to be preferable as `m2r` appears to be non-active in terms of releases.

Assuming the switch does not require much work, then let's try `recommonmark`.Perhaps there will be difficulties for files such as `README.md` which are outside the `docs/` namespace. See relevant comment: https://github.com/sphinx-doc/sphinx/issues/7420#issuecomment-610156586.

sbellem:
Fixes #438 and #440.

* Removed some outdated docs with respect to the internals and updated a tiny bit what is documented for the internals, just pulling docstrings from some modules with autodoc.
* Switch from `m2r` to `recommonmark` for using markdown with Sphinx (restructuredText).
* Use an absolute path for the location of the benchmark logs directory to prevent a file not found error when docs are built, as the benchmark logs dir was then looked up under the docs/ folder as the config was using a relative path.


amiller (approved):
This looks fine by inspection. Also I ran the docs building work flow and checked it makes sense and worked locally

PR #449, opened 2020-05-13
Improve the documentation generation workflow

sbellem:
Re-work the documentation workflow to use the `Makefile` and `docker-compose` to ease the documentation generation.

Basically, all that should be needed is to run

```shell
$ make servedocs
```

which will
* build the docs
* serve the docs at http://localhost:58888
* use watchdog to monitor changes to `.rst` files in a background process and rebuild the docs when changes are saved

The documentation on how to generate the documentation was also updated to reflect these changes. A preview can be seen at https://101-141841687-gh.circle-artifacts.com/0/docs/development/getting-started.html#building-and-viewing-the-documentation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,7 +75,7 @@ docs: ## generate Sphinx HTML documentation
 	docker-compose --file docs.yml run --rm build make -C docs clean
 	docker-compose --file docs.yml up build
 	docker-compose --file docs.yml stop view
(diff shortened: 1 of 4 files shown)


amiller (approved):
This works great! 
Ran with `make servedocs` and I get live re-rendering of the documentation

